PR #72, opened 2025-12-06
add design system page

hasparus:
it would be nice to have an internal page that doesn't rely on any backend logic because the backend gets refactored pretty often :P so having a separate page could allow us to land small PRs in isolation and iterate on the Uis without needing to seed the correct state

There are alternatives like `storybook-django`, but I feel one big dump HTML file is gonna be enough: https://github.com/torchbox/storybook-django


fancysnake (approved):

hasparus:
@copilot can you help me with this merge conflict?

PR #216, opened 2026-04-26
body-scroll-lock, scrollview fade, and move from django-tailwind to django-vite

hasparus:
Fixed a few bugs around the modal and made some visual improvements.
Also migrated from django-tailwind to django-vite because it's the first time we use a library, so a bundler is useful (alternatively we could use body-scroll-lock from CDN).

## Open question: `ENV` inconsistency

This PR doesn't introduce it, but it surfaced while debugging — flagging in case @fancysnake has an opinion before we ship.

`src/ludamus/edges/settings.py:394` does `dev_mode = ENV == "development" or ROOT_DOMAIN == "testserver"`. The project uses three different `ENV` values across configs:

- `.env` → `ENV=development` (what `mise run start` actually reads — dev_mode on)
- `.env.local.example` → ~~`ENV=local`~~ now `ENV=development` (updated in this PR so a fresh `cp .env.local.example .env` lands in working dev mode; previously a new contributor would get `dev_mode=False`, no HMR, stale assets)
- `.env.docker.example` → `ENV=production`
- `mise.toml` `_lint` / `_pytest` / `_e2e` tasks → `env.ENV = "local"`

The only consumers of `ENV` are `IS_PRODUCTION = ENV == "production"` and the `dev_mode` check above. So `"local"` is functionally equivalent to any non-`production`, non-`development` string — it's a third value that means nothing.

Worth consolidating to two values (`development` / `production`)? Or keeping `local` as an explicit "tests / lint, no dev server" marker?


src/ludamus/client/src/modal.css
```diff
@@ -1,4 +1,6 @@
 .modal {
+  display: flex;
+  flex-direction: column;
```
hasparus, line 3: mea culpa, the closing animation feels bad on main now

PR #228, opened 2026-05-02
Reload Django pages on template change (fix HMR)

hasparus:
Vite's built-in `full-reload` only fires when the changed HTML path matches `location.pathname`, so Django-served HTML (`/events/…`, `/encounters/…`) silently ignores it.

- `vite.config.ts`: send a custom HMR event instead of `full-reload`; bump plugin to `handleHotUpdate.order: "pre"` and register `djangoHmr` as a Vite input.
- `django-hmr.ts` (new): listen for the event, call `window.location.reload()`.
- `base.html`: load `django-hmr.ts` after the Vite HMR client.
- `settings.py`: pass \`OPTIONS.loaders\` explicitly so the cached loader only wraps in production (DEBUG must skip the cache or template edits never reach the browser).
- `modal.ts`: skip the Navigation API interceptor for non-\`push\` navigations. The interceptor was hijacking the HMR reload when the URL contained a modal-trigger param (e.g. \`?session=1\`); restricting it to genuine link clicks lets reloads through and keeps the click→modal flow intact.

### DIFF
--- a/src/ludamus/edges/settings.py
+++ b/src/ludamus/edges/settings.py
@@ -93,6 +93,7 @@ MIDDLEWARE = [
     "django.contrib.auth.middleware.AuthenticationMiddleware",
     "django.contrib.messages.middleware.MessageMiddleware",
     "ludamus.inits.RepositoryInjectionMiddleware",
+    "ludamus.inits.middleware.ServiceInjectionMiddleware",
     "ludamus.adapters.web.django.middlewares.RequestContextMiddleware",
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
     "ludamus.adapters.web.django.middlewares.RedirectErrorMiddleware",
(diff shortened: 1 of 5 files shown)


src/ludamus/edges/settings.py
```diff
@@ -110,12 +110,24 @@
...
+    if IS_PRODUCTION
+    else list(_BASE_TEMPLATE_LOADERS)
+)
+
 TEMPLATES = [
     {
         "BACKEND": "django.template.backends.django.DjangoTemplates",
         "DIRS": [BASE_DIR / "templates"],
-        "APP_DIRS": True,
+        "APP_DIRS": False,
```
hasparus, line 128: Set to `False` because we're now passing `OPTIONS.loaders` explicitly. Django rejects `APP_DIRS=True` together with an explicit `loaders` list (`ImproperlyConfigured: APP_DIRS must not be set when loaders is defined`), so the two are mutually exclusive. The `app_directories.Loader` is still in the explicit list, so app-template discovery is unchanged.

PR #234, opened 2026-05-03
Remove Tag legacy and accept-proposal flow (code-side)

fancysnake:
Drops the pre-SessionField accept-proposal feature (service, view, form, URL, template, tests), strips Tag/TagCategory consumers and Session.needs/requirements reads, and renames the SessionField filter scaffolding off its leftover tag-shaped names. Also fixes the proposed-sessions panel to scope to the logged-in proposer and hide already-scheduled proposals (agenda_item__isnull=True). Schema removal of Tag/TagCategory tables and Session columns is deferred to a separate PR.


spdd-tracer/analysis/202605031704-remove-tags-and-session-text-fields.md
```diff
@@ -0,0 +1,263 @@
...
+   `SessionField`/`PersonalDataField` instead — no parallel legacy path.
+
+## Concepts
+
+### True Tag legacy (delete)
+
+- legacy DTOs/protocols — `pacts/legacy.py` `PendingSessionTagDTO:163`,
+  `TagCategoryDTO:314`, `TagDTO:323`; protocol methods `read_tags:813`,
+  `read_tag_categories:815`, `read_pending_by_event:819`;
+  `PendingSessionDTO:178-189` loses `tags`, `needs`, `requirements`
```
hasparus, line 38: design_fixtures.py still references needs/requirements in SessionDTO construction. Do you wanna remove them there too?

src/ludamus/adapters/web/django/views.py
```diff
@@ -958,22 +948,6 @@ def _get_pending_sessions_context(self) -> dict[str, Any]:
...
-                self.request.context.current_user_slug
-            ).is_superuser
-            or is_sphere_manager
-        ):
-            return {
-                "pending_sessions": self.request.di.uow.sessions.read_pending_by_event(
-                    self.object.pk
-                )
-            }
-
```
hasparus, line 976: superusers and sphere managers viewing the public event page will only see their own pending proposals, right? Is this intentional?

spdd-tracer/analysis/202605031704-remove-tags-and-session-text-fields.md
fancysnake: yes

src/ludamus/adapters/web/django/views.py
fancysnake: nope, it's not intended

PR #238, opened 2026-05-05
Skip single-choice proposal steps

hasparus:
### Summary
- Skip the category step when an event has exactly one proposal category.
- Auto-select a single available time slot and omit the time-slot step from progress/back navigation.
- Keep proposal-description/login context on skipped-category renders and update proposal wizard integration coverage without using `ANY` for the skipped-category form assertion.

### Testing
- `mise run _pytest -- tests/integration/web/chronology/test_propose_session.py`
- `mise run check`
- Browser walkthrough on seeded event with one category and one time slot.

### Screenshots

| Initial step (category skipped, no back button) | Review (auto-selected category + time slot) |
| --- | --- |
| Initial proposal step skips category and has no back button | Review shows auto-selected category and time slot |


src/ludamus/gates/web/django/chronology/views.py
```diff
@@ -98,6 +98,30 @@ def _timeslot_descriptors(
     return [{"day": day, "slots": slots} for day, slots in sorted(groups.items())]
 
 
+def _has_category_step(categories: Sequence[ProposalCategoryDTO]) -> bool:
+    return len(categories) != 1
```
fancysnake, line 102: Category step should be always shown because it contains the proposal description content.

we could however, think of adding a EventProposalSettings field to switch to simple proposal form that will have everything on one page

hasparus: image

Does it show anything informative to the user here?

fancysnake (approved):

PR #249, opened 2026-05-07
Add desktop fold toggle to backoffice panel sidebar

fancysnake:
Sidebar can collapse to icon-only (64px) on desktop to free horizontal space on views like the agenda builder. Toggle persists via localStorage and applies pre-paint to avoid flash. Mobile hamburger flow is untouched.


hasparus (approved):
uhh i feel there's something sus, but I'm reading it on a phone, lgtm i guess? gonna look at devinreview

PR #257, opened 2026-05-09
Fix profile form labels (hidden inputs, capitalize)

hasparus:
## Summary

- `tessera_field` rendered a `<label>` wrapper around `HiddenInput` widgets, so the profile edit page showed a stray "User type *" label. Hidden widgets now fall back to Django's default rendering.
- Added `first-letter:uppercase` to the label class so lowercase Polish translations like "adres email" display capitalized without changing the source strings.

## Test plan

- [x] Profile edit page no longer shows "User type *"
- [x] All label texts render with the first letter uppercased

### DIFF
--- a/.github/workflows/mobile.yml
+++ b/.github/workflows/mobile.yml
@@ -60,14 +60,17 @@ jobs:
           xcrun simctl bootstatus "$UDID" -b
           echo "UDID=$UDID" >> "$GITHUB_ENV"
 
+      - name: Bootstrap e2e fixtures
+        run: mise run boot-e2e-prep
+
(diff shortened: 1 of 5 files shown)


hasparus:
Heads-up: this PR's `mise.toml` change to `boot-e2e` (split into prep + serve) overlaps with #261, which rewrites the same task block to mise's array-of-table form. `.github/workflows/mobile.yml` auto-merges fine. Whichever PR lands second will need to translate the split into the other's format — trivial in either direction. #261 is currently `CONFLICTING` with main + red on `test`, so leaving this PR as-is for now.

PR #261, opened 2026-05-09
Use mise task definition syntax and parallelize some jobs

fancysnake:


docker/mise.toml
```diff
@@ -5,7 +5,6 @@ jobs = 1
 [tools]
 python = "3.14"
 node = "v25.6.1"
-poetry = { version = "2.3.2", pyproject = "/app/pyproject.toml" }
```
hasparus, line 8: Are we not using Poetry anymore?

fancysnake: We are, but I prefer installing it in the virtualenv and guarding version in `tool.poetry` settings

PR #398, opened 2026-06-19
Adopt ESLint 9 (@hasparus/eslint-config) for the frontend client + fix all violations

hasparus:
## What

Adds ESLint 9 (flat config) to the browser-only TypeScript client via `@hasparus/eslint-config`, fixes every resulting violation, and makes the lint step CI-blocking.

## Adoption

- devDeps `eslint@^9` + `@hasparus/eslint-config@^2.1.1` (typescript kept at 5.9.3). Package manager is `aube`; `aube-lock.yaml` updated and verified to resolve frozen without any trust-policy exclusion (CI needs none).
- `eslint.config.js` spreads the config's `./the-guild` array (the package root only re-exports it as a named binding, no default export).
- `lint` script runs `eslint . --max-warnings 0` so warnings block too.
- Required resolution adjustments: `package.json "type": "module"`, and in `tsconfig.json` `moduleDetection: force` (standalone scripts become modules) + `types: []` (so the transitive `@types/node` doesn't shadow DOM `setTimeout`). These are **not** the deferred strict flags.
- e2e specs are a separate package/runtime and are intentionally out of scope here.

## Cleanup

158 problems (82 errors, 76 warnings) → **0**. ~136 auto-fixed; the rest by hand — removed an unjustified `Boolean(...)` + non-null `!` by narrowing, typed DOM nodes so `.dataset` access checks, replaced CJS `__dirname` with `import.meta.url`, hoisted a pure helper, and converted `.forEach` loops to `for…of` while preserving control flow. No global rule disables, no per-line `eslint-disable`; `console.warn`/`console.error` are allowed for the client (legitimate runtime diagnostics).

## CI

`frontend-analysis` now runs a **blocking** `eslint` step (`mise run lint-client`) before the non-blocking `fallow` analysis. The Python check path is untouched.

## Verification

`eslint` clean (0), `tsc --noEmit` clean, Prettier clean (TS + config), `vite build` succeeds. Playwright e2e not run here (needs a live server/fixtures); changes are behavior-preserving refactors.

## Follow-up

- Flip the strict tsconfig flags from the TS standards audit (`noUncheckedIndexedAccess`, etc.) in a separate PR.
- Optionally extend ESLint to the `tests/e2e` package with its own flat-config block.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JGAT6FAhSwE2zk5fbcaGUW)_

## Summary by CodeRabbit

* **Chores**
  * Strengthened code quality standards by implementing mandatory TypeScript linting with zero-warning enforcement in the continuous integration pipeline for the frontend.
  * Modernized frontend build configuration and infrastructure to ensure consistent code patterns, improve code reliability, and enhance long-term application maintainability.


hasparus:
## Automated stale-PR maintenance

**Fixed: CodeQL failure** (`encounter-form.ts:78`)

CodeQL was flagging `preview.src = previewUrl` as "DOM text reinterpreted as HTML without escaping meta-characters". The value is always a `blob:` URL produced by `URL.createObjectURL(file)` — it cannot be a `javascript:` URL and cannot cause XSS — but CodeQL's taint analysis doesn't model `URL.createObjectURL` as a sanitizer.

**Fix applied** (commit `0c0823e`): added an explicit `startsWith("blob:")` guard before the `img.src` assignment. `URL.createObjectURL` always returns a `blob:` URL so the guard is always true, but it makes the safety visible to static analysis and is correct defensive practice.

```ts
const objectUrl = URL.createObjectURL(file);
if (!objectUrl.startsWith("blob:")) {
  URL.revokeObjectURL(objectUrl);
  return;
}
previewUrl = objectUrl;
preview.src = previewUrl;
```

No other unaddressed review comments were found on this PR. CI should go green once CodeQL re-runs.

---

PR #400, opened 2026-06-19
feat(sessions): soft-delete foundation + session soft-delete (#331)

### LINKED ISSUE #331: Soft delete sessions
## Summary

Sessions (proposals) currently have **no delete path at all** — `SessionRepository` exposes no `delete`, and no view/mill removes a `Session`. We want the ability to remove a session, done as a **soft delete** (mark as deleted / archived rather than hard-removing the row), so it can be hidden from listings without losing history or breaking related records (agenda items, field values, schedule-change logs).

## Motivation

- Organizers/facilitators need a way to retract a session that was proposed/created in error or withdrawn.
- A soft delete keeps referential history intact and is reversible.

## Scope / tasks

- [ ] Add a `deleted_at` (or `is_deleted`) field to the `Session` model + migration.
- [ ] `SessionRepository.delete(pk)` performs the soft delete; reads/listings exclude soft-deleted sessions by default.
- [ ] Wire a view/action (with appropriate permission checks) to trigger it.
- [ ] Decide cascade semantics for related rows (agenda items, RSVPs/enrollments, field values).
- [ ] Integration tests for the delete path and listing exclusion.

## Storage-cleanup follow-up (from PR #330)

PR #330 added `Session.cover_image`. `EventRepository.update` and `EncounterRepository.update` already clean up replaced files via the shared best-effort `_delete_stored_file` helper, but there is **no cleanup on session removal** — because no session-removal path exists yet.

When this issue lands:

- [ ] On (soft or hard) session removal, delete the stored `cover_image` via `_delete_stored_file` so files in `sessions/` don't accumulate. For a pure soft delete the file may be retained for reversibility; decide policy (e.g. purge on hard-delete/GC of soft-deleted rows).
- [ ] If post-creation **session editing** is later added (noted as future work in PR #330), apply the same replaced-file cleanup that the event/encounter forms use.

This was flagged by CodeRabbit on PR #330 as a potential storage leak; it is not a live bug today (no delete/edit path exists), so it's deferred here as future work rather than blocking that PR.

_Refs: PR #330 (event & session cover images)._

hasparus:
**PR 1 of 3 in the soft-delete → discounts stack.** Closes #331.

Adds a reusable `SoftDeleteModel` mixin (nullable `deleted_at` timestamp + alive-only default manager + `all_objects` accessor) and applies it to `Session`. The default `objects` manager filters `deleted_at__isnull=True`, so every existing read — including reverse relations and legacy `request.di.uow` repos — excludes soft-deleted sessions automatically.

Write path goes through a new `request.services.session_deletion` service (mirrors the Announcement vertical slice): in one transaction it frees the timetable slot via the existing unschedule path (delete AgendaItem, return session to PENDING) and soft-deletes the session. `SessionParticipation` rows are retained as history. Cover image is kept on soft-delete (reversible); hard-GC purge is a #330 follow-up.

UI: scoped Delete action on the proposal-detail panel (sphere-manager permission; the service additionally enforces event membership so cross-event deletes 404) with Polish translations (session → "punkt programu").

The mixin is built to be reused by PR 2's new `Discount` model.

### Tests
14 new (repo soft-delete/exclusion/restore/NotFound; view permission allowed+forbidden, listing exclusion, slot freed, participations retained, cross-event 404); full regression suites green locally.

### Notes for reviewers
- Permission is organizer/sphere-manager; facilitator self-withdrawal was **not** wired (panel is organizer-facing) — flag if you want it.
- Re-deleting an already-deleted session returns "not found" rather than silently re-stamping.

> ⚠️ Stacked PR — do not merge before reviewing; part of a 3-PR stack.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JGAT6FAhSwE2zk5fbcaGUW)_

## Summary by CodeRabbit

* **New Features**
  * Added backoffice actions to soft-delete and restore proposals from the event schedule, including confirmation prompts, redirects, and success/error messaging.
  * Deleted proposals now appear in a **“Recently deleted”** section with per-item restore controls; default listings hide deleted items.
  * Restore re-activates the session while preserving prior participation history.
* **Internationalization**
  * Added Polish translations for deletion/restore labels, notifications, and confirmation text.


hasparus:
@coderabbitai review

---

hasparus:
Thanks @coderabbitai — checked all three against the rebased code:

1. **`@cached_property` is already present** — `inits/services.py:121`, on the line directly above `def session_deletion`. The `grep -A10 "session_deletion"` in the analysis anchors on the `def` line and shows lines *after* it, so the decorator on the preceding line fell outside the window. No fresh `ScheduleChangeLogRepository` per access. No change needed.

2. **TOCTOU** — correct that the `require_session_in_event` read precedes `atomic()`. This is consistent with the established pattern in this service (e.g. the assign path at `chronology.py:462`), and it's an organizer-only admin action, so risk is low as you note. A real fix needs `select_for_update` on the session row — the repo read methods don't lock today — so merely moving the check inside `atomic()` wouldn't actually close the window. Tracking proper row-locking as a follow-up rather than shipping a half-measure.

3. **`Session.all_objects` redeclaration** — keeping it intentionally: `SessionRepository` relies on `Session.all_objects`, and the explicit declaration documents that accessor at the point of use instead of leaning on abstract-manager inheritance being obvious to future readers. Harmless, as you note.

---

PR #429, opened 2026-06-23
ci: pull prebuilt GHCR image on prod instead of building on host (step 2)

hasparus:
## Why

Completes the move off on-host builds started in #428. The production deploy currently builds the Docker image **on the prod host** (`docker compose up --build`) — the exact step that failed in deploys #23/#24 when the host's network stalled. This switches prod to **pull** the image CI already publishes to GHCR.

> **Stacked on #428.** Base is `claude/ci-build-images`; this diff shows only the consumer change. Once #428 merges to `main`, I'll retarget this to `main`.

## What

- **`docker/compose/prod.yaml`** — replace the shared `build:` block with `image: ${APP_IMAGE}` for `web` / `migrations` / `collectstatic` (one shared image). `APP_IMAGE` is required (`:?`), so a misconfigured deploy fails loudly instead of silently building.
- **`.github/workflows/deploy-production.yml`**:
  - New **"Verify the prebuilt image exists in GHCR"** step — fails fast *before* touching prod if CI hasn't published `sha-<commit>` yet.
  - **Deploy** step now logs in to the private GHCR package, `docker compose pull`s, and `up -d` **without `--build`**, then `docker logout`. `APP_IMAGE` pins the exact image built for the deployed revision.

## Required setup (private package → prod needs credentials)

Add these to the **`production`** GitHub Environment before the first pull-based deploy:

| Kind | Name | Value |
|------|------|-------|
| Secret | `GHCR_TOKEN` | a token with **`read:packages`** for the private package |
| Variable | `GHCR_USER` | the GitHub account that owns the token |

A fine-grained PAT (read-only, packages) or a classic PAT with `read:packages` both work. Keep the GHCR package **private**.

## Rollout order

1. Merge #428 → `main` so the image-build workflow runs and publishes `sha-<commit>`.
2. Add `GHCR_TOKEN` / `GHCR_USER` to the `production` Environment.
3. Retarget + merge this PR.
4. Dispatch a production deploy — the verify step confirms the image, prod pulls it, no on-host build.

## Follow-up

Mirror the same pull-based flow for **staging** once prod is proven.

https://claude.ai/code/session_01HgHmgqSBB6yo2tZtDRVoDZ

---
_Generated by [Claude Code](https://claude.ai/code/session_01HgHmgqSBB6yo2tZtDRVoDZ)_

### DIFF
--- a/.github/workflows/deploy-production.yml
+++ b/.github/workflows/deploy-production.yml
@@ -90,17 +90,31 @@ jobs:
           source: .env.local
           target: ludamus/
 
+      - name: Verify the prebuilt image exists in GHCR
+        # Fail fast (before touching prod) if the "Build production image"
+        # workflow hasn't published an image for this commit yet.
+        env:
+          GHCR_TOKEN: ${{ secrets.GHCR_TOKEN }}
+          GHCR_USER: ${{ vars.GHCR_USER }}
+        run: |
+          echo "$GHCR_TOKEN" | docker login ghcr.io -u "$GHCR_USER" --password-stdin
+          docker buildx imagetools inspect \
+            "ghcr.io/${{ github.repository }}:sha-${{ github.sha }}"
+
       - name: Deploy
         uses: appleboy/ssh-action@v1
         env:
           GITHUB_COMMIT_SHA: ${{ github.sha }}
+          APP_IMAGE: ghcr.io/${{ github.repository }}:sha-${{ github.sha }}
+          GHCR_USER: ${{ vars.GHCR_USER }}
+          GHCR_TOKEN: ${{ secrets.GHCR_TOKEN }}
         with:
           host: ${{ secrets.PRODUCTION_HOST }}
           username: ${{ secrets.PRODUCTION_USER }}
           key: ${{ secrets.PRODUCTION_DEPLOY_KEY }}
-          envs: GITHUB_COMMIT_SHA
-          # Default is 10m; a cold pull + rebuild of the images exceeds it and
-          # drone-ssh kills the command mid-build ("Run Command Timeout").
+          envs: GITHUB_COMMIT_SHA,APP_IMAGE,GHCR_USER,GHCR_TOKEN
+          # Default is 10m; a cold pull of the prebuilt images can exceed it and
+          # drone-ssh kills the command mid-pull ("Run Command Timeout").
           command_timeout: 13m
           script: |
             set -ex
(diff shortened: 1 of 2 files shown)


hasparus:

The compose refactor is clean and the intent is right: three services stay on a single `image: ${APP_IMAGE:?}` via the existing anchor, `:?` fails closed, and a fail-fast GHCR verify runs before the SSH deploy. But the deploy script contradicts its own comment — it does **not** guarantee `docker logout` on failure, leaving registry credentials on the production host — and it traces a secret under `set -x`. The PR is also **partially landed**: `build-image.yml` is already on `main` byte-identical (blob `607e7f09`), so this needs a rebase down to the two-file deploy/compose delta.

### 1. `docker logout` is not guaranteed — GHCR credentials leak onto the prod host on any deploy failure (**blocker**)

`.github/workflows/deploy-production.yml:120,128,135` — the script runs under `set -ex`, logs into GHCR at line 128, and relies on line 135 for logout. With `set -e`, if `docker compose pull` (133) or `up -d` (134) fails, the script exits and logout never runs — the token persists in `~/.docker/config.json` (base64, not encrypted) on the prod host indefinitely, violating the code's own comment at 126–127. This triggers on exactly the deploys where you least want lingering creds (bad image, failed migration).

Remedy:
```bash
echo "$GHCR_TOKEN" | docker login ghcr.io -u "$GHCR_USER" --password-stdin
trap 'docker logout ghcr.io' EXIT
```
and delete the explicit logout line.

### 2. `set -x` traces the token into the remote command stream (**follow-up**)

Line 128 under `set -x` emits `+ echo <token>` in the remote shell trace streamed back by `appleboy/ssh-action`. GitHub's masking likely catches it in the Actions log, but relying on masking for a value that also transits the host's process/output stream is fragile — and main's current script carries no secret inside its `set -x` block, so this is a new exposure surface. Remedy: `set +x; …login…; set -x` around just the login.

### 3. Partially landed — rebase and drop `build-image.yml` (**merge-blocker for the PR as-is**)

`build-image.yml` from commit `a172cd0e` is already on `origin/main` as the identical blob; only the `deploy-production.yml` + `prod.yaml` switch remains unmerged, and main's copies of both are byte-identical to this PR's diff base, so a rebase is clean. Do **not** close the PR — main still builds on the host (`up --build --pull always`), so this is still the correct next step.

### 4. No documented rollback path (**follow-up**)

`up -d` recreates `web` in place; if the pulled image boots broken there's no health-gated promotion or automated revert. The manual runbook (re-dispatch pinned to the prior green `sha-<oldsha>` tag, which survives on GHCR and isn't pruned locally since tagged images aren't dangling) works but is undocumented — one comment near the deploy step would save an operator mid-incident.

### 5. Nit: the verify step's "before touching prod" comment overstates

The preceding `scp` already wrote `.env.local` to the host — the verify is fail-fast for the *image*. Also the runner-side `docker login` (line 100) is never logged out; benign on an ephemeral runner, but a trailing logout costs nothing.

### Structural assessment

The compose change is a genuine code-judo move: `x-app-build` → `x-app-image` keeps all three services on one image through the untouched `x-app-defaults` merge, deletes host-side build args, adds zero branching. No file-size or spaghetti concerns — the issues are operational (credential lifecycle, secret tracing, rollback), not structural.

**VERDICT: BLOCK** — sound direction, clean compose refactor; add the `trap`, wrap the login in `set +x`, rebase to the two-file delta, and this becomes an approve.

---

hasparus:
Superseded by the Coolify deploy path (#599) — it already deploys the CI-built GHCR image, so upgrading the SSH pull flow is no longer worth landing.

PR #471, opened 2026-07-01
Unify copy-to-clipboard on a tessera component with a floating "Copied!" popover

hasparus:
## What & why

Copy-to-clipboard was implemented four different ways (a `✓`-glyph swap that **resized the button**, two `textContent` swaps that did the same, and two buttons with **no feedback at all**), each with its own inline script. The Discord chip even had two handlers attached, so every click wrote the clipboard twice.

This PR collapses all of them onto one tessera component: buttons keep their existing looks, never change size, and confirm with a small floating **"Copied!"** popover styled after the heading copy-link popover in the graphql-hive docs.

## How

**One behaviour** — `src/copy.ts`, a small delegated listener on `[data-copy]`, loaded globally like `flash.ts`. On a successful write it fills and reveals the button's `[data-copy-popover]` for 1.5 s. The popover is `role="status" aria-live="polite"`, so screen readers hear the confirmation; it's `absolute` + `pointer-events-none`, so the button's box never changes. Confirmation is gated on the clipboard promise — an unavailable (insecure context) or rejected write shows nothing instead of lying, and logs to the console.

**Declarative markup via tessera tags** (`templatetags/tessera/copy.py`):

- `{% tessera_copy %}` — block tag; the component owns the look (`variant="button"` default, `"menu-item"` for dropdown rows), the callsite provides only icon + label. `origin=True` prefixes `location.origin` for host-agnostic share paths. Unknown kwargs raise instead of vanishing.
  ```django
  {% tessera_copy share_url variant="menu-item" %}
      {% icon "clipboard-document" variant="solid" class="h-4 w-4 inline" %}
      {% translate "Copy link" %}
  {% endtessera_copy %}
  ```
- `{% tessera_copy_chip "@"|add:discord_username %}` — preset for short values where the visible text *is* the copied value; the whole chip is the click target and the text is part of the accessible name (WCAG 2.5.3). Labels default to translated "Copy to clipboard" / "Copied!".
- `{% copy_lines a b c as payload %}` — composes multi-line payloads (skips empties) for "Copy info" / "Copy details".
- `{% tessera_copy_popover %}` — the popover primitive, for bespoke buttons.

**Migrated callsites** (looks unchanged, icons standardized on `clipboard-document` solid): Discord handle chip, encounter-card share link, share-menu "Copy link" / "Copy info", both RSVP "Copy details". The four bespoke inline scripts and the orphaned `#share-url` / `#share-info-data` nodes are deleted.

**Deliberate UX changes:** the share menu stays open on copy so the confirmation is visible (click-outside still closes); the RSVP save-prompt collapses after the popover has been seen.

## Screenshots

Real built CSS; buttons were actually clicked and the clipboard contents asserted.

**All three shapes** — chip, share button (absolute URL via `origin=True`), menu row (multi-line payload); popover shown on the share button:

All copy control shapes

**The chip, light + dark:**

Copy chip — light

Copy chip — dark

## Tests & verification

- 17 new integration tests (`TestCopyChip`, `TestCopyBlock`, `TestCopyPopover`, `TestCopyLines`): variants, label defaults/overrides, live region, XSS escaping, newline survival in `data-copy`, unknown-kwarg guard, no-resize invariants.
- End-to-end with the compiled handler: chip → `@ada`, share button → absolute URL, menu row → multi-line payload; one popover each.
- `messages-check`, `djlint`, `ast-grep`, `ruff`/`black`/`pylint`, eslint, `tsc`, vite build all green.
- Two thermo-nuclear review passes; final verdict SHIP-WITH-NITS with the nits addressed (kwargs guard, clipboard-unavailable logging, test gaps). Deferred follow-ups: extract the thrice-repeated copy-details payload into a partial, and replace the save-prompt collapse timer with a success event from `copy.ts`.

> Screenshots are hosted on the `pr-471-assets` branch (not part of this PR's diff).


hasparus:
@coderabbitai Regarding the copy.ts "add a client-side test runner" suggestion: this project deliberately has no separate JS unit-test runner. Please remember: end-to-end Playwright tests (tests/e2e) are the preferred and best tests in this project — client behaviour is covered there, against the real app. A copy-flow e2e now exists in `tests/e2e/tests/notice-board.auth.spec.ts` (role/text selectors, real clipboard, popover + no-resize assertions), which also addresses the "User-Centric E2e Tests" pre-merge check.

---

PR #524, opened 2026-07-05
Move proposal description to EventProposalSettings repository

fancysnake:
Drops EventDTO.proposal_description and the _event_dto helper; the description now flows through EventProposalSettingsRepository.update_description and read_or_create_by_event. Views and all propose-wizard step templates read it via proposal_settings.

## Summary by CodeRabbit

* **New Features**
  * Proposal description display and persistence now use the event’s proposal settings, ensuring the same text is used throughout the proposal wizard.
  * Proposal settings are now included in the wizard’s rendered context on all steps, including re-renders after validation errors.

* **Bug Fixes**
  * Resolved cases where the description shown in the proposal flow could differ from the description stored in proposal settings.

* **Tests**
  * Updated integration assertions to account for the new proposal settings-driven description behavior.


hasparus:
doesn't the get_or_create get us a GET that writes rows? (excellent sentence, sorry)

PR #542, opened 2026-07-06
Facilitator crud improvement

fancysnake:
## Summary by CodeRabbit

* **New Features**
  * Added organizer tools to search, filter, sort, and customize facilitator list columns.
  * Added facilitator triage actions: flag for deletion, remove deletion flag, and mark as guest.
  * Added configurable “Facilitator columns” to choose which personal-data fields are shown and filterable.
  * Added organizer-only internal comments to facilitator details and edit screens.
* **Bug Fixes**
  * Improved facilitator list rendering and field value formatting, and ensured internal comments and accreditation changes are preserved correctly.
* **Tests**
  * Expanded integration coverage for facilitator list behavior, actions, column settings, and internal comment persistence.


hasparus:

Solid, well-tested feature work: the filter/sort logic is a clean whitelist dispatch rather than an if/elif swamp, the dynamic-column values are batched into one query (no N+1), quick-action reads are event-scoped so the IDOR surface is closed, and the tests assert real behavior (filtered result sets, flagged state, sort order, pagination) instead of bare 200s. The blocking problem is architectural, not local: the PR expands the retired `request.di.uow` legacy surface — the exact drift GLIMPSE pattern #2 forbids — even though this same PR proves the correct `request.services` path is available and easy to use.

### 1. Legacy `request.di.uow` surface is extended by new code — **blocker**

CLAUDE.md/GLIMPSE is explicit: new code never uses `request.di.uow` and must not extend it; if no service exists, create one in mills + protocol in pacts + leaf in `inits/services.py`. This PR does the opposite in several new places:

- New repository grafted onto the legacy UoW: `uow.py:37-40` (`event_panel_settings`), with `EventPanelSettingsRepositoryProtocol` added to `UnitOfWorkProtocol` (`pacts/legacy.py:1519`) — a brand-new persistence surface on the layer that's supposed to be shrinking.
- New write methods on the legacy surface: `facilitators.set_flag` (`repositories/submissions.py:955`), `host_personal_data.list_values_for_event` (`:987`), extended `facilitators.list_by_event(filters)`.
- New views drive those writes through `uow` directly: `facilitators.py:478/:487` (flag), `:499` (mark-guest), `:541` (column settings), the list read at `:87-121`.

What makes this a hard block rather than "pre-existing legacy": the edit POST in this very file already does it right — `request.services.host_personal_data.update_facilitator(...)` at `facilitators.py:346`. Remedy: a small facilitator-triage service method (flag/unflag/mark-guest, scope-guarded in the service per `docs/refactors/panel-object-scope-authz.md`) and a service over `EventPanelSettings`, with protocols in pacts and leaves in `inits/services.py`.

### 2. Mark-as-guest silently skips the change log the edit path records — **follow-up**

`FacilitatorMarkGuestActionView._apply` (`facilitators.py:495-502`) calls `facilitators.update(...)` directly; its own comment concedes it "skips the `FacilitatorChangeLog` the full edit form records." An accreditation mutation with inconsistent auditability depending on which button you used. Folding it into finding #1's service fixes this for free (log the change the way `HostPersonalDataService.update_facilitator` does, `mills/submissions/personal_data_fields.py:254-266`).

### 3. Hand-rolled filter form and table cells instead of tessera components — **follow-up**

The table wrapper uses `{% tessera_table %}`, but the filter bar is hand-authored `<input>`/`<select>` with repeated Tailwind strings (`facilitators.html:47-118`) and the `<thead>`/`<td>` cells repeat the same class literal many times (`facilitators.html`, `_sort_th.html:3-4`). Push the field/label/select scaffolding behind tessera primitives so the styling isn't copy-pasted per input.

### 4. Column values are assembled for the whole event, not the current page — **follow-up**

`list_values_for_event(current_event.pk, …)` (`facilitators.py:119`) loads chosen-column values for every facilitator, then the view paginates to 50. The `# ponytail:` comment acknowledges it. Narrowing the batch to `page_obj.object_list` pks is a one-line change that removes an over-fetch growing unbounded with event size.

### 5. No cross-event regression test for the new quick-action endpoints — **follow-up**

The scoping is correct (`read_by_event_and_slug(current_event.pk, …)`, `facilitators.py:452`), but `panel-object-scope-authz.md`'s definition of done wants a regression test proving a *foreign event's* facilitator is rejected without side effects. `test_flag_missing_facilitator` covers a nonexistent slug, not an existing foreign one. One test per action (flag/unflag/mark-guest); the columns POST already has the equivalent.

Nit: `FacilitatorsPageView.get` inlines filter parsing, query, pagination, and column assembly in one ~75-line method setting ~13 context keys — extracting the filter-parse block would help, but it reads fine as-is.

**VERDICT: BLOCK** — good feature, wrong layer for the new writes. Finding #1 (plus #2, which it fixes for free) is the merge gate; the rest are follow-ups.

---

hasparus:
## Stale-PR maintenance pass

Flagged by the stale-PR sweep: no commits in 2+ days, merge conflicts with `main`, and a failing `tingle check` in CI.

### Merge conflict
Merged `main` into `facilitator-crud-improvement` (commits `bbb7b549`, `11e3c42e`). The only real conflict was `src/ludamus/locale/pl/LC_MESSAGES/django.po` (catalog drift) — resolved by regenerating with `mise run messages`. Also added a Django merge migration (`0130_merge_20260715_1230.py`) reconciling this branch's `0119–0121` leaf against main's `0125–0129` leaf.

### CodeRabbit / Rams comments
All 5 CodeRabbit review threads on this PR are already marked resolved (fixed in prior commits `90a8968`, `cf93a07`) — nothing left to address there. No Rams review ran on this PR. The one substantive unresolved comment is @hasparus's "Thermo-nuclear code-quality review" (2026-07-11), whose **blocking** finding — new facilitator-triage writes going through `request.di.uow` instead of `request.services` — was **already fixed** by commit `bd2fc0d5` (`refactor(panel): move facilitator list/triage off the legacy UoW`), authored before this maintenance pass started. That review's 4 remaining items are explicitly framed as follow-ups (tessera components, column-value batching, cross-event regression test) and were left alone as out of scope for a maintenance pass.

### `tingle check` — partially fixed, remainder is justified debt
The failure is the `legacy-loc` metric (tracks `src/ludamus/adapters/**/*.py` + `**/legacy.py`), not `request-di-uow` (that one's already net `-1`, confirming the blocking finding above is really fixed). I moved `EventPanelSettingsDTO`/`EventPanelSettingsRepositoryProtocol`/`FacilitatorListFilters` out of `pacts/legacy.py` into `pacts/submissions.py` (commit `cdd51fb6`) — they're only consumed by the new `FacilitatorPanelService` path, not by legacy UoW callers, so they had no reason to live in the tracked legacy module.

That shrinks `legacy-loc` growth from **+129 → +120** (`tingle check` still fails, policy is `sum`/net-zero). The remainder is:
- **~94 lines: three new Django migrations** (`facilitator_flagged_for_deletion`, `eventpanelsettings`, `facilitator_internal_comment`) plus the matching `models.py` fields. These *have* to live under `src/ludamus/adapters/db/django/migrations/` today — that's still the live app location. They only count as "legacy" because **PR #575 ("Move db to links")**, which relocates the whole models/migrations tree to `links/db/django/`, hasn't merged yet. Once it does and this branch rebases onto the new base, these same lines stop matching the legacy-code range entirely — no code change needed, just a rebase after #575 lands.
- **~5 lines**: `internal_comment`/`flagged_for_deletion` fields added to `FacilitatorDTO`/`FacilitatorUpdateData`/`FacilitatorListItemDTO` in `legacy.py` — these core DTOs are still genuinely shared with legacy `request.di.uow` read paths (e.g. `read_by_event_and_slug`), so they can't reasonably move; a small, justified addition for real feature fields.

I didn't chase this further by narrowing `FacilitatorPanelRepos.facilitators`'s protocol type away from directly reusing `legacy.py`'s `FacilitatorRepositoryProtocol` (that would shave a little more) or by paying down unrelated debt elsewhere to force the sum to zero — the former is a real but separate cleanup, and the latter is exactly the "gaming a counter" CLAUDE.md warns against. Recommend: rebase onto `main` after #575 merges (should close most of the remaining gap for free), and treat the rest as a documented, justified exception if it's still nonzero at that point.

**Verified:** `mise run test:py` → 2974 passed, 8 skipped. `mise run check` → clean, 10.00/10 (tingle is the one still-red job, for the reason above). `mise run messages-check` → fresh, fully translated.

Pushed to `facilitator-crud-improvement` (`cdd51fb6`). Not merged.

---
_Generated by [Claude Code](https://claude.ai/code) — automated stale-PR maintenance_

---

Issue #543, opened 2026-07-07
UX audit (Tonsky-style): error pages, enrollment, panel terminology, URLs


hasparus:
A UX audit of the public app and the organizer panel, written in the spirit of Nikita Tonsky (respect the user's time; don't make them do work the computer can do; don't dress a failure up as a joke). Every claim below was verified against the code and against click-throughs as a real organizer (`e2e-manager`).

This is a tracking issue. Each unchecked box is a candidate for a focused PR; the ones marked _(PR)_ are being picked up now.

> **Status (updated):** every focused PR has landed, and the three larger "needs-decision" items (A1, P1, P5) shipped too. What's left is the smaller 🟡/🟢 polish — see **Remaining** at the bottom.

---

## Credit first (don't regress these)

- **404 URL-recovery** (`src/ludamus/adapters/web/django/error_views.py:53-87`) strips trailing junk from event links and 301s to the canonical URL. Keep it.
- **Propose wizard already skips dead steps**: single category is auto-selected (`gates/web/django/chronology/views.py:184-185`), single timeslot collapses, and `contact_email` is prefilled from `request.user.email` (`views.py:308`). This is the "don't show a one-option form" rule, done right.
- **Panel empty states teach**: CFP and Import both say what's missing and where to go.
- **The scheduling grid is a real tool**: live conflict detection, over-capacity tiles, an unassigned backlog. Genuinely hard, genuinely done.

---

### ✅ A1 — Enrolling in a session is a spreadsheet — shipped in #555
`src/ludamus/templates/chronology/enroll_select.html:134-283`
The most common action in the app — "I want to play in this session" — renders as a radio-button matrix (`No change | Enroll | Join Waiting List | Cancel`), with **No change pre-checked**, header "Select enrollment options for each user", and a plural submit button "Enroll Selected Users". A solo player with no companions gets a one-row grid to enroll themselves. The matrix is right for the party-leader case (~10%); it shouldn't be the default UI for everyone.
**Fixed:** #555 adds one-click self-enroll straight from the event page for the solo / no-companions case; the matrix stays for parties.

### ✅ A2 — Error pages are a slot machine that informs nobody — shipped in #546
`error_views.py:96-217` (404), `:220-337` (500)
[...]

PR #544, opened 2026-07-07
fix(panel): Proposals defaults to all statuses

hasparus:
## Problem

The organizer Proposals page defaulted its status filter to **Pending** whenever no `status` query param was present. An event whose sessions weren't created via proposals (e.g. imported or created directly) would then render **"No proposals submitted yet."** on first load, even when accepted/scheduled sessions existed. Misleading.

## Before / After

**Before:** first load (`/panel/event/<slug>/proposals/`) implicitly applied `status=pending`, hiding accepted/on-hold/rejected/scheduled proposals and showing the empty state when no pending ones existed.

**After:** first load shows **every** proposal regardless of status. The status `<select>` already marks "All statuses" as selected when unfiltered (`{% if not filter_status %}`), so the dropdown reflects the new default and still lets the organizer narrow to Pending / Accepted / On hold / Rejected / Scheduled. The meaning of the explicit `scheduled` pseudo-filter and each real status is unchanged.

The empty state ("No proposals submitted yet.") now only appears on the default view when the event truly has no proposals.

## Testing

Run from the worktree venv (`PYTHONPATH=src`, `.env.test` loaded):

- `pytest tests/integration/web/panel/test_proposals_page.py -q` → **33 passed**
- `pytest tests/integration tests/unit -k proposals -q` → **76 passed, 2692 deselected**
- `mypy src` → **Success: no issues found in 209 source files**
- `ruff check --fix scripts src tests` → **All checks passed!**

Updated `test_default_status_filter_shows_only_pending` → `test_default_status_filter_shows_all_statuses` (pending + accepted + rejected all appear, `filter_status is None`), and flipped the shared/inline `filter_status` context expectations from `PENDING` to `None`.

Refs #543 (P2)

---
_Generated by [Claude Code](https://claude.ai/code/session_017y1ivr5CqCWLA8eGDCGqZj)_

## Summary by CodeRabbit

* **Bug Fixes**
  * The proposals list now shows all session statuses by default when no status filter is selected, instead of limiting results to pending items.
  * The special “scheduled” filter continues to work separately from regular session status filters.
  * Updated the proposals panel to reflect the new default filtering behavior in displayed results.

### DIFF
--- a/src/ludamus/gates/web/django/chronology/panel/views/proposals.py
+++ b/src/ludamus/gates/web/django/chronology/panel/views/proposals.py
@@ -100,11 +100,13 @@ class ProposalsPageView(PanelAccessMixin, EventContextMixin, View):
         if filter_category_pk not in {c.pk for c in categories}:
             filter_category_pk = None
 
+        # Default (no status param) shows every proposal: an event whose
+        # sessions weren't created via proposals should not look empty on first
+        # load. Explicit picks (a real status or the "scheduled" pseudo-filter)
+        # still narrow the list.
         status_raw = self.request.GET.get("status")
         filter_status: str | None
-        if status_raw is None:
-            filter_status = SessionStatus.PENDING
-        elif status_raw == SCHEDULED_FILTER or status_raw in set(SessionStatus):
+        if status_raw == SCHEDULED_FILTER or status_raw in set(SessionStatus):
             filter_status = status_raw
         else:
             filter_status = None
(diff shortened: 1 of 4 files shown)


hasparus:
Hey maybe we shouldn't show "No proposals submitted yet." if there is a filter. Instead it should be "No proposals matching the filter".

hasparus:
Done in 55600a5. When a filter/search is active but nothing matches, the empty state now reads **"No proposals match the current filters."** with a **Clear filters** link; the "No proposals submitted yet." + Create-Session copy is reserved for a genuinely empty event.

One note: I couldn't add a full-render assertion for it — rendering an *empty* panel list page to HTML in the test harness trips a pre-existing `VariableDoesNotExist` cascade in `components/avatar.html` (optional `danger_ring`/`avatar_label`/`text_size` under the dev `string_if_invalid="ERROR: Missing variable %s"` setting). Every existing render-test seeds visible rows, so it's never been exercised. Happy to fix `avatar.html` to default those vars in a separate PR if you want empty panel pages to be render-testable.

---

PR #546, opened 2026-07-07
feat(web): informative themed error pages + no silent redirect on missing events

hasparus:
## What & why

The app's themed, randomized 404/500 pages are part of the brand's personality, so this PR **keeps them** — the D&D / cyberpunk / Cthulhu flavor stays, still selected per-request via `secrets.randbelow`. What changes is that the pages are now genuinely **informative**: alongside the random flavor text they always render a stable, plain-language guidance line, and the 500 page surfaces the support contact.

### 404 (`custom_404`)
- Restores the full themed randomized message lists (title / message / subtitle / icon).
- Adds an always-present guidance line: *"The page you're looking for doesn't exist or may have moved."*
- Keeps the existing Back / Home actions.

### 500 (`custom_500`)
- Restores the full themed randomized message lists.
- Adds an always-present guidance line: *"Something broke on our end. Try again in a moment."*
- Adds a support-contact line rendering `settings.SUPPORT_EMAIL` as a `mailto:` link.
- Stays dependency-light — it only reads `settings` (no services/DB), so it still renders when the app is broken. The template remains standalone (no `{% extends %}`).

### Missing / unpublished events (A3, refs #543)
- **Retained** from the earlier revision: a 404 for a missing *or* unpublished event no longer silently drops the visitor on the home page. It redirects to the events list with a neutral flash — *"That event isn't available."* — so a 404 never betrays whether an unannounced event exists.

## Tests
Reworked to assert structural, non-flaky things instead of exact random flavor text: correct status codes, presence and stability of the guidance line, `SUPPORT_EMAIL` present on the 500 context, and the A2/A3 recovery behavior (canonical redirect on junk links, neutral redirect + flash for missing/unpublished events). All error-view tests and every `assert_response_404` consumer pass.

## i18n
New Polish strings added and translated; catalog extraction is current with no fuzzy/untranslated entries.

Refs #543 (A2, A3).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved 404 and 500 error pages with clearer guidance and updated styling.
  * Missing or unavailable events now send people to the events list with a friendly notice instead of the home page.
  * The 500 error page now shows a support contact link for easier help.
  * Polish translations were added for the updated error and support messages.

### DIFF
--- a/src/ludamus/locale/pl/LC_MESSAGES/django.po
+++ b/src/ludamus/locale/pl/LC_MESSAGES/django.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2026-07-07 14:57+0200\n"
+"POT-Creation-Date: 2026-07-07 18:44+0000\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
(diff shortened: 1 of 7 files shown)


src/ludamus/locale/pl/LC_MESSAGES/django.po
```diff
@@ -552,6 +560,10 @@ msgstr "Nyarlathotep opętał naszą infrastrukturę"
 msgid "Madness spreads through every circuit..."
 msgstr "Szaleństwo rozprzestrzenia się przez każdy obwód..."
 
+#: src/ludamus/adapters/web/django/error_views.py
+msgid "Something broke on our end. Try again in a moment."
+msgstr "Coś się zepsuło po naszej stronie. Spróbuj ponownie za chwilę."
```
hasparus: @claude maybe we write sth like "Our best people are on it." "Nasi najlepsi ludzie pracują nad tym."

hasparus:
Done in 3ad0afa — the 500 guidance now reads **"Our best people are on it."** / **"Nasi najlepsi ludzie pracują nad tym."** Actionable next-steps stay via the support mailto + Back/Home below, so the themed page keeps both the personality and a way out. Updated the three assertions and the PL catalog; error-view tests green.

---

PR #564, opened 2026-07-10
Self-heal sandbox tool installs and Playwright browsers on session start

hasparus:
Fresh sandboxes came up with `mise run check` broken: the image pre-bakes
GitHub-layout installs of shellcheck/actionlint/hadolint, so mise skips
installing them through the mise.sandbox.toml [tool_alias] backends but
cannot list their bin paths — no shims, dropped from `mise run` task PATH,
and hk dies with "actionlint: not found". The hook now purges any aliased
tool without a shim after `mise install` and reinstalls it via the alias;
pre-baked layouts that satisfy the alias (hk) are left alone.

The e2e suite was also unusable: /opt/pw-browsers ships chromium build 1194
while @playwright/test 1.58.2 needs 1208, and `playwright install
--with-deps` dies on apt (ondrej/php PPA renamed its InRelease Label). Fall
back to a dependency-less `playwright install` (CDN is reachable through
the proxy; the image already ships Chromium's OS libs) and let apt tolerate
release-info changes.

Co-authored-by: hasparus <hasparus@gmail.com>

### DIFF
--- a/.claude/hooks/session-start.sh
+++ b/.claude/hooks/session-start.sh
@@ -60,7 +60,10 @@ fi
 if ! command -v python3.14 > /dev/null 2>&1 \
   || ! command -v pipx > /dev/null 2>&1; then
   export DEBIAN_FRONTEND=noninteractive
-  apt-get update -q > /dev/null || echo "WARN: apt-get update failed"
+  # --allow-releaseinfo-change: image PPAs occasionally change their metadata
+  # (e.g. ondrej/php renamed its Label), which otherwise fails the update.
+  apt-get update -q --allow-releaseinfo-change > /dev/null \
+    || echo "WARN: apt-get update failed"
   apt-get install -y -q python3.14 python3.14-venv pipx > /dev/null \
     || echo "WARN: apt-get install python3.14/pipx failed; Python tooling may be unavailable"
 fi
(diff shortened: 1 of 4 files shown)


hasparus:

The change is well-scoped, correctly placed (purge/reinstall runs after `mise install` but before the `mise run` tasks that depend on shim PATH entries), properly quoted, and idempotent on a healthy second run (once alias-backed shims exist, `[ ! -e shims/$tool ]` is false for all tools, so nothing is purged and no extra `mise install` fires). The sed range correctly extracts only the four `[tool_alias]` keys and stops at `[tools]`, and `$tool` is regex-constrained to `[A-Za-z0-9_-]`, so the `rm -rf` has no injection/traversal exposure. The findings below are hardening and layering concerns, not correctness defects in the happy path.

### 1. Unguarded `rm -rf` under `set -e` can abort the whole hook (**follow-up**, borderline blocker)

`.claude/hooks/session-start.sh:83` — inside the `while` loop, `rm -rf "$HOME/.local/share/mise/installs/$tool"` is a bare statement, not `|| echo`-guarded. The script runs under `set -euo pipefail` (line 2), and this file's stated philosophy (lines 14-17, 68-70) is that *every* mutation step must warn-and-continue so a single failure never aborts session setup. This is the only new command that violates that invariant. `rm -rf` rarely fails, but when it does (permission, busy file, I/O error) it aborts the async hook and silently skips everything downstream — `mise bootstrap`, `hk install`, and the Playwright step never run.

Remedy: `rm -rf "..." || echo "WARN: failed to purge $tool"` — matching the discipline of every other line in the file.

### 2. Purge-succeeds-but-reinstall-fails leaves the tool fully gone for the session (**follow-up**)

`.claude/hooks/session-start.sh:94-97` — a tool with no shim is `rm -rf`'d first, then one `mise install` retry re-provisions all purged tools. If that retry fails, the affected tools are *completely absent* — strictly worse than pre-purge, where they at least resolved via `mise exec`. It self-heals next session (empty install dir → alias-backend install), so transient. Worth one comment line noting the purge is intentionally destructive-and-non-atomic, so a future reader doesn't mistake it for a bug.

### 3. Playwright fallback leaks e2e-runner knowledge into the hook (**follow-up**)

`.claude/hooks/session-start.sh:112` — the fallback duplicates the runner (`aube exec`), the working dir (`-C tests/e2e`), and the command from `test:e2e:install` (`mise.toml:213`) minus `--with-deps`. If the task's e2e layout changes, this fallback silently diverges. The graceful degradation belongs inside the task: have `test:e2e:install` try `--with-deps` and fall back internally, leaving the hook to call `mise run test:e2e:install`. (The fallback also skips the task's first `aube install` step — fine for the stated scenario, but reinforcing that the task is the right home.)

### 4. The self-heal hand-parses TOML with three implicit couplings (**follow-up**)

`.claude/hooks/session-start.sh:88-90` — assumes every `[tool_alias]` key is (a) regex-parseable without quotes/dots/comments, (b) equal to its shim basename, and (c) equal to its install-dir name. All hold for the current four tools, but `docs/agents/sandbox.md:26` invites new aliases — and a key needing quoting (a full backend spec, like the `dockerfmt` case) would be silently dropped from the self-heal. Prefer a native mise enumeration if one exists; failing that, add a comment tying the parser's assumptions to sandbox.md's "add a matching alias" instruction so the coupling is discoverable when it breaks.

---

No structural regressions, no file-size or spaghetti concerns — the new block is a cohesive, well-commented unit and the docs/PAPERCUTS updates describe it accurately. Finding #1 is the only one worth addressing before merge (a two-word change); the rest can land separately.

**VERDICT: APPROVE**

---

PR #579, opened 2026-07-13
docs(models): note superset-safe contract of conflicted_user_ids

hasparus:
## Summary

Reviewer follow-up from PR #572: `SessionManager.conflicted_user_ids`
(`src/ludamus/adapters/db/django/models.py`) returns
`set(...values_list("session_participations__user_id", flat=True))` after a
single `.filter()` that constrains user+status on the same participation row.
The reviewer flagged that its correctness relies on subtle multi-valued-join
semantics, and without a comment a future maintainer could "optimize" it into
a broken form. This PR adds a two-line comment stating the exact contract:

```python
        # Superset-safe: never misses a genuine conflict; extra ids the join might
        # return are harmless since callers only probe membership of their own user_ids.
        return set(
```

No logic changes — comment only.

## tingle offset

`models.py` falls in the `legacy-loc` line-count range, so the +2 comment
lines needed an honest offset. I tightened two nearby comments in the same
file (same wording, fewer lines, no information lost):

- `AliveManager.get_queryset` comment: 3 lines → 2
- `User.claim_token` comment: 3 lines → 2

Net diff on the file: 6 insertions / 6 deletions (line count unchanged), so
`legacy-loc` is flat and `tingle check` exits 0.

## Test plan

- [x] `MISE_ENV=sandbox mise exec -- poetry run tingle check` → exit 0 (baseline refreshed via `git branch -f main origin/main`)
- [x] `mise run check` → exit 0 (ruff, pylint 10.00/10, black unchanged, impeccable layering, djlint, etc. all green)

---
_Generated by [Claude Code](https://claude.ai/code/session_01BsdvWh3G9h76W5egZ73iu9)_

### DIFF
--- a/src/ludamus/adapters/db/django/models.py
+++ b/src/ludamus/adapters/db/django/models.py
@@ -45,9 +45,8 @@ _SoftDeleteT = TypeVar("_SoftDeleteT", bound=models.Model)
 
 
 class AliveManager(models.Manager[_SoftDeleteT]):
-    # The default `objects` manager hides soft-deleted rows so every existing
-    # read (including reverse relations like `category.sessions`) excludes them
-    # automatically. Reach soft-deleted rows through `all_objects`.
+    # Default `objects` hides soft-deleted rows from every read, including reverse
+    # relations like `category.sessions`; use `all_objects` to reach them.
     def get_queryset(self) -> models.QuerySet[_SoftDeleteT]:
         return super().get_queryset().filter(deleted_at__isnull=True)
 
@@ -125,9 +124,8 @@ class User(AbstractBaseUser, PermissionsMixin):
         default=False,
         help_text=_("Use Gravatar instead of provider avatar"),
     )
-    # Single-use handle that lets the intended person sign in and take over a
-    # managed (connected) row as their own account. Mirrors the waitlist-offer
-    # claim_token pattern. Empty for active accounts.
+    # Single-use handle for the intended person to claim a managed (connected) row.
+    # Mirrors the waitlist-offer claim_token pattern; empty for active accounts.
     claim_token = models.CharField(max_length=64, blank=True, default="", db_index=True)
     shadowbanned: models.ManyToManyField[User, Shadowban] = models.ManyToManyField(
         "self",
@@ -768,6 +766,8 @@ class SessionManager(AliveManager["Session"]):
             return set()
         start = session.agenda_item.start_time
         end = session.agenda_item.end_time
+        # Superset-safe: never misses a genuine conflict; extra ids the join might
+        # return are harmless since callers only probe membership of their own user_ids.
         return set(
             self.get_queryset()
             .filter(


hasparus:

Scope: comments-only change to `src/ludamus/adapters/db/django/models.py` (6 insertions / 6 deletions, 1 file). No logic touched. I verified the three things that matter for a comment-only PR: is the new comment *true*, did the two tightened comments lose information, and does the `legacy-loc` ratchet stay flat.

**Finding 1 — New `conflicted_user_ids` comment is accurate. [clean]**
Traced the query and every caller. The single `.filter(session_participations__user_id__in=user_ids, session_participations__status=CONFIRMED, ...)` constrains user + status on the *same* joined participation row (same `.filter()` call), and `.values_list("session_participations__user_id")` reuses that constrained join — so the result is exactly `{input user_ids that are CONFIRMED in an overlapping, non-excluded, alive same-event session}`. "Never misses a genuine conflict" holds robustly: a genuinely-conflicting input user is a confirmed participant of a matched session, so their id is always emitted. The "extra ids are harmless" clause is a deliberately conservative hedge against future join-reuse assumptions, not a false statement — and it is verifiably safe for both real call sites:
- `has_conflicts` (models.py:787): `user.pk in conflicted_user_ids(session, [user.pk])` — probes only its own input id.
- `enrollment.py:121-123 / 155`: input is `[p.user_id for p in participations]`, and the only membership probe is `participation.user_id in conflicted` over those same participations.
Neither caller ever probes an id outside its own input set, so spurious extras (if the join semantics ever changed) could not cause a wrong answer. The comment matches the code and pins down exactly the subtle invariant a future "optimizer" would otherwise break. Good, well-placed comment.

**Finding 2 — Two tightened comments lost no load-bearing information. [clean]**
- `AliveManager.get_queryset` (~L48): before/after both state that `objects` hides soft-deleted rows from every read, keep the `category.sessions` reverse-relation example verbatim, and keep the `all_objects` escape hatch. Only the filler word "automatically" was dropped. No constraint or caveat lost.
- `User.claim_token` (~L128): before/after both keep "single-use handle", the "managed (connected) row" concept, the "mirrors the waitlist-offer claim_token pattern" cross-reference, and the "empty for active accounts" caveat. "sign in and take over ... as their own account" compressed to "claim" — acceptable synonym; the sign-in mechanism is an implementation detail, not a documented constraint. No material loss.

**Finding 3 — Ratchet holds. [clean]**
`git diff --stat`: single file, 6 insertions / 6 deletions → file line count net zero. Removals 3+3 (two tightened comments), additions 2+2 (same comments) + 2 (new comment) = 6/6. Nothing but comment lines changed; no code, imports, or signatures touched. `legacy-loc` stays flat as claimed.

No structural regression, no spaghetti, no boundary leak, no file-size concern — and none are possible in a comment-only diff. The comment earns its keep: it documents a genuinely non-obvious ORM invariant at the exact site where a future refactor would be tempted to break it. No blockers, no follow-ups.

VERDICT: APPROVE

---

PR #580, opened 2026-07-13
chore(tingle): fix mock-ANY mid-line and object-src false positives

hasparus:
## Summary

Two false-positive patterns in `tingle.toml`.

### `mock-ANY` (line ~95)

**Before:** `'(?!.*(view|form)).*\bANY\b.*'`
**After:** `'^(?!.*(?:\b|_)(view|form|import)\b).*\bANY\b'` with `flags = ["MULTILINE"]`

`re.finditer` tries every start offset in the text, so the negative lookahead could begin matching *after* "view"/"form" already appeared earlier on the line — e.g. `"view": ANY,` no longer had "view" ahead of the match start, so the lookahead saw nothing to reject and the line got counted. It also counted the `from unittest.mock import ANY` line, which isn't a mock use at all.

Anchoring with `^` forces the match to start at the beginning of a line; `MULTILINE` is needed because tingle's non-diff `regex_count` matches over the whole file's text (with embedded `\n`), where plain `^` only anchors to offset 0 without that flag. tingle's diff mode (`regex_count_diff`, used by `tingle check`) already splits per line before matching, so it works with or without the flag there — but stat/report without `--diff` need it.

The exemption tokens are bounded with `(?:\b|_)(view|form|import)\b` rather than a plain `\b...\b` pair: a bare alternation would exempt any line containing the substrings (`preview`, `important`, `uniform`), while `\bform\b` alone would wrongly *count* the 11 sanctioned `"*_form": ANY` context keys — the underscore in `design_form` is a word character, so it kills the leading boundary. The `_` alternative keeps `_view`/`_form` suffixes exempt while `preview`/`important` now count.

I decided to exempt `import` lines along with `view`/`form`, since an import statement isn't a mock use — happy to revert that if the maintainer prefers imports stay counted.

### `type-object` (line ~150)

**Before:** `'(?<[\.\w])object\b'`
**After:** `'(?<[\.\w])object\b(?!-)'`

Matched inside the CSP directive string `"object-src"` in `src/ludamus/edges/settings.py:319`. A full string-literal exclusion isn't practical in regex, so this adds a lookahead refusing a following hyphen — cheap and sufficient for this one false positive.

## Verification

Confirmed tingle uses Python's stdlib `re` (via `tingle/mills/metrics/regex_count.py`), which supports both lookahead and lookbehind, so both patterns are valid.

`tingle stat` before/after (full-repo, non-diff counts):

| metric | before | after |
|---|---|---|
| mock-ANY | 335 | 42 |
| type-object | 112 | 111 |

The word-boundary tightening in the second commit left the mock-ANY count at 42 — nothing newly counted, nothing newly exempted on the current tree (verified by diffing per-line matches of both patterns across `tests/`).

Confirmed `src/ludamus/edges/settings.py:319` (`"object-src"`) no longer appears in `tingle report --metric type-object`; line 394 (`dict[str, object]`) still does.

Scratch tests (added, measured, then reverted — `git status` showed only `tingle.toml` at commit time):
- Appended `x_scratch_any_test = ANY` to a test file → mock-ANY count went 42 → 43, then reverted.
- After the boundary tightening: appended `x = ANY` and `{"preview": ANY}` → 42 → 44 (both the naked use and the previously substring-exempted `preview` line now count), then reverted.
- Appended `x_scratch_object_test: object = None` to `settings.py` → type-object count went 111 → 112, then reverted.

Branch-vs-main check (config changes move both sides of the diff equally):
```
$ git branch -f main origin/main
$ MISE_ENV=sandbox mise exec -- poetry run tingle check
# exit 0, no output
```

Also ran `mise exec -- taplo format --config=.taplo.toml` (no diff produced) and `MISE_ENV=sandbox mise run check` (exit 0).

## Test plan

- [x] `tingle stat --metric mock-ANY` / `--metric type-object` before/after comparison
- [x] Scratch add-then-revert tests proving both patterns still catch the naked case
- [x] Substring scratch test proving `"preview": ANY` is counted, not exempted
- [x] `MISE_ENV=sandbox mise exec -- poetry run tingle check` → exit 0
- [x] `mise exec -- taplo format --config=.taplo.toml` → no changes
- [x] `MISE_ENV=sandbox mise run check` → exit 0

### DIFF
--- a/tingle.toml
+++ b/tingle.toml
@@ -95,7 +95,8 @@ ranges = ["python-tests"]
 name = "mock-ANY"
 type = "regex_count"
 group = "testing"
-pattern = '(?!.*(view|form)).*\bANY\b.*'
+pattern = '^(?!.*(view|form|import)).*\bANY\b'
+flags = ["MULTILINE"]
 ranges = ["python-tests"]
 
 
@@ -149,7 +150,7 @@ ranges = ["python-src"]
 name = "type-object"
 type = "regex_count"
 group = "typing"
-pattern = '(?<![\.\w])object\b'
+pattern = '(?<![\.\w])object\b(?!-)'
 ranges = ["python-src"]
 
 [[metrics]]
--- a/tingle.toml
+++ b/tingle.toml
@@ -95,7 +95,7 @@ ranges = ["python-tests"]
 name = "mock-ANY"
 type = "regex_count"
 group = "testing"
-pattern = '^(?!.*(view|form|import)).*\bANY\b'
+pattern = '^(?!.*(?:\b|_)(view|form|import)\b).*\bANY\b'
 flags = ["MULTILINE"]
 ranges = ["python-tests"]
 


hasparus:

Config-only PR to `tingle.toml` tightening two false-positive ratchet patterns. I verified the mechanism against the installed tingle source (`.venv/.../tingle/mills/metrics/regex_count.py`, `registry.py`, `config.py`) and re-derived the metric numbers independently by grepping the tree. Both claims hold up. Findings:

**1. The `flags` key is real and fully consumed — not a silent no-op. [verified]**
`registry.py` declares `regex_count` params as `required=("pattern",), optional=("flags",)`, so `flags` is a known param. `validate_params` (regex_count.py) checks each flag against `_FLAGS = {IGNORECASE, MULTILINE, DOTALL}` — `MULTILINE` is valid — and `_compile` ORs them into `re.compile(pattern, flags)`. Worth stressing: config validation is *strict* — `mills/config.py._validate_params` raises `ConfigError` on any unknown param, so a bogus flag key would hard-fail the load, not pass silently. `MULTILINE` is properly wired.

**2. MULTILINE is load-bearing here, and the `^` anchor is the actual fix. [verified]**
The `stat`/full-repo `regex_count` reads whole-file text and runs `pattern.finditer(text)`. Without `MULTILINE`, `^` only matches at file offset 0, so the anchored pattern would match at most the first line of each file and the metric would collapse to near-zero — broken, not just wrong. `MULTILINE` makes `^` anchor every line, which is required. The real correctness gain over the old `(?!.*(view|form)).*\bANY\b.*` is the anchor: the old pattern was unanchored, so `finditer` could *start* a match past an earlier `view`/`form` token (e.g. `"view": call(ANY)`), defeating the lookahead. Anchoring at `^` forces the negative lookahead to scan from line start. This is the mechanism behind the 335→42 drop. Diff-mode (`regex_count_diff`) already matches line-by-line with terminators stripped, so `^` anchors each stripped line there too — stat total and diff delta stay consistent. Good, coherent design.

**3. The 335→42 drop is exact, and 42 has teeth. [verified]** Independent grep over `tests/`: 335 lines contain `\bANY\b`; 293 also contain `view|form|import`; **42** remain — an exact match to the claimed post value. The residual 42 are genuine un-anchored uses (`context_data=ANY`, `state=ANY`, `"event": ANY`), so the ratchet still bites.

**4. Exempting `import` lines is sound. [verified]** `from unittest.mock import ANY` is exempted, but ruff F401 forbids an unused import and any actual *use* site (`x = ANY`) is still counted — no blind spot introduced.

**5. `type-object` `(?!-)` is precisely scoped: 112→111. [verified]** In `src/**/*.py` the only `object-` occurrence is the CSP string `"object-src": [CSP.NONE]` at `src/ludamus/edges/settings.py:319`; the other `object-cover`/`object-contain` hits are in `.html` templates, outside the `python-src` range. `dict[str, object]` at settings.py:394 is followed by `]`, so it stays counted. The exclusion masks exactly one line. A stray `object - 1` keeps a space before the hyphen and is unaffected. Tight.

**6. Diff is config-only. [verified]** Only `tingle.toml`, two `[[metrics]]` entries. No source, no spaghetti, no file-size concern.

---

**[follow-up, non-blocking]** Both lookaheads use *substring* alternation, not word-bounded. `(view|form|import)` will also swallow a naked ANY on lines containing `preview`/`review`/`interview`, `information`/`transform`/`platform`/`formatter`/`performance`, or `important`/`importer`. The pre-existing pattern already had this for `view|form`; adding `import` widens the escape hatch slightly. For a ratchet meant to catch *un-sanctioned* ANY, `\b(view|form|import)\b` (or targeting the real sanctioned contexts) would be more precise and less likely to under-count in future. Coarse ratchet under a `sum` policy tolerates the noise, so this is a refinement, not a merge blocker.

No structural regression, no missed simplification (a config-only two-line ratchet tune has no code-judo move to find), no boundary or file-size issues. The change does exactly what it claims, and the numbers reproduce exactly.

VERDICT: APPROVE

---

PR #578, opened 2026-07-13
docs(plans): add plan 019 — CSP tightening and gated enforcement

hasparus:
## Summary

Adds `plans/019-csp-enforcement-path.md`, an executor-grade plan (no
source-code changes) for tightening the report-only CSP shipped by
plan 007 and eventually flipping it to enforcement. Also adds plan
019's row to `plans/README.md`'s status table.

- **Step Group A (executable now)** — register Django's built-in
  `django.template.context_processors.csp` context processor, add
  `nonce="{{ csp_nonce }}"` to all 14 inline `<script>` tags (12
  files, own grep verified), and switch `script-src` from
  `'unsafe-inline'` to the `CSP.NONCE` sentinel in
  `CSP_REPORT_ONLY_POLICY`. Cites the actual mechanism from the
  installed Django 6.0.7 source
  (`django/middleware/csp.py`'s `LazyNonce` +
  `django/template/context_processors.py`'s `csp` processor).
- **Step Group B (executable now)** — eliminate all 11 `hx-on:`
  attributes (3 files: `checkbox-field.html`,
  `timetable-session-detail.html`, `panel/base.html`) by converting
  them to real `addEventListener` calls, then drop `'unsafe-eval'`
  from `script-src`.
- **Step Group C (GATED)** — a `report-uri`/`report-to` decision note
  and the `SECURE_CSP` enforcement flip. Explicitly gated behind the
  maintainer reviewing production report-only violation data, which
  does not exist yet — spelled out as a hard STOP condition so an
  executor cannot run this step without explicit authorization.

## Test plan

- [x] `mise exec -- markdownlint-cli2 "plans/*.md"` → 0 errors
- [x] Every `file:line` excerpt in the plan re-verified against live
      code (settings.py CSP block, TEMPLATES context_processors,
      inline `<script>` grep, `hx-on:` grep, Django CSP source) before
      committing
- [x] Pre-commit hooks (markdownlint, codespell, varlock, etc.) passed
      on commit

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01BsdvWh3G9h76W5egZ73iu9

---
_Generated by [Claude Code](https://claude.ai/code/session_01BsdvWh3G9h76W5egZ73iu9)_

PR #583, opened 2026-07-13
CSP: enforcing nonce-based Content-Security-Policy (plan 019)

hasparus:
## What &amp; why

This PR completes **plan 019** end to end: it moves the site from a
report-only Content-Security-Policy to an **enforced, nonce-based** one, and
does the app-side work that flip requires. It folds in the earlier
plan-doc-only PR (#578, now closed) so there is a single reviewable change.

**Why we're doing this.** CSP is the browser-level backstop that turns an
HTML/attribute-injection bug into a *blocked resource* instead of *executed
attacker JavaScript*. Today the only XSS controls are `nh3` sanitization and
Django autoescaping — one gap in any `mark_safe` sink and there is no second
layer. CSP only earns its keep when the policy is actually **restrictive**: a
policy carrying `'unsafe-inline'`/`'unsafe-eval'` on `script-src` permits
exactly the injected-script execution it is meant to stop, so report-only-with-
unsafe-inline protects almost nothing. This PR removes both:

- `'unsafe-eval'` — only htmx `hx-on:` needed it (htmx compiles those with
  `new Function`). The 11 handlers became a delegated `panel-chrome.ts` module
  and htmx's eval is disabled (`allowEval:false`), so nothing needs it.
- `'unsafe-inline'` on scripts — replaced by **per-request nonces**. Every
  legitimate inline `&lt;script&gt;` carries a nonce; an injected one can't guess it,
  so it doesn't run.

`style-src` intentionally **keeps** `'unsafe-inline'` — inline `style="…"`
attributes are pervasive and can't take a nonce (see "Next step"). Scripts are
where code execution lives, so that's where the win is.

## What's in this PR

- **hx-on → delegated listeners** (`panel-chrome.ts`): all 11 `hx-on:`
  attributes replaced by `data-action`/`data-refresh-*` handlers, behavior
  verified equivalent and browser-tested; `allowEval:false` makes any missed
  handler fail loudly rather than silently.
- **Script nonces**: Django's `csp` context processor registered; every inline
  `&lt;script&gt;` (14 across 12 templates) carries `nonce="{{ csp_nonce }}"`;
  `script-src` switched from `'unsafe-inline'` → `'nonce-…'`.
- **Enforcement flip**: production sends `Content-Security-Policy` (enforcing).
- **Component cleanup**: `checkbox-field.html` drops `hx_on_change` for
  `data_action`/`data_required_target`; shared htmx type moved to `htmx.d.ts`.
- **Tests**: `test_security_headers.py` reworked for the enforcing header,
  asserting the header nonce matches the nonce rendered in the page body (and a
  500-page variant through the real error path); a Playwright spec
  (`csp-violations.spec.ts`) serves the **real enforcing header** and sweeps
  the public pages + panel for `securitypolicyviolation` events, asserting
  zero.

## Implementation notes (verified)

- **Nonce guard.** Django's documented `{% if csp_nonce %}` pattern has a
  footgun: `LazyNonce.__bool__` is False until the nonce is *generated*, and a
  bool check doesn't generate it — so guarding **every** script on `{% if
  csp_nonce %}` means nothing ever primes it and no script gets a nonce (the
  policy then blocks all inline scripts). We use `{% if csp_nonce is not None
  %}`: it always renders the attribute when the processor is in context, and
  the `{{ csp_nonce }}` interpolation forces generation; it still degrades to
  no attribute when `csp_nonce` is absent entirely. A test asserts the header
  nonce equals the body nonce so this can't silently regress.
- **500 page.** `500_dynamic.html` stays self-contained (doesn't extend
  `base.html`) so it renders even when something upstream broke. A regression
  test forces a real 500 through Django's exception-handling middleware and
  confirms its script nonce matches the header.
- **Google Fonts.** Enforcing (not report-only) surfaced that `index.css`
  `@import`s the Outfit font from `fonts.googleapis.com` (stylesheet) and
  `fonts.gstatic.com` (font files); both were added to `style-src`/`font-src`.
  Report-only never showed this because nothing blocks under report-only —
  exactly why the e2e runs the *enforcing* header.
- **e2e enablement.** A new `ENABLE_CSP` flag decouples `SECURE_CSP` from
  `IS_PRODUCTION` so the e2e server can send the enforcing header without the
  rest of prod hardening (HTTPS redirect, secure cookies) that would break a
  plain-HTTP server. It defaults False everywhere else, so the "no CSP by
  default" regression guard is unaffected. The `csp-violations` spec runs on
  Chromium; it's excluded from the Firefox project because Firefox doesn't
  reliably use Playwright's proxy config that this sandbox's egress proxy
  needs (a container artifact, not a CSP/app bug — Chromium passes clean).

## Risks

1. **A missed inline script breaks silently under enforcement** — no error,
   just a dead feature. *Mitigation:* all 14 inline scripts nonce'd; the
   header-nonce-matches-body-nonce test; the e2e violation sweep over the major
   surfaces. *Residual:* e2e can't cover every page/state — the long tail is
   what a report-only production period would have caught from real traffic
   (see rollout).
2. **`connect-src 'self'`** would block any future cross-origin fetch/XHR/WS
   not added to the policy; `img-src https:` stays broad for avatars/media.
3. **No reporting endpoint (accepted).** No `report-uri`/`report-to`, so
   enforced violations in prod surface only via user reports or the browser
   console, not telemetry. Wiring a collector is a documented follow-up.
4. **Firefox CSP coverage** is skipped in CI too (not just the sandbox) because
   the exclusion is unconditional — a small follow-up could gate it on the
   proxy being present so CI keeps cross-browser CSP coverage.

## Rollout &amp; human test plan

The flip reaches **staging on push** (auto deploy) and **production only on
merge** — staging is the validation gate that stands in for the plan's
original "wait for production report-only data."

**On staging, before merging:**

1. Confirm the Playwright `csp-violations` spec is green in CI (zero violations
   across the swept pages).
2. **DevTools → Console**, click through each surface watching for `Refused to
   execute inline script … Content Security Policy`:
   - Public: event list, an event detail (+ open a session modal), a session
     enroll page, the notice board (index + a detail), a flatpage.
   - Auth: log in; panel dashboard, then the panel pages that had inline
     scripts — **cfp-edit**, **proposal-edit**, **session-field-create**,
     **personal-data-field-create**, **facilitator-merge**.
   - Panel chrome (the hx-on migration): sidebar open/close, fold (reload →
     persists), event switcher (navigates), category collapse (reload →
     persists), a timetable session-detail confirm/undo (left pane refreshes),
     the connections-edit "replace secret" checkbox (secret field's required
     state toggles).
   - Theme: toggle light/dark — the render-blocking theme-init script must run
     (a nonce failure there flashes the wrong theme on load).
   - Force a **500** and confirm the error page renders and behaves.
3. Confirm the header: `curl -sI https://&lt;staging&gt;/ | grep -i
   content-security-policy` → `Content-Security-Policy:` (not `-Report-Only`),
   `script-src 'self' 'nonce-…'`, no `unsafe-inline`/`unsafe-eval` on scripts.

**If anything is blocked:** one-line revert (`SECURE_CSP` → report-only) keeps
the nonce work while falling back to non-blocking. Prefer that over re-adding
`'unsafe-inline'`.

**After merge:** watch production console/error reports for the first day — the
untested long tail (rare pages, error states) is where a missed nonce surfaces.

## Next step: tightening `style-src` (deliberately deferred)

This PR nonces **scripts** and drops `'unsafe-inline'` from `script-src`;
`style-src` still carries `'unsafe-inline'`. That's the remaining
CSP-tightening work, and it's genuinely harder:

- Inline `&lt;style&gt;` blocks nonce trivially, same mechanism.
- **Inline `style="…"` attributes are the blocker** — CSP nonces apply to
  `&lt;style&gt;`/`&lt;script&gt;` *elements*, not to `style` attributes. Dropping
  `'unsafe-inline'` from `style-src` requires either `'unsafe-hashes'` + a
  SHA-256 per attribute value (brittle — every value change is a policy
  change), or removing the attributes: static → utility classes, and the
  genuinely dynamic ones (computed positions/sizes — timetable grid placement,
  meter widths, cover-image backgrounds) → CSS custom properties driven via a
  small nonce'd `&lt;style&gt;` block. Ship it report-only first, like this one,
  because the dynamic-style long tail is wide.
- Also deferred: a `report-uri`/`report-to` endpoint so enforced violations
  are collected as telemetry.

(Both are recorded in `plans/019-csp-enforcement-path.md` → "Deferred / next
step".)

## Summary by CodeRabbit

* **New Features**
  * Enforced Content Security Policy with per-request nonces and disabled unsafe script execution.
  * Added secure interactions for navigation, panels, forms, accessibility controls, and theme handling.
  * Added automatic panel refreshes after successful actions and automatic filter submission.
  * Improved protection against duplicate form submissions.
  * Added proxy support for end-to-end browser testing.

* **Bug Fixes**
  * Preserved inline scripts and interactive features under stricter security headers.

* **Tests**
  * Added coverage for CSP headers, nonce matching, security violations, and key user flows.


hasparus:

I read the whole diff, both sides of every touched template, the target URL config, `copy.ts`/`timetable.ts` for convention, and the vendored htmx 2.0.8. This is a genuinely clean behavior-preserving refactor. Below is the equivalence audit for all 11 `hx-on` bodies, delegation soundness, and the two follow-ups worth doing.

### Equivalence: all 11 conversions verified equivalent

- **Overlay `toggle-sidebar`** (`panel/base.html:134` → `panel-chrome.ts:12-15`): original used `this.classList.toggle('hidden')` where `this` is `#sidebarOverlay`; `toggleSidebar()` targets `getElementById("sidebarOverlay")` — same element. Exactly equivalent. Clean bill.
- **Mobile menu `toggle-sidebar`** (`panel/base.html:333`): original already referenced `#sidebar` + `#sidebarOverlay` by id — identical to the shared function. Clean bill.
- **`toggle-fold`** (`panel/base.html:150` → `panel-chrome.ts:17-27`): same `data-folded` toggleAttribute, same `panel-sidebar-folded` key, same `f?'1':'0'`, same try/catch swallow. Clean bill.
- **`toggle-category` ×4** (`panel/base.html:195/241/280/307` → `panel-chrome.ts:29-38`): same `closest('[data-cat]').dataset.cat`, same `catc-<cat>` class toggle, same `panel-cat-<cat>` key. The port adds a `if (!cat) return` guard the inline version lacked — behavior-preserving under real markup (button always sits inside `[data-cat]`), strictly safer. Clean bill.
- **`switch-event`** (`panel/base.html:164` → `panel-chrome.ts:40-43`): this is the one place logic actually changed shape, so I checked it hard. Original: `window.location.href = '/panel/event/' + this.value + '/'` with `this.value` = `event.slug`. New: navigate to `selectedOptions[0].dataset.url` = `{% url 'panel:event-index' slug=event.slug %}`. `panel/urls.py:104` defines `event/<slug:slug>/` for `event-index`, which reverses to `/panel/event/<slug>/` — byte-identical to the hand-built string. `location.assign(url)` vs setting `location.href` are equivalent navigations. **Equivalent, and strictly better** (reverse() instead of a hardcoded path). Clean bill.
- **`sync-expanded-required`** (`components/checkbox-field.html` via `connections/edit.html:31` → `panel-chrome.ts:45-52`): reproduces the original branching exactly — `aria-expanded` = checked, then look up `id_secret` (now via `data-required-target`), and only add/remove `aria-required` when the field exists. `connections/edit.html` extends `panel/base.html` so the delegated listener is loaded on that page. Clean bill.
- **Timetable `htmx::after-request` ×2** (`timetable-session-detail.html:63/73` → `panel-chrome.ts:88-96`): same guard (`detail.successful`), same `htmx.ajax("GET", url, {swap:"outerHTML", target:"#left-pane"})`. Clean bill.

### Delegation soundness — sound

- click/change listeners resolve `.closest("[data-action]")` off `e.target` and dispatch on `dataset.action`; no path can double-fire (each control has exactly one action). `change` correctly serves both the `<select>` and the checkbox.
- `htmx:afterRequest` is bound to `document.body` and filters via `closest("[data-refresh-url]")`. No loop: the handler-issued `htmx.ajax` GET defaults its source element to `document.body`, so its own `afterRequest` resolves `closest("[data-refresh-url]") === null` and does not re-fire — matching the original's per-form scoping. `document.body` vs `document` is immaterial here since every form lives under body. No leak, no miss.

### allowEval capstone — valid and load-bearing

Confirmed the vendored htmx 2.0.8 (`static/vendor/htmx.min.js`) honors a `meta[name="htmx-config"]`, reads `allowEval` (default `true`), and raises `evalDisallowed` when eval is gated. So `<meta name="htmx-config" content='{"allowEval":false}'>` (`base.html:11`) genuinely disables htmx's `Function`-based eval, meaning any *missed* `hx-on` would now fail loudly (`htmx:evalDisallowedError`) instead of silently doing nothing. Correct capstone; it makes the CSP change safe to land.

### Findings

1. **follow-up — the security test does not actually guard the removal.** `tests/integration/web/test_security_headers.py:26` asserts `"script-src 'self' 'unsafe-inline'" in header`. That substring is *also* present in `script-src 'self' 'unsafe-inline' 'unsafe-eval'`, so the test would still pass if `'unsafe-eval'` came back. For a PR whose whole point is dropping `'unsafe-eval'`, the guard is toothless. Add `assert "unsafe-eval" not in header` next to it. Low effort, real coverage.

2. **follow-up (minor) — duplicated htmx type decl.** `declare const htmx: { ajax: ... }` in `panel-chrome.ts:8-10` is a verbatim copy of `timetable.ts:21-23`. It matches the existing convention (so not a blocker), but a one-line shared `htmx.d.ts` ambient decl would kill the drift risk if the `ajax` signature ever grows. Optional.

3. **clean bill — structure.** `panel-chrome.ts` is 96 lines, one delegated listener per event type, small pure helpers, comment header modeled on `copy.ts` — reads as inevitable. No file approaches 1k lines. The `settings.py:307-315` comment accurately describes the new state, and enforcement is deliberately untouched (only `SECURE_CSP_REPORT_ONLY` is wired at `:364`; no `SECURE_CSP`), consistent with the stated scope. No spaghetti, no leaked feature logic, no gratuitous abstraction.

Net: equivalence holds for all 11 bodies, the delegation is correct, and the `allowEval:false` capstone turns silent failure into loud failure. The only substantive gap is the test that doesn't assert what it claims to.

VERDICT: APPROVE

---

hasparus:
> The nonce work (Steps 1-3) and the SECURE_CSP enforcement flip (Step 8)

@claude Can we expand this PR to tackle the whole plan 19? It will be easier to test for us this way. Also, are there any end-to-end tests we can add to ensure the affected code still works?

hasparus:
@claude address these findings too, please

> 1. **follow-up — the security test does not actually guard the removal.** `tests/integration/web/test_security_headers.py:26` asserts `"script-src 'self' 'unsafe-inline'" in header`. That substring is *also* present in `script-src 'self' 'unsafe-inline' 'unsafe-eval'`, so the test would still pass if `'unsafe-eval'` came back. For a PR whose whole point is dropping `'unsafe-eval'`, the guard is toothless. Add `assert "unsafe-eval" not in header` next to it. Low effort, real coverage.

> 2. **follow-up (minor) — duplicated htmx type decl.** `declare const htmx: { ajax: ... }` in `panel-chrome.ts:8-10` is a verbatim copy of `timetable.ts:21-23`. It matches the existing convention (so not a blocker), but a one-line shared `htmx.d.ts` ambient decl would kill the drift risk if the `ajax` signature ever grows. Optional.

> 3. **clean bill — structure.** `panel-chrome.ts` is 96 lines, one delegated listener per event type, small pure helpers, comment header modeled on `copy.ts` — reads as inevitable. No file approaches 1k lines. The `settings.py:307-315` comment accurately describes the new state, and enforcement is deliberately untouched (only `SECURE_CSP_REPORT_ONLY` is wired at `:364`; no `SECURE_CSP`), consistent with the stated scope. No spaghetti, no leaked feature logic, no gratuitous abstraction.

hasparus:
Plan 019 (CSP nonces + enforcement) is now fully implemented on this branch, on top of the hx-on migration/allowEval work that was already here.

**What landed (3 commits):**
1. **`feat(csp): nonce inline scripts and flip CSP to enforcing`** — registers the `csp` context processor, nonces all 14 inline `<script>` tags (`{% if csp_nonce is not None %} nonce="{{ csp_nonce }}"{% endif %}` — deliberately `is not None`, not a plain truthy check: Django's `LazyNonce` only materializes on a direct read, so a plain truthy guard on *every* script would never generate a nonce at all, silently dropping the sentinel and blocking every inline script under enforcement). `script-src` drops `'unsafe-inline'` for the nonce; `style-src` keeps `'unsafe-inline'` (out of scope, see below) plus `fonts.googleapis.com`/`fonts.gstatic.com`, which enforcing the policy end-to-end via e2e revealed was missing (report-only never caught it). Renames `CSP_REPORT_ONLY_POLICY`→`CSP_POLICY`, flips `SECURE_CSP_REPORT_ONLY`→`SECURE_CSP` in production. Folds in `plans/019-csp-enforcement-path.md` and marks it DONE in the README, with a new "Deferred / next step" note on style-src tightening.
2. **`refactor(client): dedupe the htmx ambient type into htmx.d.ts`** — Finding 2, `panel-chrome.ts`/`timetable.ts`'s duplicate `declare const htmx` moved to a shared `htmx.d.ts`.
3. **`test(e2e): prove the enforcing CSP doesn't break real pages`** — new `csp-violations.spec.ts` collects `securitypolicyviolation` events across the public event list, an event detail page, the panel dashboard (sidebar-fold/category-toggle via `panel-chrome.ts`), and `cfp-edit` (the densest page for inline scripts), asserting zero violations. `.env.e2e` sets `ENABLE_CSP=true` (new flag, production-only otherwise) so the e2e server sends the real enforcing header, not report-only.

**500_dynamic.html resolution:** it doesn't extend `base.html` (stays self-contained so it renders even if something upstream broke), so its inline script is its own page's "first" script and needs to independently force nonce generation. Verified end-to-end through the *real* middleware/exception-handling chain (not just a bare `RequestFactory` call) that the nonce in the header matches the nonce in the rendered body — new regression test `TestCSP500PageNonce`.

**e2e enforcement:** confirmed real (not report-only) — `ENABLE_CSP=true` in `.env.e2e`, verified by inspecting the actual response header. `csp-violations.spec.ts` passes 3/3 on Chromium with zero collected violations. It's scoped to Chromium only: Firefox in this sandboxed environment can't reliably reach `fonts.googleapis.com` through the mandated egress proxy even with `playwright.config.ts` now passing `HTTPS_PROXY` through (Chromium picks it up and passes cleanly; Firefox's request handling still hangs — verified with request-level tracing), which is an environment/proxy artifact, not a CSP or product bug. Matches the existing precedent of scoping other state-sensitive specs to Chromium.

All gates green: `mise run check`, `mise run test:py` (2885 passed), `tingle check` (against `main`), `mise run lint-client` + `build-frontend`.

---

hasparus:
**Thermo-nuclear code-quality review** (CSP delta — strict maintainability pass, requested by @hasparus — automated, fresh-context reviewer)

Scope: only the NEW CSP work (nonces, enforcement flip, tests, e2e). The hx-on migration portion was reviewed/approved separately and I did not re-litigate it. I verified the load-bearing nonce claim against Django's own source rather than trusting the diff.

---

**Finding 1 — Nonce-guard correctness: CLEAN BILL (the security claim holds).**
The `{% if csp_nonce is not None %}nonce="{{ csp_nonce }}"{% endif %}` pattern on all 14 inline scripts is correct, and the reasoning in the `base.html` comment (`base.html:44-55`) is accurate. Verified against Django source:
- `django/utils/csp.py:76-77` — `LazyNonce.__bool__` returns `self._wrapped is not empty`, i.e. **False until the nonce is actually generated (read)**.
- `django/template/context_processors.py:93-97` — `csp` processor returns `{"csp_nonce": get_nonce(request)}`, which is the `LazyNonce` when the middleware ran, else `None`.
- (a) A uniform plain `{% if csp_nonce %}` (bool) guard **would break the site under enforce**: the first script's bool check is `False` (not yet generated) → the `{{ csp_nonce }}` interpolation is skipped → nonce never materializes → in `build_policy` (`csp.py:103`) `... and nonce` is falsy → the `CSP.NONCE` sentinel is *dropped*, leaving `script-src 'self'` with no `unsafe-inline` → every inline script blocked. Classic chicken-and-egg; confirmed.
- (b) `is not None` sidesteps it: the `LazyNonce` object is never `None`, so the guard is always true → `{{ csp_nonce }}` runs → `__str__` forces generation and primes the value; the same primed `LazyNonce` is then reused for every later script *and* read by the middleware's `get_nonce`, so header nonce == body nonce. When the middleware is absent the processor yields `None` → guard false → no attribute rendered → degrades safely (this is exactly the `TestCustom500` / outside-middleware path).
- Cross-checked: all 14 inline `<script>` tags carry the guard; the only bare `<script src>` is the htmx vendor bundle served from `'self'` (`base.html:148`) — nothing was missed. `TestCSPNonce` independently asserts `body_nonces == {header nonce}`, so this is regression-guarded, not just argued.

**Finding 2 — Policy scope: CLEAN.** `settings.py:332-343` — `script-src [SELF, NONCE]` (no unsafe-inline, no unsafe-eval); `style-src` keeps `unsafe-inline` + `fonts.googleapis.com`; `font-src` adds `fonts.gstatic.com`; `object-src 'none'`, `frame-ancestors 'none'`, `base-uri`/`form-action` `'self'`, `img-src` intentionally broad for Auth0/gravatar/GCS. The `unsafe-eval` drop is genuinely earned — htmx `hx-on:` attributes were converted to delegated `data-action` listeners in `panel-chrome.ts` and the `htmx-config allowEval:false` meta tag (`base.html:11`) disables htmx's `Function` eval. Scope is right.

**Finding 3 — ENABLE_CSP decoupling: CLEAN.** `settings.py:352` `ENABLE_CSP = IS_PRODUCTION or env.bool("ENABLE_CSP", default=False)`, then `settings.py:407-408` `if ENABLE_CSP: SECURE_CSP = CSP_POLICY`. Prod still enforces; default-off everywhere else keeps `test_no_csp_headers_by_default`'s guard intact; the enforcing header is decoupled from the prod-only HTTPS/HSTS/secure-cookie hardening that would break the plain-HTTP e2e server. No leak path found — `SECURE_CSP` is only bound when the flag is on, and the tests that need it set `settings.SECURE_CSP` explicitly.

**Finding 4 — Tests: CLEAN (real middleware chain).** `_directive()` (`test_security_headers.py:14-17`) correctly isolates a single directive via `(?:^|; )name ([^;]*)`. `TestCSP500PageNonce` goes through the Django test **client** (full middleware stack, not a bare `RequestFactory`), monkeypatches `EventsPageView.get` to raise, sets `DEBUG=False` + `raise_request_exception=False`, so a real `handler500` renders `500_dynamic.html` via a deferred `TemplateResponse` through the CSP middleware and proves its nonce matches the header — exactly the self-contained-root edge case that `base.html` priming can't cover. Pinning `secrets.randbelow` for an exact context assertion is a nice touch. *(Optional follow-up: the header/body-nonce extract-and-compare block is duplicated between `TestCSPNonce` and `TestCSP500PageNonce`; a tiny `_assert_header_nonce_matches_body(response)` helper would DRY it. Not blocking — two occurrences, arguably clearer inline.)*

**Finding 5 — e2e spec: CLEAN, with one follow-up.** `csp-violations.spec.ts` installs the `securitypolicyviolation` collector via `addInitScript` (before first paint), drives real interactions (modal open, sidebar fold, category toggle, cfp-edit's 3-inline-script page + duration add), and asserts the collected list is empty — against the real enforcing header (`ENABLE_CSP='True'`), not report-only. Genuine enforcement validation.
- **follow-up:** the Firefox `testIgnore` for `csp-violations.spec.ts` (`playwright.config.ts:102-111`) is **unconditional**, but the justification is a *proxied-sandbox-only* Firefox/Google-Fonts quirk (well documented, verified with request tracing). Real CI Firefox has normal egress and could run this. Consider gating the ignore on `process.env.HTTPS_PROXY` (you already compute `proxyServer` right above) so Firefox coverage isn't permanently forfeited outside the sandbox. Mild coverage gap, not a blocker.

**Finding 6 — Structural: CLEAN, one doc nit.**
- **follow-up (doc):** the `settings.py:348` comment says `ENABLE_CSP lets tests/e2e/.env.e2e opt in`, but the flag actually lives in the **root** `.env.e2e` (`playwright.config.ts` loads `path.join(repoRoot, ".env.e2e")`; there is no `tests/e2e/.env.e2e`). One-line comment fix.
- No file approaches 1k lines (`panel-chrome.ts` 92, test file 124, spec 121). `panel-chrome.ts` is the opposite of spaghetti: two document-level delegated dispatchers keyed on `[data-action]` plus one `htmx:afterRequest` refresh handler — this *deletes* the scattered inline `hx-on:` eval strings rather than rearranging them, and `switchEvent` reading a resolved `data-url` (`panel/base.html:163`) is a real improvement over the old `'/panel/event/' + this.value + '/'` string concat. Textbook code-judo: the CSP hardening and the chrome-JS cleanup are the same move.

---

No blockers. The load-bearing nonce approach is correct and independently regression-guarded; the two follow-ups (unconditional Firefox ignore, stale `.env.e2e` path in a comment) are non-blocking polish.

VERDICT: APPROVE

---

hasparus:
## Stale-PR maintenance pass

This PR had gone ~2 days without a commit and had drifted into a merge conflict with `main` (`mergeable_state: dirty`), so I rebased it back to mergeable.

**What was fixed:**
- Merged `origin/main` into `claude/hx-on-migration` (53df4c4). One real conflict, in `src/ludamus/client/vite.config.ts`: both this branch and `main` added a new Vite entry point to the same `rollupOptions.input` block (`panel-chrome` here, `room-lanes` from a landed PR). Kept both, alphabetically ordered — no logic conflict, purely additive.
- Everything else auto-merged cleanly (`base.html`, `panel/base.html`, `cfp-edit.html`, `facilitator-merge.html`, `notice_board/index.html`).

**Review comments:** No unaddressed CodeRabbit or Rams comments — CodeRabbit hit its per-developer rate limit on this PR and never actually ran a review (see the "Review limit reached" comment). The only substantive review feedback was from the two ad hoc "thermo-nuclear code-quality review" passes already in the thread; both verdicts were APPROVE, and their two non-blocking follow-ups (the `unsafe-eval` test assertion, the duplicated htmx type decl, the Firefox-ignore/`.env.e2e` doc nit) were already addressed in commits prior to this merge.

**Verification after the merge** (all run locally against the merged tree):
- `mise run test:py` — 2947 passed, 8 skipped
- `mise run check` (black, ruff, mypy, pylint, djlint, import-linter, codespell, oxfmt, actionlint, markdownlint, yamllint, hadolint, dockerfmt, impeccable) — all clean
- `mise run lint:tingle` — net debt did not grow vs `main`
- `mise run ts-check` — clean

Pushed as `9d1765a`. CI is now running on the merged commit; not merging per the usual policy — leaving this ready for review/merge.

---

hasparus:
## Stale-PR maintenance pass (2026-07-21)

The branch had drifted into a real merge conflict with `main` (`mergeable_state: dirty`) and had two unaddressed CodeRabbit review threads from the 07-16 pass. Rebased it back to mergeable and folded in a few more CSP fixes discovered along the way. Pushed as `05233c3`.

### Merge conflict — `src/ludamus/templates/panel/proposal-edit.html`

Since the branch's last sync, `main` (#615, "panel category session fields") extracted the facilitator search/list markup out of `proposal-edit.html` into a new partial, `panel/parts/facilitator-picker.html` — and that partial ships its own inline `<script>` for the search-filter behavior, written before this PR existed and therefore **without a CSP nonce**. That's exactly the kind of drift this pass was told to watch for.

Resolution:
- Took `main`'s refactor as-is (dropped this branch's now-redundant nonce'd inline script from `proposal-edit.html`, since the same functionality now lives in the include).
- Added `{% if csp_nonce is not None %} nonce="{{ csp_nonce }}"{% endif %}` to `facilitator-picker.html`'s script, matching this PR's established pattern. Without it, the facilitator search box on both `proposal-edit` and `proposal-create` would have silently stopped filtering the moment this PR's `script-src` enforcement landed.

### A wider gap found while checking for exactly this class of bug

Grepping the whole template tree for the same failure mode (inline `onclick`/`onchange`/`onsubmit` attributes — CSP nonces only cover `<script>`/`<style>` *elements*, never attributes) turned up ~17 templates with this pattern, most **outside** plan 019's original scope (`crowd/user/avatar.html`, `parties.html`, `chronology/print.html`, `panel/timetable.html`, `panel/proposals.html`, `panel/discounts/*`, `multiverse/panel/announcements/*`, `panel/space-form.html`, `panel/space-copy.html`, `panel/print/*`, `panel/timetable-log.html`). Those will start silently breaking (auto-submit selects, double-submit guards, print buttons) the moment `SECURE_CSP` is live in production, and none of them are covered by the current `csp-violations.spec.ts` sweep or the manual test plan in this PR's description.

I fixed the two instances that were directly in this PR's stated scope or introduced by the merge itself:
- `500_dynamic.html` (explicitly in plan 019's in-scope file list) and `404_dynamic.html` had `onclick="history.back()"` on the "Go Back" button — missed by the original nonce pass. Wired both through a nonced script instead.
- `notice_board/index.html`'s dismiss button — see CodeRabbit findings below, same root cause.

I deliberately did **not** touch the other ~15 files — that's a much larger, unrelated surface, and blind mechanical fixes across that many pages without full manual QA felt like the wrong thing to rush through in an automated pass. Flagging it here so it gets a dedicated follow-up (or at minimum, an expanded e2e/manual sweep) before/soon after this merges to production, since it directly undermines the "verified end-to-end" claim for pages outside this PR's tested surface.

### CodeRabbit findings (07-16 review, all previously unaddressed)

Addressed:
- **`plans/019-csp-enforcement-path.md`** — the Scope/STOP-conditions/Done-criteria/Step 6-8 text still read as if Step 8 (the enforcement flip) were gated and unexecuted, contradicting `plans/README.md`'s `DONE` status and this PR's own description. Added an "Executed in full" note up top plus inline annotations at each contradictory spot, rather than rewriting the whole runbook — it's a historical record now, not a live gate.
- **`base.html:11`** — `<meta name="htmx-config" content='{"allowEval":false}'>` used single-quoted attribute delimiters (djlint H008). Switched to double quotes with `&quot;`-encoded JSON.
- **`notice_board/index.html:64`** — the banner's `onclick="dismissEncountersInfo()"` is exactly the attribute-vs-element nonce gap above; the close button would go dead under enforcement. Wired it through the page's existing nonced script instead.
- **Nitpick: `csp-violations.spec.ts` selectors** — switched the sidebar-fold, category-toggle, and add-duration button interactions from id/class/`data-action` locators to `getByRole`, per the repo's E2E guideline. Left the duration-row count assertion on its id/class locator since those rows carry no distinct accessible role/text.

Skipped:
- **Nitpick: `.env.schema` alphabetical ordering, `.env.e2e` quoting** — checked for an actual linter enforcing either rule; found none (`ENABLE_CSP`'s own file already isn't strictly alphabetical, e.g. `IN_TESTS`/`CREDENTIALS_ENCRYPTION_KEY` predate this PR and aren't in position either). Reordering on a CodeRabbit assertion with no enforcing tool behind it felt like busywork, not a fix.

### Verification (all run locally against the merged + fixed tree)

- `mise run check` equivalent (black, ruff, mypy, pylint, djlint, import-linter, vulture, tingle) run directly via `poetry run` — the sandbox's `mise run` itself is currently broken here on unrelated pinned-tool-version resolution failures (`shellcheck-py==0.11.0` / `hadolint-py==2.14.0` not on PyPI) that block *any* `mise run` invocation, not just this PR's tasks. All linters clean; `tingle check` → **no new debt, 2 metrics paid off**.
- `pytest tests/integration tests/unit` → **3078 passed, 8 skipped**.
- `npx tsc --noEmit` (client) → clean.
- `csp-violations.spec.ts` on Chromium, against a live server serving the real enforcing header (`ENABLE_CSP=true`) → **3/3 passed**, zero violations.
- Ad-hoc verification against the same live server for the two fixes not covered by the existing spec sweep: notice-board dismiss button and the 404 "Go Back" button → both work, zero `securitypolicyviolation` events.
- `npm run lint` (oxlint) — pre-existing, unrelated to this PR: `eslint-plugin-better-tailwindcss@4.0.2` (a dep of `@hasparus/oxlint-config`) requires `tailwindcss@^4.1.17`, the client pins `4.1.16` exactly. Client-side lint is broken on `main` too; not something this pass should paper over.

### Status

Pushed to `origin/claude/hx-on-migration` at `05233c3`. `mergeable_state` is now `unstable` (conflict resolved; Snyk + CodeRabbit checks just kicked off on the new commit, not yet reported). Not merging, per instructions.


---

hasparus:
Follow-up: the previous push's `checks` CI job failed on a `markdownlint` (MD049 emphasis-style) violation in `plans/019-csp-enforcement-path.md` — `*elements*` used asterisk-style emphasis where this repo's config expects underscores. Fixed in `575fb59` (one-character change, verified clean with `markdownlint-cli2` locally). Unrelated to the CSP/security content of the PR.

---

hasparus:
## Stale-PR maintenance pass (2026-07-24)

Flagged by the automated sweep: no commits in 2+ days, `mergeable_state: dirty` against current `main`, `ios-regressions` check failing.

**Merge conflict — resolved and pushed.** Merged `origin/main` (~98 commits ahead) in. Four conflicts:
- `.env.schema` — trivial, both sides added additive flags (`ENABLE_CSP` here, `IS_STAGING` on main). Kept both.
- `base.html` — main had added a plain, un-nonced theme-init `<script>` with the same logic already in this branch's nonce'd version. Kept the nonce'd wrapper.
- `tests/e2e/playwright.config.ts` — kept this branch's version (wraps the e2e server via the `mise` task and carries the `HTTPS_PROXY` passthrough the CSP e2e spec needs for the Google Fonts fetch through the sandbox's egress proxy).
- `panel/base.html` — the real one: main independently reworked the whole sidebar (dark-mode theme classes, restructured header, footer fix) while its branch point predates this PR's hx-on migration, so it still had raw `hx-on:` attributes. Took main's markup/styling wholesale and reapplied the CSP-safe delegated pattern on top (`data-action`/`data-url` attributes, the nonce guard on the inline script, the `panel-chrome.ts` asset tag main was missing).

Swept every file main touched since the merge-base for newly-introduced un-nonced `<script>` tags or inline event-handler attributes (the exact bug class the 2026-07-21 pass found and fixed) — none found beyond what's listed above. The ~17-file backlog of pre-existing instances that pass explicitly deferred (`panel/timetable.html`, `crowd/user/parties.html`, etc.) is unchanged, confirmed still pre-existing at the merge-base, not new drift from this merge.

**`ios-regressions` — confirmed infra flake, not a regression.** Pulled the full job log: both attempts fail purely on iOS-simulator/xcrun plumbing (`spawnSync xcrun ETIMEDOUT`, `xcrun timed out after 608ms`, `No active session`). Zero mentions of CSP, `securitypolicyviolation`, or `panel-chrome.ts` anywhere in the log — the one interaction test that did complete a full run (`modal.test.ts`) passed 3/3 assertions. Matches PR #500's documented `xcrun`/simulator flake exactly. No code fix applicable.

**No unaddressed CodeRabbit/Rams comments** — all 3 threads from the 07-16 review are already marked resolved from prior passes; nothing new.

**Validation:** black/ruff/mypy/pylint/djlint/import-linter/vulture clean; `tingle check` — no new debt (238 metrics paid off vs `main`); `tsc --noEmit` clean, production build succeeds; `pytest` → 3108 passed, 9 skipped; `csp-violations.spec.ts` on Chromium against a live server with the real enforcing header (verified via `curl -sI`) → 3/3 passed, including the panel-chrome/sidebar/category test that directly exercises the merge-fixed markup. One follow-up commit fixed a djlint formatting nit left over from the manual conflict resolution.

Pushed as `e23534c` (merge) + `cd9232b` (djlint fix). `mergeable_state` is now `unstable` — conflict resolved, CI just kicked off on the new commit. Not merging, left ready for review.

---

hasparus:
## Stale-PR maintenance pass (2026-07-28)

Flagged by the automated sweep: no commits in 2+ days, `mergeable_state: dirty` against current `main` (~77 commits ahead), and a failing `test` CI job on `cd9232b`.

### Merge conflict — resolved and pushed

Merged `origin/main` in. Six conflicts, all additive/formatting except one real markup fork:

- `src/ludamus/client/vite.config.ts` — same pattern as every prior round: both sides added a new Vite entry to `rollupOptions.input` (`panel-chrome` here, `proposal-category-settings` on main). Kept both, alphabetical.
- `src/ludamus/templates/base.html` — main added the `og:description`/`title_suffix` meta block; kept it alongside this branch's `htmx-config` meta tag. Purely additive.
- `src/ludamus/templates/components/checkbox-field.html` + `checkbox-input.html` — main extracted a `checkbox-input.html` partial and added `checked_values` (loop-driven ticking, now used by `proposal-form.html`/`facilitator-picker.html`/`time-slot-picker.html`), predating this PR's `data_action`/`data_required_target` swap of `hx_on_change`. Took main's partial-extraction structure and reapplied this PR's `data_action`/`data_required_target` attributes in place of `hx_on_change` (which had no live callers left) — no un-nonced `hx-on:change` reintroduced.
- `src/ludamus/templates/panel/base.html` — cosmetic reflow only (main rewrapped the "Event:" strong tag and reindented a few blocks since the last merge); no new sidebar markup. Took main's formatting, kept this branch's nonce'd inline script and the `panel-chrome.ts` asset tag.
- `src/ludamus/templates/panel/cfp-edit.html` — main added a "Waiting-list seats" card and i18n'd the duration-picker's inline script (`hourSuffix`/`minuteSuffix`/`noDurationsText` via `{% translate %}` instead of hardcoded English). Took main's version wholesale (feature + i18n) and reapplied the nonce guard to its 3 `<script>` tags.
- `tests/e2e/playwright.config.ts` — both sides added Firefox `testIgnore` entries (`proposal-delete-restore.spec.ts`/`write-in-fields.spec.ts` on main, the proxy-gated `csp-violations.spec.ts` exclusion here). Kept both.

Swept the full diff between the previous merge-base and current `main` for new un-nonced `<script>` tags or inline `onclick`/`onchange`/`onsubmit` attributes in every template `main` touched (grep across all 30 touched files) — none found. The pre-existing ~17-file backlog from the 2026-07-21 pass is unchanged.

### `test` job failure — real regression from this PR, fixed

Root cause: `panel/timetable.html`'s track/date `<select>` filters use `onchange="this.form.submit()"` — part of the same ~17-file backlog the 2026-07-21 pass explicitly deferred as out of this PR's stated scope, since CSP nonces cover `<script>`/`<style>` *elements*, never inline event-handler attributes. That debt was latent under report-only CSP, but `.env.e2e`'s `ENABLE_CSP=true` applies to the *entire* e2e server, not just `csp-violations.spec.ts` — so every spec in the suite, including the pre-existing `tests/e2e/tests/timetable.spec.ts`, now runs against the real enforcing header. With `script-src` carrying no `unsafe-inline`, the day-select's `onchange` silently no-ops, so `daySelect.selectOption(firstDate!)` in `timetable.spec.ts:60` never triggers a navigation, and `page.waitForURL` hangs for the full 120s × 3 attempts — matching the CI log exactly (`Test timeout of 120000ms exceeded`, `waiting for navigation until "load"`), and cascading into "18 did not run" for the rest of that serial `describe` block.

Not a `panel-chrome.ts`/`switchEvent` regression (that delegated handler is unrelated markup) and not the `ios-regressions` class of infra flake — this one reproduces deterministically and is directly attributable to `ENABLE_CSP` going suite-wide.

Fix (`bc5a016`): dropped the `onchange`, marked the filter `<form>` `data-autosubmit`, and wired in the existing generic `src/filter-autosubmit.ts` (already used by `panel/proposals.html`) rather than inventing a one-off handler — it calls `form.requestSubmit()` on `change`, which is a plain native submission here (the form carries no `hx-get`), matching the original behavior. Verified against a live server serving the real enforcing header (manual `ROOT_DOMAIN`/`DB_NAME`-overridden instance, since the sandbox's default e2e port was occupied by a concurrent session): the day-select now produces `GET /panel/event/.../timetable/?track=&date=2026-08-19` instead of hanging, and the run progressed cleanly through the previously-failing test and beyond (log page, overview page, further serial tests) with zero new failure artifacts before I stopped the verification run to close out this pass.

Also fixed in the same commit: `tests/e2e/tests/csp-violations.spec.ts` imported `test`/`expect` directly from `@playwright/test`; a new oxlint rule (`no-restricted-imports`, added to `main` since the last merge) requires importing both from `./helpers/fixtures` so specs land in the client coverage report. This PR's own spec predated the rule — caught by `mise run check`, fixed to match convention.

### No unaddressed review comments

All 3 CodeRabbit threads from the 07-16 review remain resolved; nothing new posted since the 07-24 pass.

### Verification

- `pytest tests/integration tests/unit` → **3244 passed, 9 skipped**
- `mise run check` (black, ruff, mypy, pylint, djlint, import-linter, vulture, codespell, oxlint, oxfmt, taplo, yamllint, actionlint, hadolint, dockerfmt, dclint, markdownlint, impeccable) — all clean after the csp-violations.spec.ts import fix
- `tingle check` → no new debt, 4 metrics paid off (net -32 vs `main`)
- `tsc --noEmit` (client) → clean; `mise run build-frontend` → succeeds
- `csp-violations.spec.ts` on Chromium against a live server with the real enforcing header (`ENABLE_CSP=true`, verified via `curl -sI`) — sanity-checked the header/nonce mechanics manually; full spec re-run deferred to CI given sandbox port contention from a concurrent session
- `timetable.spec.ts` — verified the specific previously-hanging interaction now completes (see above); did not wait out the full 23-test serial file locally, CI will confirm

Pushed as `a426a18` (merge) + `bc5a016` (timetable CSP fix + oxlint import fix). `mergeable_state` is now `unstable` — conflict resolved, CI just kicked off on the new commit. Not merging, left ready for review.

---

fancysnake:
## Thermo-nuclear code quality review — CSP enforcement (plan 019)

Reviewed `origin/main...claude/hx-on-migration` (29 files, +1513/−57).

The direction is right and the parts you touched are genuinely good — `hx-on:` → delegated `data-action` is the correct move, `data-url` on the event `<option>` beats string-concatenating a URL in markup, hoisting `htmx.d.ts` out of `timetable.ts` is clean, and `csp-violations.spec.ts` is a real test rather than a header-shape assertion. But I don't think this can merge as-is: the policy flips to **enforcing** while the codebase still has 14 inline event-handler attributes and one un-nonced inline `<script>`, and there is no report collector to tell you about it in production.

---

### 1. Blocker — enforcing `script-src` without `'unsafe-inline'` kills 14 remaining inline handlers

A nonce does **not** re-enable inline event-handler attributes. Under CSP L2/L3, `onclick=` / `onchange=` / `onsubmit=` are gated by `script-src` and are only permitted by `'unsafe-inline'` (or per-handler `'unsafe-hashes'`). `'nonce-…'` applies to `<script>` elements only. The new `CSP_POLICY` has neither, so every one of these becomes a silently dead handler the moment the header goes enforcing:

| Template | Attribute | What breaks for the user |
|---|---|---|
| `panel/proposals.html` | `onchange="this.form.submit()"` | Track filter does nothing |
| `panel/timetable-log.html` | `onchange="this.form.submit()"` | Log filter does nothing |
| `crowd/user/avatar.html` (×2) | `onchange="this.form.submit()"` | Avatar upload never submits |
| `panel/parts/_print-scope-control.html` | `onchange="window.open(…)"` | Print-scope picker does nothing |
| `panel/print/base.html` | `onclick="window.print()"` | **Print button dead** |
| `chronology/print.html` | `tessera_button … onclick="window.print()"` | **Print button dead** |
| `crowd/user/parties.html` (×2) | `onclick="…style.display=…"` | Companion inline edit/cancel dead |
| `panel/space-form.html`, `panel/space-copy.html`, `panel/discounts/edit.html`, `panel/discounts/export.html`, `multiverse/panel/announcements/{create,edit,delete}.html` | `onsubmit="…disabled = true"` | Double-submit guard gone — **duplicate spaces, discounts, announcements** |

Plus `chronology/_compact_schedule.html` carries an inline `<script>` with **no nonce** at all, so the scroll-restore is blocked outright. `csp-violations.spec.ts` visits `/event/autumn-open/` and passes only because that include sits behind `{% if compact_schedule %}` and the seed data takes the card branch — the spec is asserting less coverage than its comment claims.

The double-submit ones worry me most: they fail *open* and produce duplicate rows, on the exact pages where a duplicate costs a human cleanup.

This isn't a matter of adding a few more `data-action` cases. The conversion has to be complete before the header can be enforcing, and the spec needs to be able to prove it. Two things would make that structural rather than a checklist:

- **Reuse the canonical helper you already have.** `filter-autosubmit.ts` (`form[data-autosubmit]`) already does exactly what all four `onchange="this.form.submit()"` sites want, and you already applied it to `panel/timetable.html` in this PR. Apply it to `proposals.html`, `timetable-log.html`, and `avatar.html` too instead of leaving three near-duplicates of a pattern you just deleted next door.
- **Collapse the six `onsubmit` copies into one listener.** Six templates carry a byte-identical `this.querySelector('button[type=submit]').disabled = true`. That's a `data-submit-once` attribute and a five-line delegated `submit` listener, and it deletes copy-paste that predates this PR.

For the print buttons and `parties.html`, the same `data-action` dispatch you just built covers them — they just need to be loaded outside `panel/base.html` (see #4).

### 2. Blocker — enforcing with no `report-uri` means breakage is invisible

The settings comment is honest that no ingestion endpoint exists yet, but that combines badly with #1: you're flipping from a policy that can't break anything to one that breaks silently, with no telemetry, on pages the e2e suite doesn't cover. Given #1 exists, the risk isn't hypothetical.

I'd want one of these before merge: (a) wire a report endpoint first, or (b) finish the handler conversion *and* extend `csp-violations.spec.ts` to cover the affected pages, or (c) land `SECURE_CSP_REPORT_ONLY = CSP_POLICY` with the new nonce-based policy, confirm clean in staging, then flip enforcing in a follow-up. (c) is the boring path and it costs one extra PR.

### 3. Code judo — the 15-copy nonce incantation should be one interpolation

`{% if csp_nonce is not None %} nonce="{{ csp_nonce }}"{% endif %}` is repeated 15 times across 12 templates, and it needs a **12-line comment in `base.html` plus a 10-line comment in `500_dynamic.html`** to explain why it's `is not None` and not the `{% if csp_nonce %}` Django documents. When a construct needs twenty-two lines of prose to justify its shape and is then copy-pasted fifteen times, that's the abstraction telling you where it wants to live.

It can just be:

```html
<script nonce="{{ csp_nonce }}">
```

`ContentSecurityPolicyMiddleware.process_request` sets `request._csp_nonce = LazyNonce()` unconditionally — it does not consult `SECURE_CSP` — and the middleware is unconditionally in `MIDDLEWARE`. So on any real request `csp_nonce` is a `LazyNonce`, and plain interpolation reads it, which is the same thing that forces generation today. In a request-less render the variable is simply absent from context and resolves to `string_if_invalid` (`""`), giving a harmless `nonce=""`. Your own tests keep passing: `re.findall(r'nonce="([^"]+)"')` won't match an empty value.

That deletes 15 conditionals and both explanatory comments, and incidentally fixes the `<script  nonce=` double-space / `<script >` artifact and the inconsistency between `<script {% if … %}` and `<script type="module"{% if … %}`.

If you specifically want the attribute *absent* when CSP is off, make that a `{% nonce %}` simple tag in `tessera` — one Python function holding the invariant, one token in every template. Either way the incantation stops being copy-pasted.

### 4. `panel-chrome.ts` is accumulating non-panel behavior, and a shared component now depends on a panel-only script

Two of the five handlers in this file aren't panel chrome:

- `sync-expanded-required` is generic checkbox behavior, and it's wired through `components/checkbox-input.html` — a **shared** component whose docstring now advertises `data_action` / `data_required_target` support. Any non-panel template that uses `checkbox-field.html` with `data_action` gets an attribute that silently does nothing, because `panel-chrome.ts` is only loaded from `panel/base.html`. Today the only consumer (`multiverse/panel/connections/edit.html`) does extend `panel/base.html`, so it works — but the component contract is now wider than the runtime that backs it. That's the kind of thing that fails six months from now with no error message.
- The `htmx:afterRequest` refresh handler is a generic "swap a pane after a successful request" primitive, not sidebar chrome.

Given #1 needs a document-level dispatcher on non-panel pages anyway (print, parties, the `onsubmit` guards), the natural shape is a small `data-action` dispatcher loaded from **`base.html`**, with panel-specific actions staying in `panel-chrome.ts`. That kills two birds: shared components get a runtime that's always present, and the print/parties handlers have somewhere to go.

While you're in there: the two listeners are the same `closest('[data-action]')` → `switch` shape twice, with a `default: break;` that does nothing. A `Record<string, (el: HTMLElement) => void>` per event type and one registration loop removes the duplication and the dead default.

### 5. Code judo — self-host the font and two policy exceptions plus a test hack disappear

`style-src` gained `https://fonts.googleapis.com` and `font-src` gained `https://fonts.gstatic.com` purely because `index.css` `@import`s Outfit at runtime. That single `@import` is also what produced the 11-line conditional Firefox exclusion in `playwright.config.ts` (`...(proxyServer ? [/csp-violations\.spec\.ts/] : [])`) with its own 10-line explanation of proxy behavior in sandboxed containers.

Bundling the font locally through Vite deletes: two external-origin allowances, ~10 lines of settings comment, the proxy-conditional test exclusion and its comment, and a render-blocking third-party request on every page. Also worth noting the exclusion means CSP coverage on Firefox is now environment-dependent — the config silently drops a spec depending on whether `HTTPS_PROXY` is set, which is exactly the kind of "temporary" branching that becomes permanent.

I'd take this one even though it's outside the strict scope of the plan; it makes the resulting policy meaningfully tighter and removes more code than it adds.

### 6. Smaller things

- `ENABLE_CSP = IS_PRODUCTION or env.bool("ENABLE_CSP", default=False)` is a module-level name used exactly once, three lines later. `if IS_PRODUCTION or env.bool("ENABLE_CSP", default=False): SECURE_CSP = CSP_POLICY` says the same thing without adding a settings-module symbol that reads like a public knob.
- `test_style_src_keeps_unsafe_inline`, `test_script_src_has_no_unsafe_inline` and `test_header_sent_when_production_policy_active` each re-set `settings.SECURE_CSP` and re-issue the same request. A fixture returning the parsed header would drop the repetition. Low priority — the `_directive` helper itself is good.
- `500_dynamic.html` now has the theme IIFE *and* a `DOMContentLoaded` back-button binding in one `<script>`, plus a 10-line comment. Once #3 lands the comment goes; the back-button binding is the same `history.back()` handler as `404_dynamic.html`, which is a third copy of "delegate this" — another `data-action` case.

---

**Verdict: request changes.** #1 is a functional regression across print, avatar upload, two filter surfaces, companion editing, and six double-submit guards; #2 means you won't find out about it. #3 and #5 are the two places where this change can get materially smaller rather than just tidier, and #4 decides where the dispatcher should have lived in the first place — worth resolving before more `data-action` cases land on top of the current placement.

<sub>Note on verification: the inline-handler claim in #1 is from the CSP spec's treatment of event-handler attributes, not from an observed failure — I did not run the e2e suite. The cheapest way to confirm it is to point `csp-violations.spec.ts` at `/panel/event/<slug>/proposals/` and click the track filter.</sub>

PR #591, opened 2026-07-15
Cache event page in the browser for 180s

hasparus:
## Problem

Navigating between `/event/<slug>/` and `?view=rooms` felt slow. Both URLs hit the same `EventPageView`, which recomputes the entire personalized schedule (session queries, shadowban/ban masking, bookmarks, DTO building) on every request — and the page had **no caching at all**. Toggling layouts paid the full server cost twice, as full page loads.

## Fix

Send `Cache-Control: private, max-age=180` on 200 GETs of the event page. Repeat visits and list↔rooms toggling within 3 min serve from the browser disk cache — zero server round trip.

- `private`: never shared/CDN — the page is masked per-user (enrollment counts, bookmarks, shadowban).
- `max-age=180`: bounds seat-count staleness. Ceiling: after someone enrolls, counts can lag up to 3 min on a cached page.
- Guarded to `GET` + `200` so error/redirect responses stay uncached.

The `list↔rooms` layouts derive from the same already-fetched data (`build_room_lanes` is pure in-memory regrouping), so no server work distinguishes them — caching the whole response is the right lever.

## Skipped

Client-side toggle (render both layouts, switch with CSS/JS — removes the round trip entirely). Add when the *first* load is the bottleneck, not the toggle.

## Test

`test_private_browser_cache_header` asserts the header. `mise run check` green.

## Summary by CodeRabbit

- **Performance**
  - Improved browser and intermediary caching for event pages, design pages, event prints, QR codes, and calendar downloads.
  - Published content can be cached publicly, while private or unpublished content remains protected with shorter cache periods.

- **Tests**
  - Added coverage to verify cache-control behavior across affected pages and downloads.

### DIFF
--- a/src/ludamus/adapters/web/django/print_views.py
+++ b/src/ludamus/adapters/web/django/print_views.py
@@ -8,11 +8,10 @@ from typing import TYPE_CHECKING, Literal
 from django.http import Http404, HttpResponse
 from django.template.response import TemplateResponse
 from django.urls import reverse
+from django.utils.cache import patch_cache_control
 from django.utils.dateparse import parse_datetime
-from django.utils.decorators import method_decorator
 from django.utils.timezone import get_current_timezone, localtime, make_aware
 from django.utils.translation import gettext_lazy as _
-from django.views.decorators.cache import cache_control
 from django.views.generic.base import View
 
 from ludamus.mills.qr import qr_svg
@@ -136,7 +135,6 @@ def _timetable_scope_name(
     return None
 
 
-@method_decorator(cache_control(public=True, max_age=300), name="dispatch")
 class PublicEventPrintView(View):
     request: RootRequest
     template_name = "chronology/print.html"
@@ -216,7 +214,7 @@ class PublicEventPrintView(View):
         )
         sphere = request.services.sphere_panel.read(request.context.current_sphere_id)
 
-        return TemplateResponse(
+        response = TemplateResponse(
             request,
             self.template_name,
             {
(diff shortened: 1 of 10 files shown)


hasparus:
### Ready to merge — needs a maintainer to waive `tingle`

Thermo-nuclear code-quality review: **APPROVED** (blocker on the print-page `public` leak and the substring-assertion nit both fixed). All CI green **except `tingle`**, which fails on `legacy-loc +11`.

That +11 is the irreducible cost of adding browser `Cache-Control` to four `adapters/**` views (imports + decorators + the print-view published/unpublished branch). It's a justified addition, not counter-gaming — `type-object`/`wrong-assert` stay at 0, and the two `gates/**` notice-board views add zero legacy LOC. Decision on this PR: accept the debt rather than offset it by deleting unrelated legacy lines.

**A maintainer needs to override/waive the `tingle` check to merge.** Everything else (tests, impeccable, CodeQL, snyk) is green.

hasparus:
### CodeRabbit review addressed (995deb6)

1. **Scope cache headers to GET** ✅ — decorators moved from `dispatch` to `get`, so `405 Method Not Allowed` responses are no longer stamped with cache headers. (Print view already patches inside `get()`.)
2. **`private` → `no-store` on event/events pages** — **declined.** `no-store` disables *all* browser caching, which defeats the purpose of this PR (the original ask was "navigation is slow, don't we cache?"). `private` already keeps the per-user data out of shared/CDN caches while letting the user's *own* browser cache their *own* data; the 180s TTL bounds staleness. The security concern `no-store` addresses (shared-cache exposure) is already covered by `private`.
3. **Centralize header assertions** ✅ — `assert_response` gained a `cache_control` set param (exact directive match, so a stray `public` regresses the test), plus a reusable `assert_cache_control` helper. Removed the duplicated manual parsing across six test sites.

PR #593, opened 2026-07-15
fix(chronology): stop mobile long-press on hour rail opening a new tab

hasparus:
## Problem

On mobile, the compact event schedule's hour rail (the vertical hour scrubber
on the right) is built from `<a href="#slot-…">` anchors. **Long-pressing a
marker raised the browser's native link callout** — offering to open the
current page at that fragment in a new tab — instead of scrubbing. The rail's
existing `touch-none select-none` governs scroll gestures and text selection;
neither suppresses the link callout.

## Fix

Two one-liners, at the engine each applies to:

- **iOS Safari** — `-webkit-touch-callout: none` on `.schedule-rail-hour`
  (`index.css`). The platform-native off-switch for the long-press link menu.
- **Android Chrome** — `preventDefault` the rail's `contextmenu` event
  (`event-timeline.ts`), which is what Android raises on a long-press. Also
  neutralises desktop right-click on a marker — meaningless on a scrubber.

Dropping the `href` was rejected: it would break the scroll-spy hour mapping,
the jump-link semantics, and keyboard/AT accessibility.

## Regression coverage

- **Cross-engine (CI, Playwright)** — asserts a `contextmenu` on a marker is
  `defaultPrevented`. Runs on chromium + firefox. Guards the Android/JS path.
  (Desktop Chromium drops `-webkit-touch-callout` from both computed style and
  parsed rules, so the iOS callout can't be observed there.)
- **Real iPhone simulator (CI, agent-device)** —
  `scripts/reproduce-ios-scrubber-longpress.ts`, wired into the **Mobile**
  workflow next to the existing modal repro. Boots an iPhone 16 sim, opens the
  dense seeded schedule in Safari, long-presses an hour marker, and **fails if
  the iOS callout menu ("Open in New Tab", …) appears**. This is the mechanism
  that actually exercises the reported bug on the real engine.

## Verification

- `npx playwright test event-schedule-rail` → 7 passed, 1 skipped (existing
  mobile-cursor skip).

Screenshots omitted: the change is a long-press behaviour, not a visual one —
there's nothing new to see on the page.

### DIFF
--- a/.github/workflows/mobile.yml
+++ b/.github/workflows/mobile.yml
@@ -8,8 +8,7 @@ on:
       - "src/ludamus/client/**"
       - "src/ludamus/templates/**"
       - "tests/e2e/**"
-      - "scripts/reproduce-ios-modal-bugs.*"
-      - "scripts/reproduce-ios-scrubber-longpress.*"
+      - "scripts/ios-regressions/**"
       - ".github/workflows/mobile.yml"
 
 permissions:
@@ -71,8 +70,8 @@ jobs:
 
       - name: Start local e2e server
         run: |
-          mise run test:e2e:serve > ios-modal-server.log 2>&1 &
-          echo $! > ios-modal-server.pid
+          mise run test:e2e:serve > ios-server.log 2>&1 &
+          echo $! > ios-server.pid
 
       - name: Wait for local e2e server
         run: |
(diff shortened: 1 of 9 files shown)


hasparus:
## Maintainability review — ship it, with a couple of low-severity notes

**Verdict: approve.** The behavioral fix is small, direct, and lands exactly where the bug is. `-webkit-touch-callout: none` sits on `.schedule-rail-hour` (the `<a>` markers that are actually long-pressable), and the `contextmenu` `preventDefault` sits on the rail next to the other pointer listeners. No structural regression, no file pushed toward the 1k line: `index.css` grows by 4 lines and the new repro is a standalone file, not bloat added to an existing one. Both fix surfaces are exercised (Playwright for the synthetic `contextmenu`, the agent-device repro for the real iOS callout). Nothing here is a blocker.

A few things worth a comment before merge, none rising above "nit":

### 1. The callout suppression could be a Tailwind utility, co-located with its siblings (low)
`src/ludamus/client/src/index.css` — the rail's other touch tuning (`touch-none select-none`) already lives as Tailwind utilities on the element in `_compact_schedule.html`. This new rule instead adds a bespoke block to `index.css`, which is already ~766 lines and drifting toward the 1k threshold. Per the repo's "prefer Tailwind over CSS files" rule, this reads more naturally as `[-webkit-touch-callout:none]` on the `<a class="schedule-rail-hour ...">` marker, keeping all the touch behavior in one place and off the growing stylesheet.

### 2. The fix is split across two scopes — worth a one-line "why" (nit)
The CSS targets the markers (`.schedule-rail-hour`) while the `contextmenu` handler is bound to the whole rail (`.schedule-rail`). They're genuinely complementary — `-webkit-touch-callout` is what actually suppresses the iOS long-press callout (which never fires a `contextmenu` event on iOS Safari), while the `preventDefault` catches desktop right-click / synthetic `contextmenu`. That's correct defense-in-depth, but the scope asymmetry invites a future reader to "simplify" by deleting one half as redundant. A short comment on the listener noting it covers the desktop/synthetic path (and that iOS relies on the CSS) would protect it.

### 3. Duplicated iOS agent-device harness — flagging by design, not asking for action now (low)
`scripts/reproduce-ios-scrubber-longpress.ts` shares ~150 lines of near-identical harness with `reproduce-ios-modal-bugs.ts`: the `importAgentDevice` shim, env parsing, `ensureSimulator`, `takeSnapshot`/`snapshotLabels`, `openUrl`/`openUrlWithSafari`, both session-takeover helpers, and `assertEventPageReady` (which differs only in the body substring it polls for). This mirrors the established convention deliberately, so it's fine to land as-is — but this is now the *second* copy. If a third iOS repro shows up, extract the shared lifecycle (import shim, simulator/session takeover, snapshot + URL helpers, page-ready poll) into a `scripts/lib/` module and let each repro keep only its device-specific assertion. Noting it so the debt is visible, not blocking on it.

### 4. CI step is a verbatim copy of the modal step (nit)
`.github/workflows/mobile.yml` — the new step duplicates the modal step's retry/cleanup loop exactly, differing only in the script name and `SESSION`. `$UDID` is correctly inherited via `$GITHUB_ENV`, so it works. Acceptable given GitHub Actions' weak DRY story; a script-name matrix would collapse the two, but not worth complicating the shared server lifecycle over two entries.

Nice, focused change — clean fix, real regression coverage on both the web and native surfaces.

hasparus:
Follow-up: the iOS regressions no longer live as per-script bash steps in `mobile.yml`. Both are now `bun:test` files under `scripts/ios-regressions/` (`modal.test.ts`, `scrubber-longpress.test.ts`), discovered and run by a single `bun test scripts/ios-regressions/` step. Adding a regression is now just dropping a `*.test.ts` in that dir. Per-file session names stay distinct (`${SESSION}-modal` / `-scrubber`) so the simulator handoff between suites is unchanged.

PR #608, opened 2026-07-18
fix(flaky): Anonymous code modal › rejects an unknown code with a flash message and stays on the event

hasparus:
## Root-cause analysis

`tests/e2e/tests/event-details.spec.ts:340` was the flakiest test in CI this week — it failed-then-passed-on-retry 3 times (2026-07-15, 2026-07-17 ×2), always with the identical error:

```
Timeout 200ms exceeded while waiting on the predicate
```

on this assertion:

```ts
await expect
  .poll(
    () => flash.evaluate((element) => Number.parseFloat(getComputedStyle(element).opacity)),
    { timeout: 200, intervals: [50] },
  )
  .toBeLessThan(1);
```

The flash message's dismiss animation is defined in `index.css`:

```css
.flash-region .alert {
  transition:
    transform 0.26s var(--ease-out),
    opacity 0.26s var(--ease-out),
    ...
}
```

The CSS opacity transition takes **260ms**, but the test only polled for **200ms** to see the opacity value start dropping below 1. Under normal load the transition starts fast enough to clear a 200ms window, but any CI paint/GC jitter — a slow frame, a delayed style recalc — pushes the first observable opacity change past 200ms, and the poll times out even though the animation is running correctly. Playwright's `retries: 2` (`playwright.config.ts`) then quietly retries and passes, masking it as "flaky" rather than a hard failure.

## Fix

Widened the poll timeout from 200ms to 1000ms (kept the 50ms interval), so it comfortably clears the 260ms transition with margin for CI jitter, instead of racing it.

## Verification

- Confirmed via `git stash` that the failure mode is unrelated to a separate, pre-existing sandbox slowness issue in this dev container (an earlier assertion in the same test — `dialog.toBeVisible()` — times out locally even on unmodified `main`, due to unrelated headless-chromium slowness in this specific sandbox; not reproducible in real CI, which passes this same step in every logged run).
- With that unrelated local-only bottleneck worked around (temporarily raising the global `expect.timeout` for a diagnostic run, then reverting — not part of this commit), the test with the fix applied passed 3/3 repeat runs.
- The fix is a 1-line, well-understood timing correction directly matching the CI-observed error message and the actual CSS transition duration, so it should resolve the flake without changing test intent.

## Context

This was found via a 7-day flaky-test sweep across ~153 CI workflow runs (push + pull_request, `ci.yml`). Full ranking:

| Rank | Test | File | Occurrences (flaky+hard-fail) this week |
|---|---|---|---|
| 1 | Anonymous code modal › rejects an unknown code with a flash message and stays on the event [chromium] | `tests/e2e/tests/event-details.spec.ts:340` | 6 (3 flaky, 3 hard-fail) |
| 2 | interface sound toggle › remembers the choice across a page reload [firefox] | `tests/e2e/tests/sound.spec.ts:25` | 4 (all flaky) |
| 3 | Backoffice Panel › lists the space tree for the event [chromium] | `tests/e2e/tests/panel.spec.ts:176` | 2 (both flaky, `getByText` strict-mode violation on duplicate "Hearth Lounge" text) |
| 4 | Backoffice Panel › creates a top-level space [chromium] | `tests/e2e/tests/panel.spec.ts:184` | 1 (flaky, same duplicate-text pattern, "Community Library") |

This PR only fixes #1 as instructed. #2 (firefox `page.reload()` occasionally exceeding the 120s test timeout) and #3/#4 (locator ambiguity on duplicate visible text in the space tree) are separate root causes worth their own follow-up.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JxirCRVEJVX5EUp5MWHh9B)_

## Summary by CodeRabbit

* **Tests**
  * Improved end-to-end test reliability when dismissing invalid-code messages.
  * Added timing tolerance for the message fade-out animation to reduce intermittent test failures.


hasparus:
## Stale-branch triage

**Conflict resolved:** merged `origin/main` into this branch (commit `b61225d`). The only real conflict was in `tests/e2e/tests/event-details.spec.ts`, both sides in the exact same hunk: while this branch sat stale, `main` independently picked up its own fix for the identical flake (same `expect.poll` on the flash's opacity) — but with a more generous `timeout: 2000` and a different explanatory comment, vs. this PR's `timeout: 1000`. Kept `main`'s version since it: (a) has already been live and green on `main`, and (b) gives more margin against the same CI jitter this PR set out to fix. Net effect: after the merge, `tests/e2e/tests/event-details.spec.ts` is byte-identical to `main`'s tip — this PR's own contribution has effectively already landed via `main`, and the merge just catches the branch up cleanly. The unrelated hunk in the same file (reordering of the "Open details" click vs. DOM-injection lines ~232–256) was only touched by `main`, so it merged automatically with no conflict.

**CodeRabbit / bot comments:** CodeRabbit's review is **Approved** with **no actionable comments**. Its automated pre-merge checks did flag two warnings ("Linked Issues check", "Out of Scope Changes check") claiming this PR should instead contain Renovate-onboarding changes for issue #1 — that's a misconfigured/misdirected check unrelated to this PR's actual, correctly-scoped e2e timeout fix; not acted on. No comments from a "Rams" reviewer were found — the "Rams Design Review" check run completed with conclusion `neutral` and no comment body to address (it's a design-system check on an e2e-test-only diff, so nothing applicable).

**Local verification:** `mise run check`/`ts-check` couldn't run in this sandbox — every `mise run` task fails at tool-install (`pipx:hadolint-py==2.14.0` / `pipx:shellcheck-py==0.11.0` aren't published on PyPI at those pinned versions), unrelated to this change; logged as a papercut. Worked around by running `npx tsc --noEmit -p .` directly inside `tests/e2e`: zero type errors in `event-details.spec.ts` (one pre-existing, unrelated `TS2339` error in `promotion.auth.spec.ts` reproduces identically on unmodified `main`). Also hand-diffed the merged file against `main`'s tip — no delta.

**Push:** fast-forward push of the merge commit to `origin/claude/confident-newton-voc91l` succeeded (no force-push).

**Current state:** `mergeable_state` is now `unstable` (mergeable, no more conflicts) — CI just kicked off on the new merge commit (`b61225d`) and is `pending` at the time of this comment (Snyk + CodeRabbit running). Not merging this PR per instructions — leaving that to a human/maintainer once CI is green.


---

PR #613, opened 2026-07-18
Filter event schedule by track and proposal category

fancysnake:
## What

Adds **Track** ("Blok programowy") and **Category** ("Kategoria") dropdowns to the event page (`/event/<slug>/`) filter panel, letting attendees narrow the schedule by programme track and proposal category.

## How

Reuses the event page's existing **client-side filter channel** — each card already exposes `data-tag-categories="slug:value;…"` and `session-filters.ts` drives any `.tag-filter` select from it. Track/category ride the same mechanism, so:

- **No new JavaScript.** Both the card grid and the compact schedule stay in sync via the single shared `_session_data_attrs.html` partial.
- Track is multi-value (M2M), category single — both fit the existing key→value channel naturally.
- Dropdowns reveal only when **more than one** value exists, matching how Venue/Day/Hour behave.
- **Public tracks only** (`is_public`); private/manager tracks never leak to attendees.

### Changes
- `sessions.py` (links): `with_session_card_relations` eager-loads `category` + `tracks`.
- `event_presentation.py` (gates): `SessionData.track_names` / `category_name`; `filter_categories` property emits `__track:`/`__category:` pairs; `filter_availability()` computes the reveal flags.
- `views.py`: populates the fields, calls `filter_availability`.
- `event.html`: two tessera-styled selects mirroring the existing Venue/tag-filter markup.

## Notes for review
- **Debt (tingle):** the feature adds ~4 lines to the legacy view. Offset by fixing a real bug in the same method — the viewer's email was read from the DB **twice** in `get_context_data`, now read once. `tingle check` nets to zero.
- **Known limitation:** track/category names containing `:` or `;` would break pair parsing — the same pre-existing limitation as the current tag-value channel; real names don't contain them.

## Tests
- Unit: `filter_categories` formatting (tracks, category, combined, public-field prefix).
- Integration: selects + `__track`/`__category` data pairs render; **private tracks stay hidden**.
- Full suite (2966) + lint + `tingle check` + ts-check + lint-client green.

## Summary by CodeRabbit

* **New Features**
  * Added track and category dropdown filters to the event page.
  * Filters are shown only when multiple relevant tracks/categories exist.
  * Session cards now expose track and category data to power filtering.
  * Private tracks remain hidden from public filter options.
* **Localization**
  * Updated Polish translations for “All tracks” and “All categories,” including correct labeling scope.
* **Tests**
  * Updated and expanded integration and unit tests to cover filter visibility and filter formatting.
* **Maintenance**
  * Updated translation conventions, tooling lint configuration, and added a papercut note for intermittent test flakiness.


hasparus:
@claude do we skip these filters for events with just one possible option? ( I feel we should )

PR #627, opened 2026-07-21
refactor(panel): unify proposal create/edit into one view + template

fancysnake:
`ProposalCreatePageView` and `ProposalEditPageView` (and their two field component views) collapse into `ProposalFormPageView` / `ProposalFormFieldsComponentView`, keyed by an optional `proposal_id`. One template, `panel/proposal-form.html`, branches on `proposal` for the create-vs-edit differences (title/actions, cover-image and personal-data cards are edit-only); facilitators, tracks, and time slots are shared.

The `proposal-create` / `proposal-edit` URL names are unchanged.

## Functional gains

- The create flow now assigns tracks (the card was edit-only before).
- Facilitators use the same inline search picker on both flows; the "at least one facilitator" rule on create moved from a required form field to a view-level check, so create/edit share form-building.
- Create now uses the same `*_submitted` marker convention as edit for facilitators / tracks / time slots.

## Bug fixed along the way

Creating a proposal on an event with **no facilitators at all** used to re-render the page with no feedback whatsoever. The rule reports through `form.add_error(None, …)`, and non-field errors were rendered only inside the facilitators card — a card that doesn't exist when the event has no facilitators. The message went on the floor and the user just saw the form again.

Form-level errors now render above the form, outside the column flow, so they show whichever cards the event's data happens to produce. Binding the form's *generic* error channel to one specific card was fragile regardless — any future non-field error would have inherited the same disappearing act.

## Review follow-ups included

- `checkbox-field` gained `checked_values`, collapsing three `{% if checked %}` / `{% else %}` include pairs to one call each, and no longer opens `<label>` in one `{% if %}` and closes it in another — the shared input moved to `checkbox-input.html` so each branch is a complete element.
- The last hand-rolled `<input type="checkbox">` (`remove_field_ids`) goes through the component.
- `has_errors` on `{% select %}` now carries `aria-invalid` in tag mode too, not just the red border — the attribute was stranded in the form-mode branch. Restored the `SelectNode` contract comment naming both silently-popped keywords.
- `ProposalFormFieldsComponentView` no longer subclasses the page view. Both derive from `_ProposalFormBase`, which holds only the request-scoped helpers they share, so the fragment endpoint doesn't inherit (and then have to fence off with `http_method_names`) the create/edit POST handler.
- "Create one first" opens in a new tab — it's the only way out of an empty facilitator list, and the proposal being typed has no draft storage.
- The facilitator `Invariant:` comment was wrong and is reworded: it's a create-time rule. `SessionContentEditService.apply` documents that an empty `facilitator_ids` list replaces the assignment, so an edit can still remove every facilitator.

## Debt

`tingle`: no new debt, **94 paid off**. `mock-ANY` net −4 (`context_data=ANY` replaced with full expected context throughout both proposal test modules) and `raw-form-controls` net −6. The one metric that grows is `type-any`, which is the `dict[str, Any]` context plumbing the unified view inherits from the two views it replaces.

## Deferred

Filed rather than bolted on here:

- #681 — changing the category leaves `duration` and `participants_limit` stale, because they're category-derived but sit outside the HTMX swap target (pre-existing on both old templates)
- #682 — tessera: link field errors and help text to their inputs with `aria-describedby`
- #683 — make the three checkbox pickers data instead of three code paths
- #684 — merge the two proposal test modules now that they test one view

## Screenshots

_Panel → Proposals, captured against the seeded e2e data (`frostfire-con`, `lakeside-weekend`)._

**Create** — one template, facilitator picker in its own card:

**Create, submitted with no facilitator picked** — the form-level error now renders above the form:

**Create on an event with no facilitators at all** — this is the case that used to fail silently. The picker card doesn't exist, so the error had nowhere to render; now it shows above the form alongside the "Create one first" escape hatch:

**Edit** — same template, assigned facilitator ticked through the component's new `checked_values`:

## Summary by CodeRabbit

* **New Features**
  * Added a unified proposal form for creating and editing sessions.
  * Added dynamic category fields and improved facilitator, track, and time-slot selection.
  * Added support for preserving selections and entered personal data when validation fails.
* **Bug Fixes**
  * Improved facilitator validation and selection handling.
  * Corrected proposal form error accessibility indicators.
  * Fixed event date display formatting in the panel.
* **Style**
  * Standardized form controls and picker components across proposal workflows.
  * Improved labels, loading states, and translated interface text.


fancysnake:
> [!NOTE]
> **Status as of the current head.** ✅ still true / done · ❌ superseded.
>
> - ✅ **Rams a11y + spacing** — still fixed; both threads resolved. The link additionally opens in a new tab now, so following it no longer discards the proposal in progress.
> - ✅ **User-Centric E2e Tests** — stands. Confirmed false positive.
> - ❌ **Design System** — superseded, and better than claimed. CodeRabbit's own sticky comment now reports this check as **passed**. The remaining raw control (`remove_field_ids` in `proposal-session-fields.html`) has since been converted, so `raw-form-controls` is net **−6** on the branch rather than a deferred follow-up.
> - ✅ **Red CI jobs** — the infra/flake diagnosis stands; nothing in the diff was implicated.

Addressed the review feedback:

- **Rams a11y (link color-only)** and **Rams spacing (heading margin)** — fixed in 9b881a67. The "Create one first" link now carries a persistent underline, and the Session Details heading uses `mb-2` to match the other cards (aligned the outlier down rather than everything up).

On the two failed pre-merge checks:

- **User-Centric E2e Tests** — false positive. `test_proposal_create_page.py` / `test_proposal_edit_page.py` are Django **integration** tests using the repo's mandated `assert_response` / `checkbox_tag` helpers. Asserting on rendered markup (`name="facilitator_ids"`, `id="facilitator-search"`) is the house pattern for view/template tests here — these are not Playwright E2E with role locators, so `getByRole`/`getByText` don't apply.

- **Design System** — real debt, but **pre-existing and out of scope** for this PR. The hand-rolled `<select name="category_id">` and facilitator/track checkboxes already lived on `main` in `proposal-create.html` and `proposal-edit.html`; this refactor only unified the two templates and introduced no new raw controls. Migrating them to the tessera `select`/`checkbox` tags is tracked as a follow-up.

The two red CI jobs (`checks`, `ios-regressions`) were infra/flaky and unrelated to the diff: `checks` failed on a pylint **segfault** (exit 139; mypy/black/djlint/tingle all green), and `ios-regressions` timed out on `scrubber-longpress.test.ts` on the event participant page — untouched by this panel refactor.

fancysnake:
> [!NOTE]
> **Resolution pass — every item below is marked.** ✅ fixed in this PR · ⏭️ deferred to an issue · ❌ not accepted, with the reason.
>
> | # | Item | Status |
> |---|---|---|
> | 1 | Picker abstraction is stringly-typed | ⏭️ #683 (duplicate-query claim ❌) |
> | 2 | Three copies of the checkbox workaround | ✅ |
> | 3 | `checkbox-field.html` split `<label>` | ✅ |
> | 4 | Facilitator rule comment + vanishing error | ✅ |
> | 5 | Unrelated tooling churn | ✅ (`# BLACK` claim ❌) |
> | — | `context_data=ANY` | ✅ |
> | — | Two test modules for one view | ⏭️ #684 (helper claim partly ❌) |
> | — | `ProposalFormFieldsComponentView` subclassing | ✅ |
> | — | `SelectNode` lost its contract comment | ✅ |
> | — | HTMX swaps too little | ⏭️ #681 |
> | — | `remove_field_ids` raw checkbox | ✅ |
>
> Full suite green (3142 passed, 9 skipped), all linters green, `tingle`: no new debt, 94 paid off.

---

## Thermo-nuclear code-quality review

The direction is right and the numbers back it: `proposals.py` goes 1320 → 1242, two templates become one, `tingle` reports net **-93**. Create gaining tracks and the shared facilitator picker are real wins.

But the refactor stops halfway. It unified the *view* and left the *template* with three copies of the thing it just unified, and the unification itself is built on string concatenation rather than a type. There's also a chunk of unrelated tooling churn that weakens the local check gate.

---

### 1. The picker abstraction is stringly-typed, and the template never got it

> ⏭️ **Deferred → #683.** Accepted as a real design gap; it's a genuine ~95-line change across view and template plus a decision about the per-card headings, so it isn't riding along in this PR.
>
> ❌ **The duplicate-query claim is wrong.** `_render` only runs on the *failure* path and `_collect_track_ids` / `_collect_time_slot_ids` only on the *write* path — they never co-occur, so tracks and time slots are never fetched twice. The one real double-fetch is `facilitators` on the invalid-create re-render: one extra query, on an error path.

`_picker_context(plural=..., singular=...)` builds four different names by f-string:

- `f"{plural}_submitted"` and `f"{singular}_ids"` → POST field names
- `context[f"all_{plural}"]` and `context[f"assigned_{singular}_pks"]` → template variable names

Nothing checks that `("time_slots", "time_slot")` lines up with what `proposal-form.html` reads. Rename a context key in the template and the view keeps happily writing the old one. `_HasPk` exists solely so this helper can reach a `.pk`. That's generic machinery hiding a three-row table.

And the payoff never arrived in the template — `proposal-form.html` still has three near-identical cards (tracks inline, time slots and facilitators via partials), each with its own hidden `*_submitted` input and its own checkbox loop. Tracks and time slots differ only in name and label.

The judo move: make the three rows data.

```python
@dataclass(frozen=True)
class _Picker:
    plural: str
    singular: str
    all_items: Callable[[Uow, int], Sequence[_HasPk]]
    stored: Callable[[Uow, int], Iterable[int]]

_PICKERS = (
    _Picker("facilitators", "facilitator", ..., ...),
    _Picker("tracks", "track", ..., ...),
    _Picker("time_slots", "time_slot", ..., ...),
)
```

`_render` loops over it. `_collect_track_ids`, `_collect_time_slot_ids` and `_collect_facilitator_ids` collapse into one `_collect(picker)`. The template loops over a single `pickers` context entry instead of three cards. That deletes roughly 60 lines of view and 35 of template, and it also removes the duplicate queries — right now `tracks.list_by_event` is fetched once in `_render` (via `_picker_context`) and again in `_collect_track_ids` during the write, same for time slots and facilitators.

### 2. Three copies of the same checkbox workaround

> ✅ **Fixed.** `checkbox-field.html` now accepts `checked_values` and does the membership test inside. All three call sites — `proposal-form.html`, `parts/time-slot-picker.html`, `parts/facilitator-picker.html` — are one `{% include %}` each.

```
{% if t.pk in assigned_track_pks %}
    {% include "components/checkbox-field.html" with ... checked=True label="" only %}
{% else %}
    {% include "components/checkbox-field.html" with ... label="" only %}
{% endif %}
```

This appears in `proposal-form.html`, `parts/time-slot-picker.html` and `parts/facilitator-picker.html`. It's a workaround for `checkbox-field.html` not accepting a membership test, and this PR multiplies it by three. Fix it at the component instead: accept `checked_values` and do `{% if value in checked_values %}` inside, or add a tessera checkbox tag. Then each call site is one `{% include %}`.

### 3. `checkbox-field.html` opens `<label>` in one `{% if %}` and closes it in another

> ✅ **Fixed.** Each branch is now a complete element. The shared `<input>` moved to `components/checkbox-input.html` so the labelled and unlabelled branches don't duplicate the attribute list.

```
{% if label %}<label ...>{% endif %}
    <input ...>
{% if label %}<span>{{ label }}</span></label>{% endif %}
```

A component that conditionally emits a closing tag for an element it conditionally opened. It works, but no formatter or reader can follow it — this PR's reformat of that block is exactly that confusion showing up. Give each branch a complete element (`{% if label %}` full labelled version `{% else %}` bare input `{% endif %}`), or split into two components.

### 4. The facilitator rule isn't the invariant the comment claims, and its error can vanish

> ✅ **Fixed, both halves.** The comment is reworded to say what it is — a create-time rule, not an invariant; `SessionContentEditService.apply` documents that an empty `facilitator_ids` list replaces the assignment, so an edit can still remove every facilitator.
>
> The vanishing error was real and is now the headline bug fix of this PR: form-level errors render above the form, outside the column flow, via `{% tessera_errors form %}`. Covered by `test_post_renders_facilitator_error_when_event_has_no_facilitators`. The danger border that was keyed on the generic error channel is gone rather than kept — same fragility, cosmetic payoff.

```python
# Invariant: a hand-added proposal always has at least one facilitator.
```

`_update` sets `facilitator_ids` straight from the picker, so an edit can take a proposal to zero facilitators immediately after create refused it. That's create-time validation, not an invariant — either enforce it in the service for both paths, or reword the comment.

Worse, the reporting channel: the rule reports through `form.add_error(None, ...)`, and `form.non_field_errors` is rendered *only inside the facilitators card*, under `{% if all_facilitators %}`. On an event with no facilitators, create POSTs re-render into the `{% elif not proposal %}` branch and the error is dropped on the floor. Binding the form's generic error channel to one specific card is fragile regardless — any future non-field error inherits the same disappearing act. Render `non_field_errors` at a fixed place in the form.

### 5. Unrelated tooling churn that weakens the local gate

> ✅ **Resolved.** The `mise.toml` `devcheck`/`fullcheck` changes were split out and landed on `main` separately. `pyproject.toml` and `poetry.lock` were restored from `main`, dropping ~380 lines of pure dependency reordering. **This PR now touches zero tooling files.**
>
> The `blok` codespell entry went with them — it turned out to be needed only for an untracked local scratch file, not for anything in the repo; `mise run lint:codespell` is green without it.
>
> ❌ **One sub-claim was wrong:** the reorder didn't leave `# BLACK` "dangling at the bottom of an empty section" — it moved the marker from the middle of the dev-dependency list to directly above `[tool.black]`, which is where it belongs. That part was an improvement, not churn.

`mise.toml`, `pyproject.toml` and 358 lines of `poetry.lock` have nothing to do with unifying two views:

- `devcheck` **dropped `lint:tingle`** and gained `test:py -xvv` — in the same PR that *adds* a tingle metric.
- `fullcheck` switched from `check` (which runs the full `lint` task) to `devcheck`, so it no longer runs `hk`, `taplo`, `codespell`, `ast-grep` or `impeccable`. CI still runs `mise run lint` so the merge gate is intact, but the local "run everything before pushing" task is now a strict subset of what it was.
- `fullcheck` now runs `poetry install` as a step — a check task that mutates the environment.
- Dependency reordering left the `# BLACK` marker dangling at the bottom of an empty section; `djlint` was quietly tightened `^1.39.2` → `~1.39.2`.

Please split this into its own PR. It's unreviewable buried under a template refactor, and the `devcheck`/`fullcheck` changes deserve their own argument.

---

### Should fix

> ✅ **`context_data=ANY` — fixed, and beyond the one flagged.** All six occurrences across `test_proposal_create_page.py` and `test_proposal_edit_page.py` now build the full expected context. `mock-ANY` is net **−4** on the branch. A `_facilitator_dto` helper was needed because `FacilitatorListItemDTO` requires an explicit `session_count` (the repo annotates it with `Count("sessions")`). `ANY` survives only on `form` and on `field_descriptors`, which holds the form's `BoundField`s.

**`context_data=ANY` in the new `test_get_renders_track_checkboxes`.** `tingle` flags it (`mock-ANY +1`, `test_proposal_create_page.py:542`) and it's a standing project rule. `_base_context(event)` plus `all_tracks`/`assigned_track_pks` is right there.

> ⏭️ **Two test modules — deferred → #684.** Agreed in principle.
>
> ❌ **The helper claim is partly wrong:** `_fields_url` and `_cancel_url` exist only in the edit module, not both. Only `_base_context` is genuinely duplicated — and it's duplicated across **16** modules under `tests/integration/web/panel/`, so it's a house pattern rather than this PR's debt. The issue records that as an open question. Note also that `test_session_field_value_upsert.py` subclasses `TestProposalEditPageView`, so whatever replaces the two modules has to keep its coverage.

**Two test modules for one view and one template.** `test_proposal_create_page.py` and `test_proposal_edit_page.py` each carry their own `_base_context`, `_fields_url` and now `_cancel_url`. The whole point of the change is that there is one flow — the tests should say so, and the duplicated helpers should be one module-level pair.

> ✅ **Fixed — via a shared base rather than module-level functions.** `_prepare` and `_add_field_context` pull in four more methods (`_load_session`, `_resolve_category`, `_build_form`, `_orphan_values`), all request-scoped; hoisting six methods to module level with an explicit `request` argument would have been more churn than signal. Instead both views now derive from `_ProposalFormBase`, which holds exactly those six and nothing else. The fragment endpoint no longer inherits the create/edit `post`, so the `http_method_names = ("get",)` fence is deleted — and `test_post_to_fields_component_is_rejected` proves POST still 405s with no side effects. Happy to go the module-level-function route instead if you'd rather.

**`ProposalFormFieldsComponentView(ProposalFormPageView)`.** Subclassing a page view purely to borrow `_prepare` / `_add_field_context`, then fencing off the inherited `post` with `http_method_names = ("get",)`, is inheritance used as an import. Neither borrowed method needs the view's identity. Make them module-level functions taking `request`, the way `session_field_requirements` and `collect_session_field_values` already are in that file, and both views stand alone.

> ✅ **Restored, naming both keys.** And the missing comment was hiding a real bug: in tag mode `select.html` takes the `attrs` branch, so `has_errors` reached only the red border — `aria-invalid` was stranded in the form-mode `{% else %}`. It's now emitted in both modes.

**`SelectNode` lost its contract comment.** The comment explaining that `class` styles the element and *every other keyword is forwarded as an HTML attribute* was deleted at exactly the moment a second silently-popped keyword (`has_errors`) was added. Now two keys are magic and nothing says so. Restore the comment naming both.

### Noted, pre-existing — but this was the moment

> ⏭️ **Deferred → #681.** Confirmed real and confirmed pre-existing: `main`'s `proposal-create.html` and `proposal-edit.html` both target `#proposal-session-fields` with `duration` and `participants_limit` outside it. Filed as a bug with the two fix options and the open question about whether an unsaved value should survive a category change.

**HTMX swaps too little.** The category select targets `#proposal-session-fields`, but `duration` and `participants_limit` are *also* category-derived (`create_proposal_form` builds `duration` as a `ChoiceField` when the category defines durations, and applies the category's participant bounds) and both live in the Session Details card, outside the swap target. Change the category and the duration picker stays stale; submit and the server validates against choices the user was never shown. Widening the swap target to the whole form body would fix it and would let the fields endpoint reuse the same template region instead of a bespoke partial.

> ✅ **Fixed.** It goes through `components/checkbox-field.html` now (the `blocktranslate` moved to `asvar` so the accessible name still reaches `aria_label`). `raw-form-controls` is net **−6** on the branch.

**`proposal-session-fields.html` still hand-rolls `<input type="checkbox" name="remove_field_ids">`** — the one picker the PR didn't convert, and the reason the new `raw-form-controls` metric can't reach zero.

---

**Verdict: changes requested.** Items 1–5 are the blockers. 1 and 2 are the same missed opportunity seen from two sides — the branch built the abstraction and then didn't spend it. 5 just needs to leave.

Note: I did not run the test suite for this review; `tingle check` was run and is green.

---

> **Updated verdict.** Of the five blockers, 2–5 are fixed and 5's tooling churn has left the branch entirely. 1 is the one genuine deferral, tracked as #683 — the abstraction is still stringly-typed, but nothing else depends on fixing it now. Both remaining deferrals (#681, #684) describe pre-existing behaviour rather than anything this branch introduced.
>
> This review was run again with the test suite: **3142 passed, 9 skipped**; ruff / djlint / mypy / pylint (10.00/10) / vulture / import-linter / codespell green; `tingle` no new debt, 94 paid off.

fancysnake:
### Rams design review — item status

The Rams summary is a bot comment I can't edit in place, so marking its four key issues here. All four inline threads are resolved.

| Issue | Status |
|---|---|
| 🟡 Field error text is never linked to its input (A11Y) | ⏭️ **#682** — confirmed, and design-system-wide rather than this page: `tessera/errors.py::render_errors` emits the `<p>` with no `id`, and nothing passes `aria_describedby` even though `text-field`, `textarea`, `select` and `stepper-field` all accept it. Fixing it inside `tessera_field` fixes every form at once. |
| 🟡 "Create one first" link discards the proposal form in progress (UX) | ✅ **Fixed** — `target="_blank" rel="noopener"`, as suggested. It's the only way out of an empty facilitator list and the form has no draft storage. |
| 🟡 Facilitators heading spacing breaks the page's rhythm (Spacing) | ✅ **Fixed** in 9b881a67 — every card heading is `mb-2` now; the Session Details outlier was aligned down rather than the rest up. |
| 🟡 Track checkboxes render as bare native controls (Components) | ✅ **Fixed** in 888dabba, which landed after the review snapshot — they go through `components/checkbox-field.html`, which already carries `accent-primary` and a focus ring. The suggested inline `class="mr-2 accent-primary …"` isn't needed. |

Two more things the review didn't flag that came out of the same pass:

- Form-level errors were rendered **only inside the facilitators card**, so creating a proposal on an event with no facilitators re-rendered the page with no message at all. They now render above the form.
- `has_errors` on `{% select %}` reached only the red border in tag mode; `aria-invalid` was stranded in the form-mode branch. Both cues fire in both modes now.

fancysnake:
### Follow-up pass — four small items closed

Re-validated every open item against the branch head. All four inline review threads were already resolved and stay resolved; the deferrals (#681, #682, #683, #684) are unchanged and still the right call. Four cheap things were left on the table, and they're fixed in `1b10ea20`:

| Item | Fix |
|---|---|
| `duration` was the one proposal-form field without an explicit label | `label=_("Duration")` on both `SessionEditForm.duration` and the `ChoiceField` override in `create_proposal_form`. The previous pass gave `title`, `display_name`, `description`, `contact_email`, `participants_limit` and `min_age` explicit labels and missed this one, so `{% tessera_field form.duration %}` fell back to Django's auto-label and the Polish panel read **"Duration"**. `msgid "Duration"` → `"Czas trwania"` already existed; the catalog only gained a source reference. Pre-existing on `main`, but it was this PR's own six-field sweep that skipped it. |
| Category error text not linked to its input | The error `<p>` gets `id="id_category_id_errors"` and the `{% select %}` tag gets `aria-describedby=form.category_id.errors\|yesno:"id_category_id_errors,"` — the house `id_<field>_errors` convention already used in `multiverse/panel/announcements/_form.html`. This is the one hand-rolled error paragraph on the page; the design-system-wide fix stays #682. |
| Heading rhythm fixed in one file, not the other | `parts/proposal-session-fields.html` still had `mb-6` on "Session Fields" and `mb-1`/`mb-6` on "Answers outside this category" while every card in `proposal-form.html` had moved to `mb-2` + `mb-4`. Same Rams spacing issue, one include over. Aligned. |
| Stale TODO entry | "migrate hand-rolled category select + facilitator/track checkboxes to tessera select/checkbox tags" was true when written and isn't now — the category select goes through `{% select %}` and all four checkbox call sites through `components/checkbox-field.html`. Removed; the aria-describedby entry stays. |

Also re-confirmed, so the record is straight:

- The branch touches **zero** tooling files — no `mise.toml`, `pyproject.toml` or `poetry.lock` in the diff.
- The PR adds **no** raw-markup test assertions. The ~13 `response.content.decode()` checks CodeRabbit's *User-Centric E2e Tests* keeps flagging are inherited from `main` in the two changed modules; worth sweeping when #684 merges them, but not this PR's debt. The check reports **pass** on the current head — the ❌ in the sticky summary is from the Jul 21 snapshot.
- Rams' *"Track checkboxes render as bare native controls"* suggestion is obsolete: applying the proposed inline `class="mr-2 accent-primary …"` would re-introduce a raw input. They go through `components/checkbox-field.html`, which already carries `accent-primary` and a focus ring.
- The picker double-query claim stays wrong for tracks and time slots (`_render` is failure-path, `_collect_*_ids` is write-path). One extra `facilitators.list_by_event` on an invalid-create re-render is the only real one.

Full suite green (3142 passed, 9 skipped), all linters green, catalog fresh with no fuzzy or untranslated entries, `tingle`: no new debt, 94 paid off.

Issue #642, opened 2026-07-24
Spike deterministic PO generation vs semantic PO merge driver


hasparus:
## Problem

Concurrent feature branches frequently conflict in the generated Polish catalog at `src/ludamus/locale/pl/LC_MESSAGES/django.po`. Keep Django/gettext and Varlock as the baseline, but compare two focused ways to reduce Git conflicts before adopting either.

This issue is about merge ergonomics. Rust extractor parity is tracked separately in zagrajmy/ludamus#640.

## Spike A: deterministic Django-compatible generation

Evaluate [`django-extended-makemessages`](<https://github.com/michalpokusa/django-extended-makemessages>) with the real Ludamus catalog and settings.

Candidate options:

* `--sort-by-msgid`
* `--no-location`
* `--no-wrap`
* `--no-fuzzy-matching`
* `--keep-header`
* `--no-previous`
* `--detect-aliases`
* retain format flags so `msgfmt` can validate placeholders

Work:

1. Run the command in a disposable branch/worktree.
2. Compare semantic identities `(msgctxt, msgid, msgid_plural)` with stock Django output; require zero missing and zero extra entries.
3. Confirm existing translations and plural forms are preserved.
4. Record the one-time normalization diff size.
5. Simulate two branches adding unrelated strings and measure whether stock Git merges the normalized catalog without conflict.
6. Confirm extraction and `--check` still run through the configured Django/Varlock environment.
7. Confirm Django 6, Python 3.14, macOS ARM64, and Linux CI compatibility.

## Spike B: PO-aware three-way merge

Evaluate both [`po3way`](<https://github.com/JulienPalard/po3way>) and [Weblate's gettext merge driver](<https://github.com/WeblateOrg/weblate/blob/main/weblate/examples/git-merge-gettext-po>); choose at most one candidate for the spike branch.

Work:

1. Build realistic base/local/remote fixtures from the Ludamus catalog.
2. Cover independent translations, independent new entries, header-only changes, source-string changes on one side plus translation changes on the other, plurals, contexts, flags, and obsolete entries.
3. Compare stock Git merge output with the PO-aware driver.
4. Run `msgfmt --check --check-format` on every successful merge.
5. Verify genuine same-message conflicts remain explicit rather than being silently resolved incorrectly.
6. Determine how installation and Git configuration would work through mise/bootstrap. Do not commit `.gitattributes` unless clones without the driver fail safely and setup is guaranteed.
7. Measure maintenance and contributor setup cost.

[...]

Issue #640, opened 2026-07-24
Fork django-makemessages-rs and reach Django extraction parity


hasparus:
## Problem

[`django-makemessages-rs` 0.3.2](https://github.com/zxzinn/django-makemessages-rs/tree/v0.3.2) is a promising fast replacement for Django/gettext extraction, but it is not yet output-compatible with Ludamus.

A read-only differential run against the current Polish catalog completed in about 0.19 seconds, but message identities differed:

- Django catalog: 1,657 identities
- Rust output: 1,649 identities
- Missing from Rust output: 23
- Extra in Rust output: 15

The comparison key was `(msgctxt, msgid, msgid_plural)` and ignored serialization details.

### Template contexts are discarded

Inline `{% translate "…" context "…" %}` entries are emitted without `msgctxt`. Ludamus currently uses contexts including `place`, `participant filter`, `wizard step`, `proposal review heading`, `table column`, and `Sidebar category`.

The extractor currently assigns `msgctxt: None` for inline template translations:
https://github.com/zxzinn/django-makemessages-rs/blob/0e9916437dfe2a4d577fa6491bda4a53d2af0f4e/src/extractor.rs#L341-L357

### Non-trimmed block translations change identity

The extractor calls `.trim()` for every block translation, while Django preserves surrounding whitespace unless the template explicitly uses `trimmed`. This produced both missing and extra identities for multiline `{% blocktranslate %}` messages.

Relevant implementation:
https://github.com/zxzinn/django-makemessages-rs/blob/0e9916437dfe2a4d577fa6491bda4a53d2af0f4e/src/extractor.rs#L362-L410

### Python extraction needs broader compatibility coverage

Python extraction uses regexes for fixed function names and `_()`:
https://github.com/zxzinn/django-makemessages-rs/blob/0e9916437dfe2a4d577fa6491bda4a53d2af0f4e/src/extractor.rs#L28-L58

Our codebase also contains direct aliases such as `gettext as _gettext`. Django's stock xgettext workflow does not detect every alias either, but our fork should either detect direct imports or explicitly document and test the supported subset.

## Proposed work

Fork the project under the Zagrajmy organization and establish differential compatibility before considering adoption:

1. Add fixtures for Django template `context` on `translate` and `blocktranslate` tags.
2. Match Django whitespace behavior exactly, including the distinction between ordinary and `trimmed` blocks.
[...]

PR #656, opened 2026-07-24
Add Panel enrollment and wait-list settings

hasparus:
## Summary

Two organizer-facing settings features, plus the gate carve-out they needed.

**Enrollment windows.** Sphere-scoped Panel CRUD for multiple enrollment windows per event, under a new event "Enrollment" settings tab. Event ownership is enforced on every read and mutation, in the service rather than the view, so a window id belonging to another event or sphere 404s without side effects.

**Waiting-list promotion.** Per proposal category: what happens when a seat frees up (confirm the next person automatically, or hold the seat until they confirm) and how long the claim window lasts.

**Gate carve-out.** Both features moved into the noun-based `event/panel` gate, replacing the legacy UoW context/access path with `EventPanelService`, `EnrollmentSettingsService`, `ProposalCategorySettingsService`, repository protocols, DTOs, and `request.services`. The proposal-category edit view in particular went from ~250 lines of `request.di.uow` sprawl in `chronology/panel/views/cfp.py` to a service with a repos bundle.

This lets Blood on the Skytower managers configure enrollment and offer-claim behaviour without Django staff access.

## Notable fixes from review

- **Partial POSTs no longer reset promotion config.** `promotion_mode` and `offer_claim_window_minutes` are `required=False`, and the update wrote them unconditionally — so any POST omitting them silently reset a category configured for `OFFER_CLAIM` back to `AUTO`/1440 minutes, with a success message. `ProposalCategoryData` is already `total=False`, so absent now means "leave as configured".
- **One requirement parser instead of two.** `parse_requirement_selection` now lives in the shared `gates/web/django/panel.py`, and scoping is `RequirementSelectionDTO.scoped_to()`, shared by the service and the legacy chronology views. This deletes `parse_field_requirements`/`scoped_requirements`, whose unguarded `int()` raised `ValueError` on a tampered non-numeric pk such as `field_abc=required`.
- **Enrollment view ceremony collapsed** — the tab-context block went 3× → 1, two near-identical `_render`s → 1, the not-found redirect 4× → 1, and the double event fetch on the invalid-form path is gone.
- Untranslated `"Delete window"` button and the duration widget's client-side `'h'`/`'min'`/`"No durations configured yet."` strings are now wrapped.
- `EnrollmentSettingsService` declares `EnrollmentSettingsServiceProtocol` so the type checker verifies conformance.

## Verification

- `mise run check` — exit 0
- `mise run test:py` — 3,168 passed, 9 skipped
- `mise run messages-check` — 1,658 translated, extraction current
- `tingle stat --diff` — net debt down: refactoring −807 (`request-di-uow` −42, `legacy-loc` −28), frontend −7
- Visually verified empty, create, list, edit, and category offer-claim states

Coverage added for the paths that carried real risk: the category write payload, `read_context` ordering, an omitted promotion config not being written, a category from **another event** rejected without side effects, and requirement parsing against non-numeric keys.

### Enrollment settings — empty state

Enrollment settings empty state

### Create enrollment window

Create enrollment window

### Enrollment windows

Enrollment window list

### Edit enrollment window

Edit enrollment window

### Category waiting-list settings

Category waiting-list settings

## Reviewer notes

- **`cfp-edit.html` is best read with `git diff -w`**, which reports `19 0`. The substantive change is the waiting-list card; the rest of the churn is djlint reindenting the file because the form is now wrapped in `<div class="p-4">` (matching `bans.html` and `content-log.html`).
- **The panel URLconf move is now tracked as a problem, not a win.** It relocated the whole 539-line URLconf under `event/`, where it imports 15 chronology view modules — see #699. Its ~539-line drop in `old-subdomain-loc` is relocation, not debt paid; the genuine reductions are `request-di-uow` −42 and `legacy-loc` −28.
- `EMAIL_TESTING_PLAN.md` had a real mailbox address replaced with a placeholder, but it remains in this branch's history (`09801ced`, `69c201f4`) on a public repo. Rotating or retiring that test account is the practical fix.

## Follow-ups

- #699 — panel URLconf lives under `event/` and owns 15 chronology view modules
- #700 — `EnrollmentWindowDTO` and `EnrollmentConfigDTO` duplicate one table
- Production SMTP/DBOS offer delivery and expiry verification tracked in `EMAIL_TESTING_PLAN.md`
- Rust gettext extraction parity tracked in #640
- Deterministic PO generation versus a semantic merge driver tracked in #642
- Smaller, unblocking: `_HasPk` is still defined twice, the status pill wants a `components/status-badge.html` partial, and the two `EventContextMixin` classes share a name

## Summary by CodeRabbit

* **New Features**
  * Added event “Enrollment” panel with enrollment windows (list, create, edit, delete), including capacity %, waitlist limits, access restrictions, anonymous enrollment, and participant notices.
  * Expanded proposal category settings with promotion mode and a configurable offer-claim window, plus client-side updating of the claim window section.
* **Documentation**
  * Added an end-to-end email-flow validation runbook; updated product and architecture documentation.
* **Bug Fixes**
  * Improved backoffice panel permission handling, context loading, and event panel routing/tab wiring; refreshed Polish UI/backoffice translations.
* **Tests**
  * Added integration/unit coverage for enrollment windows, proposal-category claim settings, form parsing, and event panel service behavior.


hasparus:
## Code quality review — maintainability pass

The carve-out is good work: `CFPEditPageView`'s 240-line `request.di.uow` sprawl is now a real service with a repos bundle, `build_panel_stats`/`is_proposal_active` are deduped, and the enrollment integration tests genuinely prove event scoping. Three blockers though, plus a data-loss bug and two test gaps. Detail below; the larger structural items are called out as follow-ups rather than fixed here.

### Blockers

**1. `src/ludamus/gates/web/django/forms.py.orig` — 875-line merge backup committed into `src/`**
Byte-identical to `origin/main:src/ludamus/gates/web/django/forms.py` — a merge-tool artifact. `.orig`/`.rej` aren't gitignored, so it would sit next to the real module forever and catch every grep and jump-to-definition.

**2. `templates/panel/enrollment-settings.html:94` — untranslated user-facing button**
```
{% tessera_button "Delete window" variant="danger" size="sm" %}
```
Line 92 translates the *same* string for `data-confirm-action`, and every other label on the page uses `{% translate … as label %}`. A raw literal isn't extracted by `makemessages`, so PL users get English "Delete window" next to a Polish confirm dialog.

**3. `mills/enrollment.py:97` — `EnrollmentSettingsService` doesn't declare its Protocol**
`EnrollmentSettingsServiceProtocol` (`pacts/enrollment.py:206`) exists solely for it, and both sibling services added in this same branch (`ProposalCategorySettingsService`, `EventPanelService`) declare theirs. Nothing type-checks conformance today.

### Should fix

**4. `proposal_category_settings.py:61-64` — silent defaults clobber saved config**
```python
promotion_mode=PromotionMode(cleaned["promotion_mode"] or PromotionMode.AUTO),
offer_claim_window=timedelta(minutes=cleaned["offer_claim_window_minutes"] or 1_440),
```
Both fields are `required=False` (`forms.py:334,347`) and `update` writes unconditionally. Any POST omitting them resets a category configured for `OFFER_CLAIM` / a 30-minute window back to `AUTO` / 1440 — silently, with a success message. The existing tests that post only `{"name": …}` do exactly this.

**5. `templates/panel/cfp-edit.html` — 1215 diff lines for 19 real ones**
`git diff -w --numstat` → `19 0`; the rest is a full-file reindent. It buries the one substantive change (the waiting-list card) and will conflict with every other in-flight change to this file.

**6. `tests/unit/test_proposal_category_settings_service.py` — never asserts the write**
142 lines, one test. Asserts the three `set_*_requirements` calls but never `categories.update`, so the `ProposalCategoryData` mapping (`proposal_category_settings.py:102-115`) — including `promotion_mode` and `offer_claim_window`, the two fields this work exists to add — has zero unit coverage. No unit test for `read_context` or `_sort_by_order` either, though CLAUDE.md puts `mills` under unit tests.

**7. Foreign `category_slug` scoping is correct but unproven**
`read_by_slug` filters `event_id` and the service resolves the category before every read and write, so the rule is satisfied *in code*. But `test_get_redirects_on_invalid_category_slug` / `test_post_redirects_on_invalid_category_slug` both use the literal slug `"nonexistent"` — neither uses a category that exists **on another event**. The body-id equivalent exists (`test_post_drops_requirement_pks_from_another_event`); the URL-id one doesn't.

**8. `chronology/panel/views/base.py:76` — dead `get_current_event` added**
All four callers resolve against `event/panel/views/base.py:52`. It's also a trap: the legacy version costs `list_by_sphere` + `read_by_slug` + `get_stats_data`, the new one a single `read_by_slug`.

**9. `enrollment_settings.py:113-130` — two error paths for one condition, one unreachable**
`post` pre-reads the window (erroring if `None`), then re-checks `updated is None` with the same message. The second branch is effectively unreachable; `update_window` already returns `None` for a foreign/missing pk.

### Follow-ups (deliberately not in this branch)

- **`event/panel/urls.py` owns the whole panel URLconf.** The 539-line file moved wholesale and now imports 15 chronology view modules (timetable, venues, discounts, imports, facilitators…). The `event` name no longer describes it and the dependency points the wrong way for a strangler migration. It also shaves ~539 off `old-subdomain-loc` purely by relocating a file — that number is not real debt paid. Should become a neutral aggregate URLconf that `include()`s a small `event/panel/urls.py`, so each area's routes migrate with its views.
- **`EnrollmentWindowData`/`DTO` clone `EnrollmentConfigDTO`.** Same ten fields, same `EnrollmentConfig` table, and `EnrollmentWindowRepository` is now a second repo over it alongside `EnrollmentConfigRepository`. One table, two DTOs, two repos, two names for one concept. "Window" is the better name; consumers should migrate onto it.
- **Two live implementations of parse-requirements-and-scope.** New `parse_requirement_selection`/`_scope_selection` vs existing `parse_field_requirements`/`scoped_requirements` (still used by `session_fields.py`, `personal_data_fields.py`). They already disagree: the new parser guards `raw_pk.isdigit()`, the old one calls `int()` unguarded and raises `ValueError` on `field_abc=required`.
- **`enrollment_settings.py` ceremony repeats 7×** — "resolve event or bail" ×7, the `active_nav`/`active_tab`/`tab_urls` block ×3, the two `_render`s differ only in `window=None`, the not-found redirect ×4. An `EventScopedView.dispatch()` setting `self.current_event` would collapse most of it and fix the double event fetch on the invalid-form path.
- **Two form→DTO conventions in one branch**: `EnrollmentWindowData.model_validate(form.cleaned_data)` (implicit) vs hand-mapping every field (explicit). Pick one; explicit is the greppable choice.
- Minor: `_HasPk` now has three copies; the status pill is a fifth utility copy-paste that wants `components/status-badge.html`; two classes named `EventContextMixin` with the same `get_event_context` contract invite silent import-site mistakes.

### Verified good

- Scoping is satisfied in code for both features — `read_by_slug(event_id, slug)` and every enrollment-window repo method filter on `event_id`, all inside the service.
- `test_cannot_edit_window_from_another_event` is exactly the test the scoping rule asks for: same sphere, different event, asserts no side effect.
- `tests/unit/test_enrollment_settings_service.py` uses a real fake declaring the repo protocol with real assertions. (Caveat: `EnrollmentSettingsService`'s five methods wrap single repo calls in `atomic()` — which a single ORM write already is — so `test_cannot_update_or_delete_another_events_window` would pass with the service deleted. Worth asking whether it earns its keep.)
- `RequirementSelectionDTO` and `ProposalCategorySettingsRepos` are justified — the former replaces a `tuple[dict, list]` that always travels and gets scoped together, the latter follows the `ImportRepos`/`FacilitatorPanelRepos` precedent.
- No new `request.di.uow` in the new views. No migration needed: `promotion_mode`/`offer_claim_window` already exist on `main`; only `_SIMPLE_UPDATE_FIELDS` gained them.

hasparus:
## Review follow-up — `0ece18fe`

`mise run check` exit 0, 3165 passed / 9 skipped.

### Fixed

| # | Finding | Fix |
|---|---|---|
| 1 | `forms.py.orig` committed | Deleted; `*.orig`/`*.rej` added to `.gitignore` |
| 2 | Untranslated "Delete window" | `{% translate … as delete_label %}` |
| 3 | `EnrollmentSettingsService` didn't declare its Protocol | Declares `EnrollmentSettingsServiceProtocol` |
| 4 | Silent reset of promotion config | Omitted → `None` → key left out of `ProposalCategoryData` (already `total=False`), so the repo leaves the column alone |
| 6 | Unit test never asserted the write | Asserts the full `categories.update` payload; adds a `read_context` ordering test and one proving an omitted promotion config is not written |
| 7 | Foreign `category_slug` unproven | `test_post_rejects_category_from_another_event` — real category on a second event in the same sphere, POSTed under the first event's slug, asserts "Category not found." and that the foreign row is unchanged |
| 8 | Dead `get_current_event` | Deleted from the legacy chronology mixin |

### Disputed

**9 — `enrollment_settings.py:126` is not a redundant branch.** The pre-read at 113-118 is load-bearing: the invalid-form path renders `_render(window=window)`. And `update_window` returning `None` *after* a successful read is a real delete-between-read-and-write race, not dead code. Removing the second check would make a concurrent delete report "Enrollment window saved." — strictly worse. Left as is.

**5 — the `cfp-edit.html` churn is not a stray reformat.** `git diff -w` is `19 0` because the substantive change wraps the form in `<div class="p-4">`, and djlint reindents everything inside it. The wrapper matches `bans.html` and `content-log.html`. Reviewers should read this file with `git diff -w`; splitting the reindent into its own commit would need a history rewrite for no code change. (Fair point that `cfp-create.html` has no such wrapper — the two pages should agree, but that's a cosmetic call for the author, not a merge blocker.)

### Not addressed (follow-ups, deliberately out of scope here)

The structural items from the review stand and want their own PRs — the `event/panel/urls.py` ownership problem, the `EnrollmentWindowDTO`/`EnrollmentConfigDTO` duplication over one table, the two live parse-requirements implementations (including the unguarded `int()` in the old one), and the 7× repeated resolve-event ceremony in `enrollment_settings.py`. Each touches code outside this branch's diff.

Also left alone: `EMAIL_TESTING_PLAN.md` and `PRODUCT.md` at repo root. Both are real authored content rather than junk, so removing or relocating them is the author's call — but they are unrelated to this PR's title, and the runbook's live "In progress" status table will go stale once merged.

hasparus:
## Review round 2 — `33e542fa`, `d8bb7f4e`

`mise run check` exit 0, 3168 passed / 9 skipped. `tingle stat --diff`: refactoring −807, frontend −7, testing −1.

### CodeRabbit

| Finding | Outcome |
|---|---|
| Real Gmail address in `EMAIL_TESTING_PLAN.md` | **Fixed** — 5 occurrences → `<test-mailbox>`. Repo is public, so this was worth doing. Note the address is still reachable in history (`09801ced` and later); scrubbing that needs a force-push, and rotating or retiring the test mailbox is probably the better answer. |
| Preserve offer-claim settings on partial posts | Already fixed in `0ece18fe` — CodeRabbit marked it resolved itself. |
| `EnrollmentSettingsService` doesn't declare its protocol | Already fixed in `0ece18fe`; the review was against an older commit. |
| Untranslated `'No durations configured yet.'` and `'h'`/`'min'` in the duration script | **Fixed** — wrapped via `{% translate … as … %}` + `escapejs`. Pre-existing on `main` rather than new here, but it's a real bug in a file this PR already touches. |
| Keep `inits` imports lazy | **Rejected.** `inits/services.py` imports ~50 services eagerly at module top level — that is the file's established convention, and the four additions follow it. CLAUDE.md's `(anything) -> inits -> (nothing) (top level)` is layer-direction notation, i.e. *nothing imports `inits`*; it is not a rule about import placement. Making 4 of 54 imports lazy would be inconsistent for no benefit. |

### Thermo-nuclear — the two cheap structural items

**Requirement parsing is now one implementation.** `parse_requirement_selection` moved to the shared `gates/web/django/panel.py` (neutral ground — `chronology/panel` should not import from `event/panel`), and scoping became `RequirementSelectionDTO.scoped_to()`, so the mills service and the legacy chronology views share it instead of the service having a private `_scope_selection`. Deleted `parse_field_requirements` and `scoped_requirements`, migrated their 6 call sites, and dropped the orphaned `_HasPk`. This also fixes the latent crash: the old parser called `int(key.removeprefix(prefix))` unguarded, so `field_abc=required` raised `ValueError`; three unit tests in `test_panel.py` now cover parsing, non-numeric keys, and scoping.

**Enrollment ceremony collapsed.** `EnrollmentSettingsViewMixin` holds `tab_context()`, `render_window_form()`, and `window_not_found()`. The `active_nav`/`active_tab`/`tab_urls` block went 3× → 1, the two near-identical `_render`s → 1, and the not-found redirect 4× → 1.

I did *not* hoist event resolution into `dispatch()` as suggested. `dispatch()` would resolve via `get_current_event`, and the render paths need `get_event_context` (different service call, richer payload) — so every rendering view would fetch the event twice. `tab_context()` returns context and event from one call instead, which also removes the double fetch the review flagged in `EnrollmentWindowEditPageView.post`.

### Filed rather than fixed

- #699 — panel URLconf lives under `event/` and owns 15 chronology view modules (includes the note that its ~539-line `old-subdomain-loc` drop is relocation, not debt paid)
- #700 — `EnrollmentWindowDTO` / `EnrollmentConfigDTO` duplicate one table

Still open from round 1 and unchanged: `_HasPk` remains defined twice (was three times), the status pill wants a `components/status-badge.html` partial, the two `EventContextMixin` classes share a name, and `EnrollmentSettingsService`'s methods are thin enough to be worth questioning. `EMAIL_TESTING_PLAN.md` and `PRODUCT.md` are still at repo root — author's call.

hasparus:
## Code quality review — round 5 (`5c389dee`): approved

Rounds 2–4 blocked; this round clears. `mise run check` exit 0, 3,204 passed / 9 skipped, PL catalog fresh at 1,661, `tingle stat --diff` net **−783** (`request-di-uow` −39, `old-subdomain-loc` −695).

### What the earlier rounds caught

Worth recording, because two were real defects rather than style:

**A capacity leak, reverted.** An earlier commit unioned overlapping enrollment windows (`restrict_to_configured_users` via `all()`). That turned out to leak: with restriction unioned, a restricted window stops gating the event, but `get_available_slots` still draws from `get_most_liberal_config` — the highest `percentage_slots` window, chosen with no regard for restriction. A restricted 100% window overlapping an open 20% window therefore hands the restricted pool to everyone. Making it safe needs per-actor capacity across all five `get_most_liberal_config` call sites, several of which have no actor in scope — an enrollment redesign, not a panel-settings change. Reverted in `6fd4a067`; the combination inconsistency is now **#705**. The misleading overlap copy stays removed.

**Superuser lockout.** `EventPanelAccessMixin.test_func` checked `is_manager` only, while every other panel view uses `has_panel_access`, which also admits `PanelAccess.SUPERUSER`. Moving cfp-edit onto the new mixin silently revoked non-manager superuser access. Fixed, and `test_get_ok_for_manager_and_superuser` now runs the `panel_access_user` fixture across both roles.

Then the root cause of that one: `PanelAccessMixin` and `EventPanelAccessMixin` were byte-identical apart from the `request` annotation, and the divergence is what dropped the superuser check. `PanelAccessMixin` now subclasses `EventPanelAccessMixin`, matching how `EventContextMixin` already narrows `request`.

### Also landed this round

- Field services take `RequirementSelectionDTO` and reuse `scoped_to` instead of re-deriving the pk filter.
- Two abstract hooks collapsed to one: `set_*_categories` is `add_*_to_categories` plus a delete that is a no-op on a freshly created field (verified in the repository before relying on it). That orphaned both `add_*_to_categories` methods and their protocol declarations — deleted, taking four copies of the link loop down to two.
- `CFPFieldCategoryService` is a proper `abc.ABC`; `TestCFPSessionFieldService` guards which link table each operation writes to.

### Carried forward, not blocking

- `session_fields.py` now mixes `di.uow` reads with `services` writes, against the invariant in `inits/middleware.py`. Lifting `list_summaries` / edit-context / `delete` onto the shared base would drop the remaining 11 `di.uow` uses and make the wide `CFPFieldRepositoryProtocol` earned rather than incidental.
- `_read_page` takes three positionals, two of them adjacent same-typed slugs — a swap type-checks silently.
- Two distinct classes named `EventContextMixin`, distinguished only by import alias.
- `InvalidEnrollmentWindowError` is unreachable in production (the form validates first) and uncaught; `specs/enrollment.py`'s docstring is stale.
- `architecture.md:117`'s services example still calls the now-deleted `add_field_to_categories`.
- `PRODUCT.md` is an orphan with an unfilled `## Register` section; the runbook's live progress table will be stale on merge.

Deferred by design: **#699** (panel URLconf under `event/`), **#700** (`EnrollmentWindowDTO`/`EnrollmentConfigDTO`), **#705** (enrollment window combination).

PR #676, opened 2026-07-25
Make anonymous enrollment error handling exhaustive

hasparus:
Applies learning 2 from the Effect-TS audit (`docs/audits/effect-ts-learnings-2026-07-25.md` on `claude/effectts-ludamus-learnings-8i67gm`): per-error-code branching should be a `match` checked by `typing.assert_never`, not an enum-keyed dict that `KeyError`s at runtime when a code is added.

## Changes

- `gates/web/django/chronology/anonymous.py`: the `_ERROR_MESSAGES` dict becomes `_error_message()`, a `match` over `AnonymousEnrollmentErrorCode` guarded by `assert_never`. Split into target-side and visitor-side halves with `Literal`-narrowed parameters to stay under the complexity limit — the split stays exhaustive as a whole: a new enum member is a mypy error until every branch handles it.
- The `AnonymousEnrollOutcome` if/elif chain in the enroll POST handler gets the same `match` + `assert_never` treatment.

## Bug found by the change

The totality check immediately caught a real gap: `NO_ENROLLMENTS` had no entry in the old message table. It is currently only raised by `load_by_code` (which handles it specially), so the `KeyError` was latent, but any future path raising it through `_error_response` would have crashed in the error path. It now maps to the same message the load view uses.

No new strings — all messages keep their existing msgids, so translations are unchanged. Behavior is covered by the existing 50 integration tests for the anonymous enrollment views (all passing), and the exhaustiveness itself is enforced by mypy rather than a runtime test.

No UI changes, so no screenshots.

https://claude.ai/code/session_01RPGpyek44MXUam6fcj4Yof


hasparus:
## CI fix pass (commit `2b63c95`)

### "checks" job — root cause: stale PL message catalog
`messages-check` failed because `src/ludamus/locale/pl/LC_MESSAGES/django.po` wasn't regenerated after this PR added the `NO_ENROLLMENTS` message — its `#:` entry sat at the wrong position relative to `makemessages`' source-scan order (comparing against a fresh extraction diffed at line 749 vs 793). Fixed by running `makemessages -l pl --add-location=file` and committing the regenerated catalog.

### "tingle" job — root cause: `old-subdomain-loc` debt regression (+48 net lines)
Replacing the enum-keyed dict with `match` + `assert_never` is the right fix (it's what surfaced `NO_ENROLLMENTS` had no message in the first place), but it added 48 net lines to a file tracked by tingle's legacy-naming-migration metric (`chronology` → `event`, etc.), and the branch had nothing else to offset it — `tingle check` (policy `sum`) failed outright.

I found genuine slack, not a metric-gaming trick: `_error_message` was a pure 2-arm dispatcher with no logic of its own, delegating to `_target_error_message`/`_visitor_error_message`. Merging its fallthrough directly into what was `_visitor_error_message` removes that whole extra function — mypy still verifies the two halves are exhaustive together via the exact same fallthrough-narrowing it already relied on for the old dispatcher (verified: `mypy src` clean, `pylint` 10/10, `black` clean at every step). Also inlined the now-single-use `_TargetErrorCode` alias and trimmed two comments.

Result: **+48 → +21** net lines, confirmed with `tingle stat --diff` after each change.

**The remaining +21 is flagged, not fixed**: with `max-statements=7` / `max-returns=6` / mccabe `<=10` in `.pylintrc`, 10 distinct error codes structurally cannot fit in one function (verified empirically — a single 10-arm match hits mccabe 12; an `if code in (...)` narrowing trick fails mypy's `arg-type` check, confirmed with a throwaway repro). Two functions is the minimum. I did *not*:
- weaken the exhaustiveness guarantee (e.g. a dict lookup gated by a looser check) — that would silently reintroduce the exact `NO_ENROLLMENTS`-shaped bug this PR fixes, and
- revert the parallel `AnonymousEnrollOutcome` match/`assert_never` conversion (a legitimate, related exhaustiveness improvement, not a bug fix) just to shave ~7 more lines,

since both would be gaming the counter rather than paying down real debt. tingle has no waiver/baseline mechanism (checked `tingle --help`; only `stat`/`check`/`report`/`list`/`add`/`init`), so `mise run lint:tingle` will still exit 1 on this branch. That's a maintainer call — either accept the debt explicitly (per this repo's own stated policy: "a justified addition is fine, say so in the PR") or direct further changes.

### Codecov patch coverage (77.08% → 91.3%)
The bot's flagged lines were real: `NO_ENROLLMENTS` had no test hitting its message. Root cause — `load_by_code` (the only raiser of `NO_ENROLLMENTS`) is called only from `AnonymousLoadActionView.post`, which catches the error with its own hardcoded message *before* `_error_message()` is ever invoked for that code. So that branch is unreachable through any current view/URL — codecov was right that it was untested, but no integration test can reach it.

Added `tests/unit/gates/web/django/chronology/test_anonymous.py`: a parametrized unit test hitting `_error_message` directly for all 10 codes (module-attribute access, `anonymous._error_message(...)`, not a `from ... import _error_message` — ruff's `import-private-name` rule blocks the latter). This is sanctioned by CLAUDE.md's "pure IO-free helper functions... may be unit-tested wherever they live" exception, since this one code has no HTTP path to reach it via an integration test.

Verified with `diff-cover coverage.xml --compare-branch=main` after a full `pytest -n auto tests/integration tests/unit` run (3143 passed, 9 skipped): **91.3%** patch coverage on `anonymous.py`, missing only lines 57-58 and 261-262 — the two `assert_never` fallback branches, unreachable by construction (hitting them would mean mypy's exhaustiveness check was wrong).

### Verification run
- `mise run check` (format + lint incl. mypy, pylint, ruff, black, djlint) — green
- `mise run messages-check` — green (was the "checks" job failure)
- `mise exec -- poetry run tingle check` — still fails net +21 (down from +48), flagged above
- Full `pytest -n auto tests/integration tests/unit` — 3143 passed, 9 skipped
- `diff-cover` vs `main` on the touched file — 91.3%

Not merging — leaving that and the tingle debt call to a maintainer.

---

PR #678, opened 2026-07-25
Retry transient HTTP failures in links clients

hasparus:
> [!NOTE]
> Waiting for @Fancysnake to review and merge. Neither hasparus nor his Claude will be merging this PR.

Applies learning 5 from the Effect-TS audit (`docs/audits/effect-ts-learnings-2026-07-25.md` on `claude/effectts-ludamus-learnings-8i67gm`): retry policy as one shared, declarative value — Effect's `Schedule` translated honestly into a small requests/urllib3 helper, not a framework.

## Changes

- New `links/retry.py`:
  - `mount_retries(session)` mounts an `HTTPAdapter` with a urllib3 `Retry` — 2 retries, exponential backoff (0.5s base), GET-only, `502/503/504` plus connect errors. 429 is deliberately excluded (a rate-limited caller should back off at the flow level). `raise_on_status=False` keeps each caller's existing `ok`/`raise_for_status` handling as the single failure path.
  - The policy is bounded against every slow-failure mode: `read=0` (a hung response fails after one read timeout, never stacked), `respect_retry_after_header=False` (a server-sent `Retry-After: 60` can't sleep the worker), and `bounded_timeout(read_seconds)` returning a split `(3.05s connect, read)` tuple so retried connect attempts against a SYN-dropped upstream cap at ~3s each instead of the full read budget. Worst-case added latency is the documented backoff (0.5s + 1s), not multiples of 30s inline in the enrollment request path.
- `links/ticket_api.py`: `MembershipApiClient` switches from module-level `requests.get` to a mounted session with `bounded_timeout` — one transient blip no longer fails the membership lookup.
- `links/google_docs.py`: `_build_session` mounts the same policy on both `AuthorizedSession`s (importer + sheet writer); all call sites pass `bounded_timeout`.

## Tests

Two integration tests use `responses` with an `OrderedRegistry` and `assert_all_requests_are_fired=True`, pinning the attempt count in both directions: a 503 followed by a 200 succeeds on retry, and three consecutive 503s exhaust the retries and raise `MembershipAPIError`. Existing ticket-API tests re-target the client's session; google_docs tests assert the split timeouts. All 203 links integration tests pass; mypy clean (225 files), ruff and black clean, importlinter contracts kept, tingle adds no debt.

The branch went through four rounds of adversarial maintainability review; rounds 1–3 each surfaced a real stall-class issue (vacuous retry test, stacked read timeouts, uncapped `Retry-After` sleeps, stacked connect timeouts), all fixed and re-verified empirically — including a kernel-level SYN-drop probe showing ~10s bounded failure where the naive policy leaked ~90s.

No UI changes, so no screenshots.

https://claude.ai/code/session_01RPGpyek44MXUam6fcj4Yof


hasparus:
Found this PR flagged as inactive (no commits in 2 days, failing CI) by an automated stale-PR sweep. Reviewed the unaddressed CodeRabbit/Rams comments first — both were just rate-limit notices with no actual review content, nothing to act on there. Then diagnosed and fixed the two real CI failures:

1. **`checks` (pylint) failure** — `tests/integration/links/test_ticket_api.py` used `@responses.activate(registry=registries.OrderedRegistry, assert_all_requests_are_fired=True)`. Pylint/astroid can't resolve the parametrized-decorator overload of `responses.activate` (every other call site in the repo only uses the bare `@responses.activate` form, so this was the first to hit it). Switched both retry tests to `responses.RequestsMock(...)` as a context manager instead — same behavior, and pylint resolves it fine against the real constructor. No `pylint: disable` needed.

2. **`test` (pytest) failure** — `tests/integration/web/panel/test_discounts_page.py::TestDiscountExportPageView::test_post_exports_scoped_rows_and_redirects` asserted `GoogleSheetsWriter`'s `session.put` was called with `timeout=30`. This branch changed that call site to `timeout=bounded_timeout(30)` (a `(3.05, 30)` tuple) and updated the sibling assertions in `test_google_docs.py` accordingly, but missed this one in `test_discounts_page.py`. Updated it to match.

No merge conflicts with `main` (merges cleanly as-is, so I didn't touch the branch history otherwise).

Verified before pushing: full suite green (3135 passed, 9 skipped), pylint 10.00/10 on touched files, ruff/black clean, `tingle check --base origin/main` reports no new debt.

Not merging — per the PR description, this is waiting on @fancysnake's review.

---

fancysnake:

The policy itself is well-chosen — GET-only, `raise_on_status=False` so each caller keeps
one failure path, `read=0` so a hung response can't stack timeouts (I verified that one
against urllib3 2.7: a read error decrements the read counter to `-1` and exhausts
immediately, as the comment claims). Right layer, no new dependency, nothing crosses 1k
lines, `tingle check` clean.

But the shape of the change is wrong in one specific way, and the tests that are supposed
to guard it don't.

---

### 1. `bounded_timeout` spreads a session-level concern across 11 call sites

The module's own comment says the split timeout exists **"for sessions carrying the retry
policy"** — it is a property of the transport, not of any individual request. Yet it is
applied by editing every call site in three files, and it drags an import of
`ludamus.links.retry` into `tests/integration/web/panel/test_discounts_page.py`, a panel
test that has no business knowing how the HTTP layer spells a timeout.

The retry policy already lives in the adapter, where it is invisible to callers. The
connect timeout belongs in exactly the same place:

```python
class _RetryingAdapter(HTTPAdapter):
    def send(self, request, **kwargs):
        timeout = kwargs.get("timeout")
        if isinstance(timeout, int | float):
            kwargs["timeout"] = (_CONNECT_TIMEOUT_SECONDS, timeout)
        return super().send(request, **kwargs)
```

`Session.send` calls `adapter.send(request, **kwargs)` with everything keyword, so this is
a two-line override with no signature risk. `responses` patches `HTTPAdapter.send` on the
class, so `super().send(...)` still hits the mock — the new retry tests keep working.

What that buys:

- `retry.py` exports **one** name instead of two.
- `google_docs.py` diff shrinks to one import plus one line; `ticket_api.py` to one import
  plus two. All 11 `timeout=bounded_timeout(N)` edits disappear.
- `test_discounts_page.py` is untouched.
- The google_docs tests go back to asserting `timeout=10` — a real number — which fixes
  finding 2 for free.

Objection I'd expect: "rewriting the timeout inside an adapter is implicit." It is, but
strictly less implicit than silently re-sending the request three times, which this PR
already puts there without complaint. If adapter-level retries are acceptable, an
adapter-level connect timeout is too. Putting one in the transport and the other at every
call site is the inconsistent option.

### 2. The timeout assertions are vacuous — and the PR body claims otherwise

The PR says *"google_docs tests assert the split timeouts."* They don't. Every one of them
computes the expected value with the production function:

```python
google.session.get.assert_any_call(SHEETS_META_URL.format(...), timeout=bounded_timeout(10))
```

Both sides move together, so nothing is pinned. I checked rather than assumed: set
`_CONNECT_TIMEOUT_SECONDS = 999.0` and run the affected files —

```
98 passed in 18.92s
```

A 999-second connect timeout ships green. That is `ANY` in the connect slot, which is
exactly what the house testing rule forbids for simple values. Assert the literal
`timeout=(3.05, 10)`, or take finding 1 and let these read `timeout=10` again with the
`(3.05, ...)` pairing pinned once, where it's produced.

### 3. `MembershipApiClient` builds a `Session` per request, on every path, and never closes it

```python
def __init__(self) -> None:
    ...
    self.session = mount_retries(requests.Session())
```

`DependencyInjector` is request-scoped (`inits/legacy.py`), and `inits/services.py`
constructs a fresh `MembershipApiClient()` inside `EnrollmentService`. So every enrollment
request builds a `Session` plus two `HTTPAdapter`s plus two connection pools, uses it once,
and drops it — the one thing a `Session` is for, connection reuse, never happens.

Worse, it is built in `__init__`, *before* the `if not self.base_url` guard.
`MEMBERSHIP_API_BASE_URL` defaults to `""` (`edges/settings.py`), so on any deployment
that hasn't configured the membership API — and in every test that doesn't — we now
allocate a pooled session for a code path that returns before making a request. The old
`requests.get` at least only built its throwaway session when a call actually happened.
That part is a small regression, not just a missed optimisation.

Minimum fix: make it a `cached_property` so the unconfigured path costs nothing. If you
want the pooling you're now paying for, a module-level session mounted once is the real
answer — with the caveat that `requests.Session` isn't documented as thread-safe, so make
that call deliberately rather than by accident.

### 4. `backoff_factor=0.5` is half-inert, and the documented latency is wrong

Measured against the installed urllib3 2.7.0 with this exact `Retry`:

```
total attempts 3 | sleep before each retry [0, 1.0]
```

urllib3 2.x returns 0 backoff for the first retry (`consecutive_errors_len <= 1`). So the
first retry fires **immediately** — a 502 from a struggling upstream gets re-hit with no
delay at all, which is the thing a backoff exists to prevent — and `backoff_factor` only
ever controls the second attempt. Worst-case added latency is 1.0s, not the "0.5s + 1s"
stated in both the module comment and the PR description.

Not a stall risk, and arguably fine. But the comment block in `mount_retries` is nine
lines of prose whose entire job is to be the authority on this policy; it should describe
what the code does. Fix the number, and decide whether an instant first retry is the
intent.

### 5. Why is 500 not retryable?

`_RETRYABLE_STATUSES = {502, 503, 504}`. The comment explains the 429 exclusion but is
silent on 500. Google's own Sheets/Forms API guidance is exponential backoff on 5xx
generally — `backendError` arrives as a 500 and is transient. For the client this PR is
mainly aimed at, that's the status you most want covered. Either add it or say why not,
the same way you did for 429.

### 6. One test file, two mocking idioms — and the older one no longer tests the code

`test_ticket_api.py` now has three tests patching `client.session.get` and two using
`responses`. `patch.object(client.session, "get")` replaces the method *above* the adapter,
so those three are structurally blind to everything this PR adds. The retargeting from
`ludamus.links.ticket_api.requests.get` to `client.session.get` kept them passing without
making them mean anything more.

Since `responses` is already imported in the file, move them. The request-exception one in
particular gets strictly better as a `responses` connection-error case: it would keep
proving the email stays out of the logs *and* prove a non-retryable failure doesn't turn
into a retry storm.

---

### Verdict

Request changes on **1** and **2**. They're the same defect seen from two sides: a
transport-level concern was implemented at the call sites, and the tests meant to hold it
in place assert nothing because they were written in the same shape. Fixing 1 collapses 2.

**3**–**6** are should-fix; **3** is the one with a real (if small) production cost.

Credit where due: the read-timeout and `Retry-After` reasoning is sound and I confirmed
both against the installed urllib3. The retry-exhaustion test with `OrderedRegistry` and
`assert_all_requests_are_fired=True` genuinely pins the attempt count — it's the best test
in the diff, and it's the model the other five should follow.

Issue #10, opened 2025-07-23
Migrate to HTMX


fancysnake:
Move interactive panel and form flows from full-page POST/redirect cycles to HTMX partial swaps, so editing, list mutations and multi-step wizards update in place. The panel is form-heavy (CFP, proposals, venues, tracks, facilitators); full-page reloads on every mutation are the main UX cost CLAUDE.md calls out ("are we asking for needless clicks?").

## Where we are

- Bootstrap → Tailwind is done; styling is Tailwind 4 through Vite in `src/ludamus/client/`.
- HTMX is partially adopted: about 27 templates carry `hx-*` attributes, `confirm.ts` backs `data-confirm` on destructive actions everywhere.
- There is no inventory of which flows are HTMX-driven vs full reload, and no documented partial-swap convention. That is the missing artifact.

## This issue

1. Inventory the panel: list each mutating flow and mark HTMX vs full-page reload, starting with the highest-traffic list pages (proposals, venues, tracks).
2. Convert one flow end to end as the pattern reference, with the partial template + `hx-target`/`hx-swap` convention written down where the `product-design` skill and `hector` can point at it.
3. Before/after screenshots for the PR.

Converting the remaining flows follows the written convention as separate per-page issues; this issue is done when the convention exists and one flow proves it.

hasparus:
Can we close this? We're already using HTMX in a few places and it seems the usage will be growing on case by case basis.

PR #692, opened 2026-07-25
Lay the timetable calendar out from one shared track list

hasparus:
The sticky header and the scrolling body were two independent `w-max`
flex rows. Each sized itself to its own content, and the two contents are
not comparable: room names give the header real intrinsic width, while
the body's sessions are absolutely positioned and contribute none. The
header therefore ran wider than the columns it labels, by a gap that grew
with every day and room -- 392px at three days by five rooms, enough to
put a session under the wrong room name.

Both strips are now CSS grids over the same explicit tracks, one per
column across all days, declared once on the calendar as
`--total-columns`. `minmax(9rem, 1fr)` pins the lower bound so no cell's
content can widen a track, and each strip's width is computed from the
column count rather than derived from `max-content`. Days become subgrids
so a day's own cells inherit those tracks instead of re-splitting their
share -- which is also why day wrappers must stay free of padding and
border, noted in the CSS.

Adding a day, a room, or a longer room name now moves both strips
identically or not at all.

Co-authored-by: hasparus <hasparus@gmail.com>

## Summary by CodeRabbit

* **New Features**
  * Added resizable timetable columns using mouse or keyboard controls.
  * Column widths now persist between visits and can be reset with a double-click.
  * Added accessible labels and instructions for column resizing.
* **Improvements**
  * Improved timetable header and room-column alignment, including when content changes.
  * Enhanced layout stability across dynamic timetable updates.
* **Documentation**
  * Recorded troubleshooting notes for sandbox tooling, browser setup, and frontend asset updates.

### DIFF
--- a/PAPERCUTS.md
+++ b/PAPERCUTS.md
@@ -167,3 +167,11 @@ If you fix a papercut, remove it.
   Playwright still transpiles and executes the changed spec successfully.
 - 2026-07-24: playwriter 'session new' printed a fresh id but the relay 404'd it
   (Session 14 not found); had to reuse an old session id from session list
+- 2026-07-25: Web sandbox session started with an empty .venv (no pytest, no
+  django) and mise still wedged on pipx:shellcheck-py/hadolint-py, so no mise
+  task could bootstrap it; ran poetry install by hand and invoked pytest as
+  .venv/bin/python -m pytest with PYTHONPATH=src plus .env.test sourced
+  manually.
+- 2026-07-25: /opt/pw-browsers held chromium_headless_shell-1208 as an empty
+  directory, so every e2e test failed with "Executable doesn't exist" until
+  npx playwright install chromium refetched it.
--- a/PAPERCUTS.md
+++ b/PAPERCUTS.md
@@ -175,8 +175,11 @@ If you fix a papercut, remove it.
 - 2026-07-25: /opt/pw-browsers held chromium_headless_shell-1208 as an empty
   directory, so every e2e test failed with "Executable doesn't exist" until
   npx playwright install chromium refetched it.
-- 2026-07-25: After `aubr -C src/ludamus/client build`, a `runserver --noreload`
-  started earlier keeps serving the previous vite manifest, whose asset files
-  the rebuild has deleted - the page renders with no CSS at all and looks like
-  a layout bug rather than a stale server. Restart the server after every
-  client build.
+- 2026-07-25: Iterated on timetable CSS/TS against a `.env.e2e` server, where
+  `ENV="test"` turns django_vite `dev_mode` off, so each edit needed
+  `aubr build` plus a restart (`--noreload` caches the manifest, and the
+  rebuild deletes the hashed files it points at - the page then renders with
+  no CSS and reads as a layout bug). For an edit loop against the e2e seed,
+  export `ENV=development` and `VITE_PORT`, run the vite dev server, and keep
+  the rest of `.env.e2e`: assets come from vite with HMR, no build, no
+  restart. Build only before handing the page to Playwright.
--- a/PAPERCUTS.md
+++ b/PAPERCUTS.md
@@ -175,3 +175,8 @@ If you fix a papercut, remove it.
 - 2026-07-25: /opt/pw-browsers held chromium_headless_shell-1208 as an empty
   directory, so every e2e test failed with "Executable doesn't exist" until
   npx playwright install chromium refetched it.
+- 2026-07-25: After `aubr -C src/ludamus/client build`, a `runserver --noreload`
+  started earlier keeps serving the previous vite manifest, whose asset files
+  the rebuild has deleted - the page renders with no CSS at all and looks like
+  a layout bug rather than a stale server. Restart the server after every
+  client build.
--- a/src/ludamus/client/src/timetable.css
+++ b/src/ludamus/client/src/timetable.css
@@ -2,7 +2,11 @@
   .timetable-calendar {
     --minute-px: 1px;
     --timetable-time-column: 4rem;
-    --timetable-column-min: 9rem;
+    /* Every column is this wide, and the resize handles drag it (timetable.ts
+       writes --timetable-column-width on the document element, so the value
+       outlives each HTMX swap of the grid). The literal is only the default. */
+    --timetable-column-min: var(--timetable-column-width, 9rem);
+    --timetable-resizer-width: 0.875rem;
   }
 
   /* The sticky header and the scrolling body are separate elements, so they
@@ -70,6 +74,50 @@
 
   .timetable-room-cell {
     grid-row: 3;
+    position: relative;
+  }
+
+  /* One shared width means any handle drives the whole grid, so every room
+     cell carries one. It stays inside its cell rather than straddling the
+     border: an overhang would widen the calendar's scrollable area and put the
+     header back out of step with the body. */
+  .timetable-column-resizer {
+    position: absolute;
+    top: 0;
+    bottom: 0;
+    right: 0;
+    width: var(--timetable-resizer-width);
+    z-index: 1;
+    padding: 0;
+    border: 0;
+    background: transparent;
+    cursor: col-resize;
+    touch-action: none;
+  }
+
+  /* Focus is left to the global :focus-visible ring; this bar is the pointer
+     affordance -- it marks the border the grip acts on. */
+  .timetable-column-resizer::before {
+    content: "";
+    position: absolute;
+    inset-block: 0.375rem;
+    right: 0;
+    width: 2px;
+    border-radius: 1px;
+    background: transparent;
+    transition: background-color 120ms;
+  }
+
+  .timetable-column-resizer:hover::before,
+  .timetable-column-resizer.is-resizing::before {
+    background: var(--color-primary);
+  }
+
+  /* Pointer capture keeps the drag alive over the whole page, so the resize
+     cursor has to follow it there too. */
+  .timetable-resizing,
+  .timetable-resizing * {
+    cursor: col-resize !important;
   }
 
   .timetable-time-axis,
--- a/src/ludamus/client/src/timetable.css
+++ b/src/ludamus/client/src/timetable.css
@@ -1,17 +1,75 @@
 @layer components {
   .timetable-calendar {
     --minute-px: 1px;
+    --timetable-time-column: 4rem;
+    --timetable-column-min: 9rem;
   }
 
-  .timetable-day-grid,
+  /* The sticky header and the scrolling body are separate elements, so they
+     line up only if both are laid out from the same track list. Tracks are
+     therefore fully explicit and nothing inside a cell can resize them:
+     `minmax(min, 1fr)` pins the lower bound, and the strip width is computed
+     from the column count instead of being derived from content. Sizing to
+     content is what drifts -- room names widen the header while the body,
+     whose sessions are absolutely positioned, contributes nothing. */
+  .timetable-day-strip {
+    display: grid;
+    grid-template-columns: var(--timetable-time-column) repeat(
+        var(--total-columns),
+        minmax(var(--timetable-column-min), 1fr)
+      );
+    width: max(
+      100%,
+      calc(var(--timetable-time-column) + var(--total-columns) * var(--timetable-column-min))
+    );
+  }
+
+  /* Day name, room group, room name. A tier with no cells collapses to zero --
+     that is how the single-day view drops the date row. */
+  .timetable-day-strip-header {
+    grid-template-rows: auto auto auto;
+  }
+
+  /* A day re-uses the parent tracks rather than splitting its own width, so
+     adding days or rooms cannot shift the columns out of step. Subgrids must
+     stay free of padding, border and margin: those inset the subgrid and would
+     shove every track inside the day sideways. Chrome goes on the cells. */
+  .timetable-day-header,
+  .timetable-day-grid {
+    display: grid;
+    grid-template-columns: subgrid;
+    grid-column: span var(--column-count);
+  }
+
   .timetable-day-header {
-    flex: 1 0 calc(9rem * var(--column-count));
-    min-width: calc(9rem * var(--column-count));
+    grid-row: 1 / -1;
+    grid-template-rows: subgrid;
+  }
+
+  /* Pinned explicitly: the day headers span rows, and auto-placement resolves
+     row-spanning items before auto-placed ones, so an unpinned corner would be
+     pushed past the first day. */
+  .timetable-corner {
+    grid-row: 1 / -1;
+    grid-column: 1;
+  }
+
+  .timetable-time-axis {
+    grid-column: 1;
+  }
+
+  .timetable-day-title {
+    grid-row: 1;
+    grid-column: 1 / -1;
   }
 
   .timetable-header-cell {
-    flex: var(--columns) 1 0;
-    min-width: calc(9rem * var(--columns));
+    grid-row: 2;
+    grid-column: span var(--columns);
+  }
+
+  .timetable-room-cell {
+    grid-row: 3;
   }
 
   .timetable-time-axis,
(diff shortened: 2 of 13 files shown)


hasparus:

Scope: `49b67ca..691caeb` (grid unification, column resize, papercut).

**Verdict: request changes.**

The layout fix is the right call and I want to say so plainly: replacing two independently-sized `w-max` flex rows with one explicit track list and subgrids is the correct structural answer, not a patch over the symptom. The CSS comments in `timetable.css` explaining *why* the tracks must be explicit and why subgrids must stay padding-free are genuinely good — they encode the failure mode, not the mechanism.

The resize feature bolted on top of it is where the problems are, and one of them reaches back into the layout change itself.

---

### 1. `total_columns` invents a variability that does not exist, and pays for it twice in the templates

`src/ludamus/pacts/chronology.py:430` · `src/ludamus/templates/panel/parts/timetable-grid.html:42,47,97`

```python
return sum(len(day.columns) for day in self.days)
```

`_build_day_grid` (`src/ludamus/mills/timetable.py:186`) builds `columns` by iterating the single shared `spaces` list. Every day has exactly `len(grid.spaces)` columns — always. The template already leans on this: `grid.groups` is built once from `spaces` (`mills/timetable.py:105`) and re-rendered inside *every* day's header. If days could differ, that line would already be broken.

So the sum is a fiction, and the fiction has a cost. Because `total_columns` is presented as "whatever the days happen to add up to", each day has to restate its own count — `style="--column-count: {{ day.columns|length }}"` is stamped at line 47 *and* line 97, the same invariant value, in two places, on every day.

The code-judo move: put two numbers on `.timetable-calendar` once and let inheritance do the rest.

```html
style="--days: {{ grid.days|length }}; --columns-per-day: {{ grid.spaces|length }}"
```
```css
.timetable-day-strip {
  grid-template-columns:
    var(--timetable-time-column)
    repeat(calc(var(--days) * var(--columns-per-day)), minmax(var(--timetable-column-min), 1fr));
}
.timetable-day-header, .timetable-day-grid { grid-column: span var(--columns-per-day); }
```

Both per-day inline styles disappear, and so does the new DTO property. If you genuinely expect days to diverge later, then say so *by construction* — hoist columns onto the grid and let a day hold only its sessions — rather than with a sum that would silently render a misaligned grid the day it stops holding.

**Must fix**, or justify why the per-day count needs to stay addressable.

---

### 2. The test for `total_columns` passes for the wrong reason

`tests/integration/web/panel/test_timetable.py:223-227`

```python
expected_day_count = 2
...
assert grid.total_columns == expected_day_count
assert "--total-columns: 2" in content
```

The fixture has one space. Two days × one column = 2, which is also the day count — hence the variable name, which asserts the wrong concept. A `total_columns` implemented as `len(self.days)` passes this test unchanged. So does one implemented as `len(self.spaces)`. The second assertion re-hardcodes the same coincidence as a literal.

The property's whole point is that columns ≠ days. Test it with ≥2 spaces so the three candidate implementations actually diverge. **Must fix** — this is the only test of the new property and it currently proves nothing.

---

### 3. `setColumnWidth` persists to `localStorage` on every `pointermove`, and forces two layout flushes per frame

`src/ludamus/client/src/timetable.ts:462-467`, `440-443`, `479-486`

Every pointer move currently does:

- `localStorage.setItem` — synchronous, disk-backed, at pointer frequency;
- `roomCells()` → `querySelectorAll` + array spread, twice (once in `columnWidthFromPointer`, once via `syncResizers` → `renderedColumnWidth`);
- `getBoundingClientRect()` *immediately after* mutating `--timetable-column-width` on `documentElement` — a guaranteed style-recalc + layout flush, every frame, on a grid with up to 15 subgridded columns.

`setColumnWidth` is conflating three jobs: apply, persist, announce. Split it. `applyColumnWidth(px)` sets the custom property during the drag; persist and `syncResizers()` once in `onEnd`. Keyboard steps can keep calling all three since they are discrete.

While you are in there: the handle index and the first cell's left edge are constant for the duration of a drag but are recomputed on every move (`columnWidthFromPointer` re-queries the DOM and re-runs `indexOf`). Capture both at `pointerdown` and the move handler collapses to arithmetic. **Must fix** the per-move `setItem` at minimum.

---

### 4. The resizer's ARIA contract is split across two layers and is invalid until the bundle loads

`src/ludamus/templates/panel/parts/timetable-grid.html:69-81` · `src/ludamus/client/src/timetable.ts:452-460`

The template ships `role="separator"`, `tabindex="0"`, `aria-label`, `aria-orientation`. The script supplies `aria-valuemin`, `aria-valuemax`, `aria-valuenow`, `aria-valuetext`. Neither half is a complete widget.

Server-rendered HTML is therefore a focusable separator with no value — invalid ARIA on first paint, and if the bundle fails to load you are left with a keyboard-reachable control that announces nothing and does nothing. The comment at `timetable.ts:450` claims "One source of truth for the bounds", but the widget's *identity* lives in the template, so there are two.

Pick one owner. Either render the values server-side from the same constants, or (better — see #5) have the script create the exposed handle outright, which makes the whole element one artifact of one layer. **Must fix**; the DoD's "keyboard-reachable semantic HTML" is not met by a control whose semantics arrive asynchronously.

---

### 5. Fourteen duplicate `aria-hidden` divs and a template branch that exists only to choose which copy gets the a11y attributes

`src/ludamus/templates/panel/parts/timetable-grid.html:69-81`

```django
{% if forloop.first and forloop.parentloop.first %}
    <div class="timetable-column-resizer" role="separator" ... tabindex="0"></div>
{% else %}
    <div class="timetable-column-resizer" title="..." aria-hidden="true"></div>
{% endif %}
```

Copy-pasted markup, differing only in attributes, selected by a nested-forloop conditional. At 3 days × 5 rooms that is 15 nodes, 14 of them decorative. It also forces the `resizers()` selector in the script to filter on `[role="separator"]` — a DOM query standing in for "the one that's real".

The judo move: make the grip a pseudo-element and keep exactly one real handle.

```css
.timetable-room-cell::after { /* the grip; inherits the current hover/resizing styling */ }
```

Events on a pseudo-element hit-test to the generating element, so `pointerdown` still fires with `.timetable-room-cell` as the target — you just check the pointer landed within `--timetable-resizer-width` of the cell's right edge. That deletes the template branch, the duplicated markup, all 15 divs, and the `[role="separator"]` filter in one move. The single remaining handle is created by the script, which resolves #4 for free.

Separately: CLAUDE.md is explicit that hand-rolled components belong in `templates/components/` behind tessera, not inlined twice into a page part.

---

### 6. Drag sensitivity varies 15× depending on which grip you happen to grab

`src/ludamus/client/src/timetable.ts:479-486`

```ts
return (clientX - grabOffset - first.getBoundingClientRect().left) / (index + 1);
```

The leftmost grip gives 1px of column width per 1px of pointer travel. The fifteenth gives 1/15px — you drag an inch and the columns barely move. Same control, same affordance, wildly different response. The rightmost grip is also at the far edge of a horizontally scrolled container, where widening pushes content the drag is chasing.

The math is correct for the model; the model is the problem. "One shared width" and "a grip on every column border" pull against each other — the grips promise per-column control and deliver a global one, at fifteen different gear ratios. Either expose a single grip (which #5 gets you anyway) or reconsider whether per-column widths are the honest model here. The current middle is the worst of both.

Not a merge blocker on its own, but I would not ship the multi-grip affordance as-is.

---

### 7. Two sources of truth for the default width; `COLUMN_WIDTH_DEFAULT` is dead weight

`src/ludamus/client/src/timetable.css:8` (`9rem`) · `src/ludamus/client/src/timetable.ts:421` (`144`)

These agree only at a 16px root font size, and nothing enforces that. Worse, `COLUMN_WIDTH_DEFAULT` is barely reachable: `renderedColumnWidth()` falls through to it only when there are *no* room cells at all (nothing to size anyway), and line 542 already guards on `getItem(...) !== null` before calling it.

```ts
function storedColumnWidth(): number | null   // was: number, defaulting to 144
```

Delete `COLUMN_WIDTH_DEFAULT` and let the CSS own the default — it already does. Line 542 then becomes `const stored = storedColumnWidth(); if (stored !== null) setColumnWidth(stored);` instead of reading `localStorage` twice and encoding "is it set?" separately from "what is it?". Optional but cheap, and it removes a unit-mismatch trap.

---

### 8. Layout implementation detail leaked into `pacts`

`src/ludamus/pacts/chronology.py:430-435`

The comment on the new property explains CSS track lists and sticky-header alignment — inside the bottom-layer shared contract module. `timetable.css:12-18` already explains the same thing, better, where a reader can act on it. The next consumer of `TimetableGridDTO` (print view, export) inherits a property whose name only makes sense to one stylesheet.

If the property survives #1, keep the domain rationale here and the CSS rationale in the CSS.

---

### 9. The body lost its per-day separator

`src/ludamus/templates/panel/parts/timetable-grid.html:96`

`{% if not forloop.first %}border-l border-border{% endif %}` was dropped in favour of `bg-bg-secondary`. The header still marks day boundaries (`border-l` on `.timetable-day-title` and `.timetable-header-cell`), but the scrolling body no longer does — so once you scroll past the sticky header in the all-days view, adjacent days are visually indistinguishable from adjacent rooms.

Intentional? CLAUDE.md asks for screenshots of affected pages in the PR description, and this is exactly the kind of thing they would settle.

---

### 10. Smaller things

- `timetable.ts:509-512` — `pointerup` and `pointercancel` are both registered `{ once: true }`. Only one fires; the other stays attached to that handle for the life of the page, accumulating one stale listener per drag. Harmless today because `onEnd` is idempotent, but it is a leak with no upside.
- `timetable.ts:522-525` — the `steps` record is reallocated on every `keydown` anywhere in the document, including the ones that return immediately. Hoist it to module scope.
- `timetable.ts:409` and `:518` — two document-level `keydown` listeners in one file now, neither aware of the other.
- `timetable.css:87-88` — `padding: 0; border: 0;` are `<button>` resets applied to a `<div>`. Leftovers from an earlier shape of the control; they do nothing.
- `timetable.css:71,76` — `grid-row: 2` / `grid-row: 3` are hardcoded against a template that conditionally omits row 1. The collapse-to-zero trick is documented and works, but the coupling is invisible from the template side.
- `localStorage.setItem` at `timetable.ts:465` throws in Safari private mode / blocked-storage contexts. Mid-drag that throws out of the move handler each frame. `pointerup` still cleans up so it degrades rather than wedges, but per the DoD this is a failure path with no useful error.

---

### 11. Test coverage gaps

`tests/e2e/tests/timetable.spec.ts`

- `cell.textContent = "Room name long enough to stretch a content-sized column"` wipes the cell's children — including the *only* `[role="separator"]` handle on the page. The alignment assertions that follow still pass, so the test is green, but it leaves the DOM in a state that would silently invalidate any handle assertion added to that test later. Use `insertAdjacentText` or append a text node.
- The width test hardcodes `start + 40` and `start + 56` against the seed's column count and the `/(index + 1)` divisor. A change to `bootstrap_timetable.py` breaks it with an arithmetic mismatch rather than a legible failure. Derive the expected delta from the observed column count.
- Nothing covers the persisted width surviving an **HTMX grid swap**, which is the specific scenario the "store it on `documentElement`, not the calendar" design exists for (`timetable.ts:540`). The test reloads the page — that exercises `localStorage`, not the swap path. A track/date filter change would.

---

### Summary

The grid unification is sound and I would take it as-is modulo #1 and #2. The resize feature needs another pass: it duplicates markup, splits a widget's semantics across two layers, writes to disk on every pointer frame, and ships an affordance whose response varies by an order of magnitude across nominally identical controls. #5 is the move that collapses most of the rest.


---

hasparus:
Worked through the review. Nine of ten addressed in d5271d4; one I'm pushing back on.

## Fixed

**#1 — fictional per-day variability.** Correct, and the tell was exactly as described: `groups` is one list reused for every day, so summing per day described a state that cannot occur. `total_columns` is now the page repeated once per day, and saying so let a single `--columns-per-day` on the calendar replace the `--column-count` stamped onto each day header *and* each day grid. Two inline styles gone.

**#2 — test passing for the wrong reason.** It asserted `2` against one room and two days, which `len(days)` alone satisfied. Now three rooms over two days, so only the product passes, and the two custom properties are asserted directly.

**#3 — per-frame `localStorage` + forced layout.** Split into `applyColumnWidth` (paint, per frame, touches nothing else), `commitColumnWidth` (paint + persist + announce, on drag end and on each discrete keyboard step) and `announceColumnWidth`. The drag now measures once at `pointerdown` — the first column's left edge cannot move, since the time column ahead of it is a fixed track — and reads no layout after that.

**#4 + #5 — the ARIA contract and the fifteen divs.** Taken together, as suggested. Grips are now `.timetable-room-cell::after`, which is hit-testable, so all fifteen borders stay draggable with zero nodes; `pointerdown` recognises a grab by proximity to the cell's right edge. One real handle remains, on the first border, carrying focus and the ARIA contract. It ships as a *static* separator and the script adds `tabindex` plus the value range — so a missing bundle leaves no dead tab stop rather than an unlabelled one, which is the sharper half of #4. The `[role="separator"]` filter is gone with the duplicates.

**#7 — two sources of truth for the default.** `storedColumnWidth()` returns `number | null` and `COLUMN_WIDTH_DEFAULT` is deleted; the default is the CSS fallback alone. Same treatment for the grip width — it's measured off the real handle rather than restated in pixels. (That one bit immediately: my first attempt read the custom property and `parseFloat("0.875rem")` gave `0.875`, so the hit zone was one pixel wide.)

**#8 — CSS rationale in the bottom-layer DTO.** Moved with the value: `total_columns` is now computed in `mills/timetable.py` beside `total_minutes`, `total_pages` and `total_spaces`, which are all derived-and-stored there already.

**#9 — day boundary lost in the body.** Real regression, not intentional. Restored on the day's first column, since a border on the day itself would inset its subgrid tracks — noted in the CSS so it doesn't get "fixed" back onto the wrapper.

**#10 — misc.** Listener leak fixed (`pointerup`/`pointercancel` are removed explicitly in `onEnd`), the step map is hoisted to a module constant, the `padding: 0; border: 0` button-resets are gone from what is a div.

**#11 — tests.** The alignment test writes into the room name instead of the cell, leaving the handle alone. **The HTMX-swap gap is now covered** — that was the best catch in the list, since surviving a swap is the entire reason the width lives on `documentElement`; the test dispatches `timetableChanged`, waits for the grid response, and asserts both the width and the re-promoted handle. The `+40` is now derived from the grabbed border's index rather than hardcoded.

## Pushing back

**#6 — drag sensitivity varying by handle index.** I don't think this is a defect. Every grip moves *the border you grabbed* to exactly where your cursor is; that the leftmost border needs less travel is the consequence of that, not a departure from it. Grabbing the last border and having it lag your cursor by 15× would be the surprising behaviour. The divisor is what keeps the border under the cursor rather than running away from it, which is the property the drag is built around — and it's what the e2e test pins.

## Not done

The tessera-component point under #5 stands; the grip is a two-rule pseudo-element and a single handle rather than something the catalog would carry, so I've left it local.

Also worth recording: `main` being ~80 commits behind was my fault, not the branch's — my local `main` ref was stale, which is also why my earlier `tingle check` reported "257 paid off" when CI correctly said `+7`. Against the real base this branch is now `-5`: the `TIMETABLE_*` constants move to `specs/timetable.py`, where pure constants read only by mills belong, and that pays for the field added to `pacts/chronology.py`.

---

hasparus:
On the pre-merge check about accessible selectors — half of it is right, so the handle is now queried by role in bc7169c:

```ts
const handle = page.getByRole("separator", { name: "Column width" });
```

The `.timetable-*` locators in the rest of these tests stay, because they aren't standing in for a user-facing query — they *are* the subject. The assertions are geometric: every column's left and right edge, header against body, before and after a resize. There is no accessible query for "the fourth column's bounding box", and the whole point of these two tests is that the boxes line up, which is a fact about elements rather than about roles. A role-based locator would either be unable to express it or would resolve to the same nodes less legibly.

The drag targets a room header's right edge for the same reason: the grip is a `::after` pseudo-element, so there is no node to address at all — the test grabs the border by coordinate, which is exactly what a user does.

---

PR #693, opened 2026-07-26
Name the timetable columns, and test them by that name

hasparus:
Follow-up to #692, which shipped the timetable's shared track list and column resizing. A pre-merge check there flagged the new E2E tests for locating things by `.timetable-*` class instead of by role, label or text.

I initially pushed back, arguing the class was the subject rather than a stand-in for a user-facing query. That was half an excuse. The reason no user-facing query existed for a timetable column is that **the column had no accessible identity**: a session block announced its title and nothing about which room it sat in. The selector complaint was pointing at a real gap.

## The change

One attribute pair on the body column:

```html
<div class="timetable-column" role="group" aria-label="{{ col.space.name }}">
```

A screen-reader user moving through the grid now gets the room from the column and the day from the `<section aria-label>` that already wraps it. And the query exists, so the tests can use it.

Both new tests now locate everything by role, name or text:

| before | after |
| --- | --- |
| `.timetable-room-cell` / `.timetable-column` edges | `getByText(room)` vs `getByRole("group", { name: room })` |
| `.timetable-calendar` visible after reload | the room name is visible |
| `.timetable-room-cell` right edge for the drag | the named column's border, at the height of the room name |

The assertion changed with the locators, and I think for the better. It used to compare header-cell edges against body-column edges — true, but a statement about DOM boxes. It now asks the question a person reading the schedule asks: **is this room's name sitting over this room's column?** — comparing centres.

## Verifying the weaker assertion still bites

Going from edge-equality to centre-equality is the risk in this change, so I checked it rather than assuming. Mutating the CSS back to content-sized tracks (`minmax(min-content, 1fr)` plus `width: max-content`) and re-running:

```
Expected: <= 1
Received:    12.734375
```

It fails on the seeded **two-room** event — the original bug needed three days and five rooms before it was visible by eye. CSS restored afterwards; the only production change in this PR is the two ARIA attributes.

## Checks

- 23/23 timetable E2E (chromium)
- 44 timetable integration tests
- `tingle check` clean

No screenshots: nothing about the rendering changes, only what assistive tech is told about it.

---
_Generated by [Claude Code](https://claude.ai/code/session_01WcmdYnQDahkTsdsiUq2WGV)_

## Summary by CodeRabbit

* **Accessibility**
  * Updated the timetable’s room columns to be exposed as labeled groups to assistive technologies.

* **Bug Fixes**
  * Fixed timetable room-header alignment, including when room names are much longer.
  * Improved timetable column resizing behavior and consistency.
  * Increased reliability when reloading and continuing resized timetable interactions.

* **Tests**
  * Strengthened end-to-end coverage for alignment and resizing using measured layout positions.

### DIFF
--- a/src/ludamus/templates/panel/parts/timetable-day-grid.html
+++ b/src/ludamus/templates/panel/parts/timetable-day-grid.html
@@ -2,7 +2,10 @@
 {% with date_value=date_selection|date:"Y-m-d"|default:date_selection %}
     {% load tessera %}
     {% for col in day.columns %}
+        {# Names the column after its room: a session block says what it is, and this says where. #}
         <div class="timetable-column relative border-l border-border"
+             role="group"
+             aria-label="{{ col.space.name }}"
              data-space-pk="{{ col.space.pk }}">
             {% for label in grid.time_labels %}
                 <div class="timetable-slot-line absolute left-0 right-0 border-t border-border"
--- a/src/ludamus/templates/panel/parts/timetable-day-grid.html
+++ b/src/ludamus/templates/panel/parts/timetable-day-grid.html
@@ -2,7 +2,7 @@
 {% with date_value=date_selection|date:"Y-m-d"|default:date_selection %}
     {% load tessera %}
     {% for col in day.columns %}
-        {# Names the column after its room: a session block says what it is, and this says where. #}
+        {# A session block names itself but not its room; the column supplies that. #}
         <div class="timetable-column relative border-l border-border"
              role="group"
              aria-label="{{ col.space.name }}"
--- a/tests/e2e/tests/timetable.spec.ts
+++ b/tests/e2e/tests/timetable.spec.ts
@@ -1,4 +1,4 @@
-import { expect, test } from "@playwright/test";
+import { expect, type Page, test } from "@playwright/test";
 
 test.describe.configure({ mode: "serial" });
 
@@ -149,78 +149,80 @@ test.describe("Timetable", () => {
     });
   });
 
-  test("room headers stay aligned with their columns, whatever they contain", async ({ page }) => {
+  // The seeded rooms, in the order the timetable lays them out.
+  const ROOMS = ["Garden Table", "Willow Table"];
+
+  // What a reader of the schedule actually checks: is this room's name sitting
+  // over this room's column? Centres rather than edges, because the name is what
+  // is visible -- a column that drifts from its label drifts by whole columns.
+  const expectLabelledColumns = async (page: Page, rooms: string[]) => {
+    for (const room of rooms) {
+      const labels = page.getByText(room, { exact: true });
+      const columns = page.getByRole("group", { name: room });
+      const dayCount = await columns.count();
+      expect(dayCount).toBeGreaterThan(0);
+      await expect(labels).toHaveCount(dayCount);
+
+      for (let day = 0; day < dayCount; day++) {
+        const label = (await labels.nth(day).boundingBox())!;
+        const column = (await columns.nth(day).boundingBox())!;
+        const centre = (box: { width: number; x: number }) => box.x + box.width / 2;
+        expect(Math.abs(centre(label) - centre(column))).toBeLessThanOrEqual(1);
+      }
+    }
+  };
+
+  test("room names stay over their own columns, whatever they contain", async ({ page }) => {
     await page.setViewportSize({ width: 480, height: 800 });
     await page.goto("/panel/event/sunhaven-festival/timetable/?date=all");
 
-    const columnEdges = () =>
-      page.locator(".timetable-calendar").evaluate((calendar) => {
-        const edges = (selector: string) =>
-          Array.from(calendar.querySelectorAll(selector), (cell) => {
-            const box = cell.getBoundingClientRect();
-            return { left: box.left, right: box.right };
-          });
-        return { body: edges(".timetable-column"), header: edges(".timetable-room-cell") };
-      });
-
-    const expectAligned = async () => {
-      const { body, header } = await columnEdges();
-      expect(header.length).toBeGreaterThan(1);
-      expect(body).toHaveLength(header.length);
-      for (const [index, cell] of header.entries()) {
-        expect(Math.abs(cell.left - body[index].left)).toBeLessThanOrEqual(1);
-        expect(Math.abs(cell.right - body[index].right)).toBeLessThanOrEqual(1);
-      }
-    };
-
-    await expectAligned();
+    await expectLabelledColumns(page, ROOMS);
 
     // Column widths must come from the track list, not from what a cell holds:
-    // sessions are absolutely positioned and contribute no width, so any
-    // content-driven header would drift away from the body it labels. Written
-    // into the room name, leaving the resize handle that shares the cell alone.
+    // sessions are absolutely positioned and contribute no width, so a
+    // content-driven header would widen the first room and shove every room
+    // after it off the column it labels.
     await page
-      .locator(".timetable-room-cell div")
-      .last()
+      .getByText(ROOMS[0], { exact: true })
+      .first()
       .evaluate((name) => {
         name.textContent = "Room name long enough to stretch a content-sized column";
       });
-    await expectAligned();
+    await expectLabelledColumns(page, ROOMS.slice(1));
   });
 
   test("dragging one column edge resizes every column, and the width sticks", async ({ page }) => {
     await page.setViewportSize({ width: 1280, height: 800 });
     await page.goto("/panel/event/sunhaven-festival/timetable/?date=all");
 
-    // One entry means every column agrees; drift is header vs the body it labels.
-    const geometry = () =>
-      page.locator(".timetable-calendar").evaluate((calendar) => {
-        const boxes = (selector: string) =>
-          Array.from(calendar.querySelectorAll(selector), (cell) => cell.getBoundingClientRect());
-        const header = boxes(".timetable-room-cell");
-        const body = boxes(".timetable-column");
-        return {
-          drift: Math.max(...body.map((box, index) => Math.abs(box.left - header[index].left))),
-          widths: [...new Set(header.map((box) => Math.round(box.width)))],
-        };
-      });
+    const columns = () => page.getByRole("group", { name: new RegExp(ROOMS.join("|")) });
+    const columnWidths = async () => {
+      const boxes = await columns().evaluateAll((found) =>
+        found.map((column) => Math.round(column.getBoundingClientRect().width)),
+      );
+      return [...new Set(boxes)];
+    };
     const expectWidth = async (expected: number) => {
-      const { drift, widths } = await geometry();
+      const widths = await columnWidths();
+      // One entry means every column agrees on the shared width.
       expect(widths).toHaveLength(1);
       expect(Math.abs(widths[0] - expected)).toBeLessThanOrEqual(1);
-      expect(drift).toBeLessThanOrEqual(1);
+      await expectLabelledColumns(page, ROOMS);
     };
 
-    const start = (await geometry()).widths[0];
+    const [start] = await columnWidths();
 
-    // Grabbing the border after the second column, so the two columns before it
-    // split the travel -- the border itself tracks the cursor either way.
+    // Grab the border on the far side of the second room, so the two rooms
+    // before it split the travel -- the border tracks the cursor either way.
     const grabbed = 2;
-    const border = (await page
-      .locator(".timetable-room-cell")
+    const border = (await columns()
       .nth(grabbed - 1)
       .boundingBox())!;
-    const [edgeX, edgeY] = [border.x + border.width - 2, border.y + border.height / 2];
+    const name = (await page
+      .getByText(ROOMS[grabbed - 1], { exact: true })
+      .first()
+      .boundingBox())!;
+    const [edgeX, edgeY] = [border.x + border.width - 2, name.y + name.height / 2];
     const travel = 80;
     await page.mouse.move(edgeX, edgeY);
     await page.mouse.down();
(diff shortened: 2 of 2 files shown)


hasparus:
## Thermo-nuclear code quality review — `claude/agenda-column-sync-ruvr85` (541669b)

Diffed `main...HEAD` with local `main` verified equal to `origin/main` (`297f90b`). Two files, +59/−58.

**Verdict: REQUEST CHANGES.** The direction is right and the headline judgement call is defensible (see "On the assertion trade" below — I do not think the weaker geometry is a problem). But the drag test now drives an interaction from the geometry of the wrong element, and the diff leaves a clear code-judo move on the table: two of the three new concepts (`ROOMS`, the `RegExp`, the index-zip between labels and columns) can be deleted outright rather than maintained.

---

### 1. The drag test measures the body column to hit a grip that lives in the header — `tests/e2e/tests/timetable.spec.ts:218`

```ts
const border = (await columns().nth(grabbed - 1).boundingBox())!;
...
const [edgeX, edgeY] = [border.x + border.width - 2, name.y + name.height / 2];
```

`columns()` is `getByRole("group")` — the **body** column from `timetable-day-grid.html`. The drag grip is the `::after` on `.timetable-room-cell` (`src/ludamus/templates/panel/parts/timetable-grid.html:59`), in the **header**. Before this PR the grab point came from the header cell that actually owns the grip; now `edgeX` comes from the body and `edgeY` from the header label, and the two only meet on the grip because header and body are column-aligned — which is precisely the invariant the sibling test exists to prove.

That is circular test design. If the track lists desync again (the exact #692 regression), the alignment test fails *and* the drag test fails by missing the grip and reporting a width mismatch, so the diagnostic points at resizing rather than at alignment. Two failures, one cause, neither naming it. Same construction at `:259`/`:262` for the double-click reset.

The variable is also still named `border` while holding a column box, and the coordinate pair is now sourced from two different elements — a reader cannot check that line without re-deriving the layout.

Fixes, cheapest first:

- Take both coordinates from the header cell that owns the grip (e.g. `getByText(room, { exact: true }).first().locator("xpath=..")`). Role-first locators are a good rule, but a hit-testable pseudo-element has no accessible representation — resolving it structurally is honest, not a lint rule to route around.
- Or make the grip real: render `role="separator"` on every room border with `tabindex="-1"` on the non-first ones, so `getByRole("separator").nth(grabbed - 1)` locates it directly. `timetable-grid.html:64-68` deliberately traded nodes for a pseudo-element, so this is a real cost — but it would delete the geometry hack and make the drag target locatable instead of computed.

**Blocking.** It passes today; it is wrong about what it is measuring.

### 2. The seed names are encoded three times, and all three copies can be deleted — `tests/e2e/tests/timetable.spec.ts:152-153`, `:198`

`ROOMS` duplicates `tests/e2e/scripts/bootstrap_timetable.py:95,101` with nothing linking them; the comment adds a load-bearing ordering claim ("in the order the timetable lays them out") that nothing enforces; and `:198` re-encodes the same list a third time as `new RegExp(ROOMS.join("|"))` — an unanchored regex built from unescaped strings, fine for `Garden Table` and a latent trap for any room name carrying a regex metacharacter.

Every column already carries its name in the DOM, courtesy of this very PR. Read it back instead:

```ts
const roomNames = async (page: Page) =>
  [...new Set(await page.getByRole("group").evaluateAll(
    (columns) => columns.map((column) => column.getAttribute("aria-label")!)))];
```

That deletes `ROOMS`, deletes the fixture coupling, deletes the regex (`getByRole("group")` alone already selects exactly these columns — nothing else on the page uses the role), deletes the ordering assumption, and makes both tests correct for any seed. `ROOMS.slice(1)` at `:191` becomes a `const [first, ...rest]` destructure.

Related: `:191` passes `ROOMS.slice(1)` into a helper whose only non-emptiness guard sits *inside* the per-room loop (`:163`). Hand it an empty array and it asserts nothing and passes green. Add `expect(rooms.length).toBeGreaterThan(0)` at the top of `expectLabelledColumns` — the old code guarded this with `expect(header.length).toBeGreaterThan(1)`, and that guard was dropped.

### 3. Label↔column pairing is an index-zip of two page-global locators — `tests/e2e/tests/timetable.spec.ts:160-168`

`labels.nth(day)` and `columns.nth(day)` are separate queries married by document order. Nothing in the DOM says label *i* labels column *i*; it holds by coincidence of template ordering. And `getByText(room, { exact: true })` is unscoped to the calendar, where the old code did everything inside `.timetable-calendar`. It passes today (`timetable-session-detail.html:50` renders `Room: X`, not an exact match, and no other node on this page emits a bare room name) — but `toHaveCount(dayCount)` will fail the day someone adds a room filter chip or a `<select>` of spaces, for a reason that has nothing to do with layout.

One change fixes both, and is the judo move I would most like to see here — `aria-labelledby` instead of `aria-label`:

```html
<div class="text-sm font-medium text-white wrap-anywhere"
     id="room-{{ day.date|date:'Y-m-d' }}-{{ col.space.pk }}">{{ col.space.name }}</div>
```

```html
<div class="timetable-column ..." role="group"
     aria-labelledby="room-{{ day.date|date:'Y-m-d' }}-{{ col.space.pk }}">
```

`getByRole("group", { name: room })` keeps working — name computation follows the reference. But the header/body pairing becomes a DOM fact rather than an ordering coincidence, the test can read the label box *from the column* in one round-trip instead of zipping indices, and the markup stops carrying two independently-editable copies of the same room name. Right now `aria-label` in `timetable-day-grid.html:8` and the visible text in `timetable-grid.html:60` can silently drift apart, and the alignment test is the only thing that would notice.

### 4. `expectLabelledColumns` trades one round-trip for ~40, and reports failures without saying which room or which day — `tests/e2e/tests/timetable.spec.ts:158-173`

The old `columnEdges()`/`geometry()` did one in-page `evaluate` returning everything. The new helper does `count()` + `toHaveCount()` + two `boundingBox()` calls per room per day, and `expectWidth` (`:210`) invokes it on every one of its five width checks. On a serial e2e suite that is a real cost for no gain.

Worse is the diagnostic: `:170` fails as `expected 12.7 to be <= 1`. No room, no day, no side. The 12.7px mutation figure quoted in the PR description is exactly what a maintainer would see, with nothing attached to it. Either collect `{ room, day, drift }[]` in one `evaluateAll` and assert on the array, or at minimum use the message argument: ``expect(drift, `${room}, day ${day}`)``.

While there: the alignment test no longer asserts the columns share a width — `columnWidths()` exists in the file but is only used by the resize test. Adding it to `expectLabelledColumns` restores the part of the old edge assertion that centres genuinely do not cover, for one extra line.

### 5. `role="group"` is defensible, but do not bill it as an accessibility fix — `src/ludamus/templates/panel/parts/timetable-day-grid.html:5-8`

It is not ARIA-invented-for-tests: a screen-reader user tabbing into a session block now gets the room as group context, which is real information they did not have, and the name is a data value so no i18n is owed. Keep it.

But it is a partial ARIA story, and the PR body should say so rather than claim the timetable got more accessible. The visible header cell that labels the column is still not associated with it (finding 3 fixes that); the group wraps decorative `.timetable-slot-line` divs alongside the sessions; and the session blocks inside remain `tabindex="0"` divs with no role at all — that is the accessibility problem in this component, and it is untouched. A `group` around them does not address it.

The comment at `:5` — "a session block says what it is, and this says where" — is doing more literary work than CLAUDE.md's comment budget wants. "Screen readers announce the room on entering the column" says the same thing and will still be true after the next refactor.

### 6. Minor — `tests/e2e/tests/timetable.spec.ts:235`, `:262`

- `:262` swaps `locator.dblclick({ position })` for `page.mouse.dblclick(x, y)`, dropping Playwright's actionability and stability checks for raw coordinates. It also reuses `edgeY` computed at `:225`, ~35 lines and one `page.reload()` earlier. That survives because the header row's y is stable across the reload — an unstated invariant now holding up an assertion.
- `:235` uses a visible room name as the post-reload readiness gate, where the old code waited on `.timetable-calendar`. A painted room name does not imply `timetable.ts` has re-promoted the handle. The following assertions retry, so this is a flake-surface note, not a defect.

---

### On the assertion trade (the thing flagged for hardest scrutiny)

**Sound. I would not block on it.** The reasoning matters, so: `.timetable-room-cell` is `text-center` and the name is a full-width block child (`timetable-grid.html:59-60`), so the label's box spans the cell's inner width and `centre(label) == centre(cell)` exactly, independent of text length. This is still a header-vs-body assertion — it dropped width equality, not the alignment claim.

What centres actually miss is a *symmetric* per-column width change, which needs a middle column to hide in. The seed is 2 rooms × 2 days (`timetable.spec.ts:83` asserts 2 day grids), so there is no middle column, and any uniform header/body width divergence still moves every centre. The 12.7px mutation check confirms the real regression is caught. Restoring the shared-width check per finding 4 closes the remaining gap for one line.

Worth crediting explicitly: the old mutation targeted `.timetable-room-cell div` `.last()`, which with `capacity: 8` seeded was the **"8 cap." div, not the room name** — the old comment claiming it was "written into the room name" was already wrong. `:186` targets the name by its text and genuinely tests what the comment says. That is a correctness fix hiding inside a locator refactor.

---

hasparus:
Addressed in 452f2be.

**#1 (blocking) — the drag measured the body to hit a grip in the header.** Fair, and the circularity is real. The test now asserts alignment *before* computing the grab point, so the grip's position is established rather than assumed; a desync fails there, naming itself, instead of surfacing downstream as an unexplained width mismatch. `border` is renamed and the grab point is derived once in a `gripAt()` helper that both the drag and the reset re-measure through.

**#2 — seed names encoded three times.** Gone. Room names are read off the page (`aria-label` on the columns), so `bootstrap_timetable.py` is the single source; the ordering claim and the unanchored regex are both deleted. The empty-array hole is closed too — renaming the first room now asserts there is a second one left to check, since `ROOMS.slice(1)` on a one-room fixture would otherwise have passed green.

One correction to the finding: **plain `getByRole("group")` does *not* select exactly the columns.** On this page it also matches the conflicts `<details>` and the theme-switcher `<fieldset>` — six groups, not four. My first attempt at the fix failed on it. Only the *named* groups are columns, and every subsequent lookup asks for a specific room, which those two can never answer to.

**#4 — round-trips and context-free failures.** Failures now name the room and day. Re-running the mutation check against content-sized tracks:

```
Error: Garden Table, day 1: the room name is not over its column
Expected: <= 1
Received:    12.734375
```

And the alignment helper asserts the shared width it had dropped, which closes the residual gap you identified in the centres-vs-edges trade.

**#5 — the comment.** Trimmed to one line.

**#3 (`aria-labelledby` to make the label↔column pairing a DOM fact) — not done.** It is the right idea and I'd take it, but it needs a per-day unique id on each header name div, and this PR is already a follow-up to a follow-up. Filed as the natural next step rather than grown into here. The pairing is currently guarded by the count assertion plus `exact: true`, which the `Room: X` detail-pane string does not satisfy.

**#6** — the reset re-measures through the same `gripAt()` helper, so no stale coordinate.

Thanks for the catch on the old mutation targeting `.last()` — that was the `8 cap.` div, not the room name. Genuinely wrong before, and I had not noticed.

Full timetable E2E: 23/23. `tingle check`: no new debt.

---

hasparus:
## Re-review — 452f2be

**Verdict: APPROVE.** The blocking finding is genuinely resolved, and the fixes did not introduce anything new that blocks merge.

### #1 (blocking) — resolved

`gripAt()` still derives `x` from the body column and `y` from the header label, but that is no longer circular, because the invariant it depends on is now *asserted* rather than assumed: `expectAlignedColumns` runs immediately before the grab, and again inside every `expectWidth`, so the reset's second `gripAt()` is also covered. A header/body desync now fails at the alignment assertion, naming the room and the day, before any drag runs. That is the correct shape of the fix — establish the precondition, then derive from it — and it is strictly better than what I proposed, since the grip's location stays a measured fact instead of a structural guess.

Centralising both grab points in one `gripAt()` also closes #6: the reset re-measures rather than reusing a coordinate computed 35 lines and one reload earlier.

### Your correction on `getByRole("group")` — you are right, I was wrong

Checked rather than taken on trust. `<details>` and `<fieldset>` both carry implicit role `group`, and both are on this page: `panel/timetable.html:103` (`#conflicts-fold`) and `components/theme_switcher.html:2` via `base.html:143`. So plain `getByRole("group")` matches six, not four. My finding #2 asserted the opposite and was simply incorrect.

The new code handles it correctly for a reason worth stating: neither of those two carries an `aria-label` *attribute* — the fieldset is named by `<legend>`, the details by `<summary>` — so `getAttribute("aria-label")` returns `null` for both and `.filter(Boolean)` drops them. It filters on the attribute, not the computed accessible name, which is what makes it exact. And `getByRole("group", { name: room })` matches accessible name as a substring by default, so the two intruders would need "Garden Table" inside "All clear" or "Select a display theme:" to collide. Over-matching would fail loudly at `toHaveCount(dayCount)` anyway rather than passing green.

### #3 deferral — acceptable, not blocking

Reassessing it against the code that actually landed rather than against my own description of it: both sides of the pairing are now filtered to the same room (`getByText(room, { exact: true })` and `getByRole("group", { name: room })`), so `nth(day)` zips *days within one room*, not two page-global lists married by document order. Sibling days rendered by one loop are a far weaker assumption than the cross-room ordering I was objecting to, and `toHaveCount(dayCount)` guards it. `aria-labelledby` is still the better markup — the room name is still stored twice and can still drift — but it is a template change with an id-uniqueness question attached, and it is not what this PR is about. Filing it is the right call.

#2, #4, #5 all check out: `ROOMS`, the fixture coupling and the unanchored regex are gone; the empty-array hole is closed at both ends; failures name room and day; the shared-width check is back.

### Non-blocking, for whenever you next touch this file

- `expectWidth` asserts shared width itself (`rest` is empty) *and* calls `expectAlignedColumns`, which asserts it again — one of the two is redundant.
- Round-trips went up, not down, versus what I flagged in #4: `expectAlignedColumns` is now four-plus `boundingBox` calls per invocation and `expectWidth` runs it five times. You traded that for per-room/per-day diagnostics, which is the right trade — noting it only so the cost is a choice rather than a surprise.
- `return names as string[]` — `.filter(Boolean)` doesn't narrow in TS. A `(n): n is string` predicate drops the cast. `tsc` on the suite reports no new errors from this diff; the four it does report are all pre-existing, including one in this file at `:47`, outside the diff.

---

PR #697, opened 2026-07-27
feat(seo): replace "Event webapp" meta description

hasparus:
Every page shipped `<meta name="description" content="Event webapp">`, and every
link preview used a hardcoded `https://zagrajmy.net/static/logo.png` regardless
of which sphere the link pointed at.

## What the competition actually does

I pulled the real tags off the competition before writing anything:

| Site | Homepage / listing | Object pages |
|---|---|---|
| Meetup | one static sentence — *"Find Meetup events, join groups, or start your own…"* | per-group / per-event |
| Eventbrite | static per listing type | per-event: title + date + venue |
| Luma, Partiful, Warhorn | one static sentence site-wide | — |

Google rewrites meta descriptions on ~70% of SERPs and has never used them for
ranking, so hand-writing one per page buys nothing there. The payoff is link
previews — Messenger, Discord, Slack. So: one default on listing pages, a
generated description on object pages. No per-sphere DB field.

## Descriptions

The default splits by audience, because one string was serving two readers.

**`zagrajmy.net`** — an organizer sizing up the tool:

> Konwent bez chaosu w arkuszach. Zgłoszenia programu, harmonogram sal i bloków
> tematycznych, zapisy z limitem miejsc i listą rezerwową.

**A sphere's own subdomain** — someone who already picked the convention, so it
describes rather than pitches:

> Program i zapisy na Pyrkon.

**Object pages** build from their own data. Event pages use the event
description, falling back to the default when it is blank. Encounter pages get a
description **they never had** — they set only `og:`/`twitter:`, so search
results and any crawler without OG support saw "Event webapp".

## Also

- One value feeds `description`, `og:description` and `twitter:description`.
  Every template that overrides one overrides all three.
- Link-preview image: a `meta_image_url` tag over `request.build_absolute_uri`,
  replacing a hardcoded absolute URL and a partial that hand-rolled the same
  thing. Event covers and encounter header images still win over the brand logo.
  Deliberately *not* the sphere logo: those may be SVG, which every preview
  crawler rejects.
- Titles gain a brand tail: `{page} • {sphere} • Zagrajmy`, skipped on the root
  sphere. Panel pages no longer render the sphere name twice.
- Dropped `<meta name="keywords">`. Ignored since 2009, and ours listed
  "seminars, professional development, tech events, design workshops".

## Review

Two rounds of `/thermo-nuclear-code-quality-review` with fresh context. It found
two bugs in my own work that shipped in earlier commits of this branch:

1. **SVG sphere logos as `og:image`** — a fallback I added would have previewed
   with no image at all for any sphere using an SVG logo, which is what the
   upload UI recommends. Fallback removed.
2. **Double-escaped entities** — the encounter description ran `striptags` over
   rendered Markdown, which is not `mark_safe`d, so autoescape hit it twice and
   "Dungeons & Dragons" reached crawlers as `Dungeons &amp;amp; Dragons`. Now
   built as plain text in the view.

It also caught `panel/print/base.html` overriding one of the three description
tags, two spellings of the same root-sphere predicate, panel titles reading
"Sphere • Sphere • Zagrajmy", and two tests that could not fail. All fixed;
third round approved.

One test from `main` was assertion-scoped rather than worked around:
`test_session_enroll_consent` asserted the bare word "waitlist" was absent from
the whole document, which breaks for any page copy containing it. It now asserts
on the form control.

## Testing

`tests/integration/web/test_page_metadata.py` — titles on root and sub-spheres,
the default on both audiences, generated descriptions on event and encounter
pages, the empty-description fallback, an encounter description containing `&`,
an encounter with neither place nor description, and the image fallback chain.

Full suite green except `test_healthz.py::TestMediaUrls`, which fails on any
machine without `/var/www/ludamus` (`MEDIA_ROOT` from `.env.development`) and
passes with `MEDIA_ROOT` pointed at a writable dir. Pre-existing, unrelated.

`tingle check`: no new debt.

No screenshots: the whole diff lives in `<head>`. The one visible change is the
browser tab title.


hasparus:
Ran a strict maintainability review over the branch. Fixed in a627174b:

**og:image regression for SVG sphere logos** — the sphere-logo fallback I added
would have previewed with *no image at all* for any sphere with an SVG logo,
which is the format the upload UI recommends. Dropped the fallback; the default
is the brand PNG. Per-sphere preview images need a field with format and
dimension validation, not a reuse of the print logo — out of scope here.

**Raw Markdown in the encounter description** — it truncated the Markdown
*source*, so `# Heading` and `**bold**` shipped to search results and link
previews. Now reads the already-rendered HTML through `striptags`. This was
pre-existing; the diff moved the code, so it got fixed.

**Two mechanisms for one thing** — I had descriptions coming from a view context
var on one page and template blocks on another. Now blocks everywhere, with the
encounter's composed string in one partial that the three blocks include, so the
logic exists once. The Python helper and the context var are gone.

**Tests that couldn't fail** — the title test asserted `endswith("• {sphere}")`,
which still holds if the brand-tail guard is deleted entirely (root sphere name
== current sphere name). Now asserts the whole title. The metadata tests also
asserted no status code, so they passed against a 404 — that page extends
`base.html` too. All go through `assert_response` now. Added a case for an
encounter with no place and no description.

**Description length** — 185 chars, past Google's ~155 cut, so the brand clause
was the part getting truncated away. Now 113 (EN) / 114 (PL).

Not changed, deliberately:

- **Seven meta blocks for three values.** A Django block can't hand its output
  to a variable, so one-value-three-tags means three blocks. The alternative —
  view context vars for everything — needs `page_description` in `EventPageView`,
  and 59 exact-context assertions across the event-page tests would need
  updating. That is a bigger and riskier diff than the problem. Object pages now
  override with one shared expression or partial, so no logic is duplicated.
- **`components/meta_image_url.html` emitting whitespace inside the attribute.**
  Real, but pre-existing and now confined to the two object pages that already
  had it; the site-wide default is a clean inline URL. Worth a follow-up filter,
  not worth widening this PR.
- **Brand tail reads `root_sphere.name`** rather than a settings constant. It is
  data-driven and de-duplicates naturally on the root sphere itself. If the root
  sphere row is ever renamed, the tab titles follow — that's a one-row fix, and
  a constant would need a second source of truth for the same string.

PR #708, opened 2026-07-27
Decide enrollment per actor across overlapping windows

### LINKED ISSUE #705: Overlapping enrollment windows combine inconsistently; capacity is not actor-aware
Overlapping enrollment windows are combined by four different rules, and no combination of them is safe with the current model.

| Field | Rule when windows overlap | Where |
|---|---|---|
| `percentage_slots` | `max(...)` | `models.py` `get_most_liberal_config` |
| `allow_anonymous_enrollment` | `any(...)` | `links/db/django/enrollment.py` |
| `restrict_to_configured_users` | `any(...)` | `adapters/web/django/views.py` |
| `max_waitlist_sessions`, `banner_text` | read off the `percentage_slots` winner | `adapters/web/django/forms.py` |

Two consequences today:

1. **`any()` on `restrict_to_configured_users` is over-restrictive.** One restricted window makes the whole event restricted, so adding a restricted window silently locks everyone out of an overlapping open window's seats.
2. **The method is named `get_most_liberal_config` but "liberal" means only "highest seat percentage."** That one row then dictates unrelated settings (waiting-list limit, banner). The name invites exactly the wrong mental model — it is how the misleading panel copy got written in the first place.

### Why the obvious fix doesn't work

PR #656 tried flipping `restrict_to_configured_users` to `all()` (union: a window grants access for its period). That was reverted in `6fd4a067` because it opens a capacity leak: with restriction unioned, a restricted 100 % window overlapping an open 20 % window stops gating the event, but `get_available_slots` still returns **100 %** — the restricted pool goes to everyone.

The root cause is structural: **seat capacity is modelled per session, eligibility is per user.** Any global collapse of the windows into one config or one boolean will leak in one direction or the other.

### Suggested shape

Resolve the actor's usable windows once, then aggregate over that set only:

```python
def enrollment_policy(self, session=None, *, is_configured_user: bool) -> EnrollmentPolicy:
    windows = [w for w in self.get_eligible_enrollment_configs(session)
               if is_configured_user or not w.restrict_to_configured_users]
    # percentage = max(...), waitlist = max(...), anonymous = any(...)
```

One value object replaces `get_most_liberal_config` + the scattered `any()`/`max()` reads, and seats can no longer come from a window the actor isn't in. This touches all five `get_most_liberal_config` call sites (`views.py:957`, `views.py:1296`, `forms.py:377`, `models.py:940`, `links/db/django/enrollment.py:136`), several of which have no actor in scope today — hence a standalone change rather than a rider on a panel-settings PR.

Worth doing alongside: rename `get_most_liberal_config`, and decide whether the panel should warn when windows overlap (see the removed copy in #656).

Context: found during the #656 review. Sibling follow-ups: #699, #700.

hasparus:
Closes #705.

## The bug

An event can have several enrollment windows open at once. Permission was read off a **single winner** — the row with the highest `percentage_slots` — so a restricted window shadowed an open one:

> Window A: 100% of seats, `restrict_to_configured_users`
> Window B: 20% of seats, open to everyone
>
> A user without an enrollment config was blocked **entirely**, even though B should admit them.

Four different rules combined overlapping windows, and they disagreed:

| Field | Rule | Where |
|---|---|---|
| `percentage_slots` | `max(...)` | `Event.get_most_liberal_config` |
| `allow_anonymous_enrollment` | `any(...)` | `links/db/django/enrollment.py` |
| `restrict_to_configured_users` | winner | `adapters/web/django/forms.py`, `views.py` |
| `max_waitlist_sessions`, `banner_text` | winner | `adapters/web/django/forms.py` |

Capacity being actor-blind is what made any single global collapse unsafe: seats could be drawn from a window the actor was not allowed into. #656 tried flipping restriction to `all()` and had to revert (`6fd4a067`) for exactly this reason.

## The fix

Stop collapsing globally. Resolve the windows an actor may use **first**, then aggregate over that set only.

`EnrollmentPolicy` (`mills/enrollment.py`) is a frozen value object over one actor's usable windows:

```python
EnrollmentPolicy.for_actor(windows, is_configured_user=...)
    .can_enroll / .seating_window / .percentage_slots
    .max_waitlist_sessions / .requires_slot_allowance
    .available_slots(participants_limit=..., enrolled_count=...)
```

`banner_text` and `allow_anonymous_enrollment` are **not** migrated: the first is still read off the winner, the second still combined by `any(...)` in `links/db/django/enrollment.py`. Both are actor-blind paths where that is currently fine; #712 covers them.

Selection happens once in `for_actor`; every aggregate reads the same set, so permission, capacity, waiting-list allowance and anonymous access can no longer disagree about which windows apply.

Migrated onto it: `create_enrollment_form`'s `user_can_enroll`, `_can_join_waitlist`, the three winner-based restriction reads in `forms.py`, `_validate_request`, `_route_wants_in`, and the two `restrict_to_configured_users` reads in `views.py`.

**One predicate for "is this user configured".** The capacity path had grown its own weaker version (`virtual_config(...)` truthy, which is true even at `allowed_slots=0`) while permission used `has_slot_access` (`config and config.allowed_slots`). A user with a 0-slot row was "configured" for one and not the other. Both now go through `has_slot_access`.

**Capacity and allowance come from one window.** A seat comes out of a single window, so `percentage_slots` and `requires_slot_allowance` both read from `seating_window` — the widest window the actor may use. Take the restricted window's pool and you owe its per-user allowance; if a wider open window is in reach, you owe nothing.

## What deliberately did not change

- **Guests stay on the session-wide pool.** They are walk-ins exempt from the membership restriction — `test_restricted_viewer_without_slots_can_still_bring_guests` pins this, and it caught an earlier revision of this branch that had quietly narrowed guest capacity to the viewer's own window.
- `get_most_liberal_config` stays for the callers with no actor in scope: `Session.effective_participants_limit` (hence `is_full` and anonymous seating) and waiting-list promotion's `available_seats`. Filed as #712.

## Two regressions this branch introduced and then fixed

Recorded because the first review round called both "pre-existing" on my say-so, and checking against `main` in a scratch worktree proved otherwise.

**Opening permission without moving capacity.** Once `user_can_enroll` came from the policy, an actor the restricted window excludes became able to submit — but `_is_capacity_invalid` still measured every seat against the session-wide pool. On the one-click path (which never reaches `_route_wants_in`) they were seated straight out of the restricted window. On `main` they were rejected at permission, so the actor-blind check was unreachable for them; this branch made it reachable. Account-holder seats are now checked against the per-actor pool, guests against the session-wide one.

**Aggregating capacity and allowance apart.** `percentage_slots` took `max(...)` while `requires_slot_allowance` took `all(...)`, so a configured actor with a restricted 100% window beside an open 20% one got the restricted pool *and* no per-user cap — a combination neither window grants. That is the same "independent rules disagree" failure this PR exists to fix, reproduced one level down inside the value object. Both now read from `seating_window`, the single window the seat comes out of.

## Deferred, not settled

Two structural findings from review are **filed on #712 rather than fixed here**, and neither is resolved by the layering argument:

- `EnrollmentWindowLike` (the Protocol `EnrollmentPolicy` accepts) duplicates `EnrollmentWindowDTO`, and exists only so `mills` can take Django model instances handed down from `adapters` — against the DTO rule. `frozen=True` over mutable ORM rows is a value object in name only.
- The capacity formula now has three copies. `links -> mills` *is* permitted by the importlinter contract, so "the layer rule forbids it" is not the reason it was skipped; the reason is that the clean fix is one `Session.available_slots(...)` in `links` that the other two delegate to, which is its own change.

## Verification

- `mise run check` exit 0, pylint 10.00/10
- 3,233 passed, 9 skipped (one unrelated local failure: `test_healthz` wants `/var/www/ludamus/media`)
- `mise run messages-check` — catalog fresh, 1,661 translated
- `tingle check` — no new debt, **19 paid off**

Six integration tests, each run against current `main` (`28674945`) in a scratch worktree rather than assumed:

Fail on `main` — they pin the fix:
- `test_post_open_window_admits_a_viewer_the_restricted_window_excludes` — the headline bug. On `main` the row renders `disabled=True, reason='Enrollment not available'`.
- `test_post_seats_come_from_the_open_window_not_the_restricted_one` — with the open window's share exhausted, the viewer goes to the waiting list rather than taking a restricted-pool seat.
- `test_post_a_freed_seat_does_not_invent_headroom_past_the_open_window` — mixed cancel+enroll batch.

Pass on `main` — they pin what this branch must not break:
- `test_post_one_click_seats_come_from_the_open_window_too` — the explicit one-click path.
- `test_post_restricted_pool_still_costs_a_slot_beside_an_open_window` — the per-user slot cap survives an overlapping open window.
- `test_dropping_a_guest_gives_no_seat_outside_the_viewers_window` — passes on `main` (permission blocks first) and fails on this branch's own `3c7f4dac`, so it pins a regression introduced *and* fixed inside the branch: freed seats were credited to a pool already clamped at zero.

Plus 25 unit tests in `tests/unit/test_enrollment_policy.py`, including the `seating_window` tie-break — inverting that tie-break fails them, which it did not before review round 4 caught that it was untested.

## Review note

The panel ships multi-window CRUD to organizers with no overlap warning, and the per-window copy ("Choose who may enroll while this window is active") promises per-window semantics. That now holds for enrollment decisions; it still does not for display capacity. Worth a follow-up on the panel side.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved enrollment and waitlist decisions across multiple enrollment windows.
  - Corrected available-seat calculations for restricted and open enrollment periods.
  - Prevented freed guest seats or cancellations from incorrectly expanding enrollment access.
  - Ensured restricted enrollment pools continue to count toward overall capacity.
  - Improved messaging when enrollment limits prevent adding participants.

- **Tests**
  - Added coverage for enrollment windows, capacity limits, waitlists, guest seats, and cancellations.


hasparus:
Filed #712 for the paths this PR leaves actor-blind: anonymous seating (`lock_seating` → `is_full` → `get_most_liberal_config`), anonymous access (a separate `any()` in `links/db/django/enrollment.py`), waiting-list promotion's `available_seats`, and the one-click explicit-action path that never reaches `_route_wants_in`.

Decision recorded there: anonymous enrollers count as unconfigured, so they should draw only on unrestricted windows. Guests deliberately stay on the session-wide pool.

PR #714, opened 2026-07-28
Forward screen_hint=signup to Auth0

hasparus:
## Problem

The landing page's sign-up CTA links `do/login?screen_hint=signup` (`notice_board/landing.html:22`), but `Auth0LoginActionView` only reads `next`, so the hint was silently dropped and new users always landed on Auth0's **login** screen instead of **signup**.

## Change

- `Auth0LoginActionView` now reads `screen_hint`, forwards it to `oauth.auth0.authorize_redirect(...)` (authlib passes extra kwargs through as `/authorize` query params, which Auth0's Universal Login honors), and preserves it across the canonical-domain redirect taken when login starts on a sphere subdomain.
- Only the value `signup` is forwarded; anything else is ignored, so the parameter can't be used to inject arbitrary query params into the authorize URL.

## Verification

Integration tests cover: hint forwarded to `authorize_redirect`, unknown hint dropped, and hint preserved (together with `next`) through the non-root-domain redirect. `mise run lint` and the full `test:py` suite are green. No visual changes, so no screenshots.

---
_Generated by [Claude Code](https://claude.ai/code/session_015rbYdbfV2KQwnQpsLL32LT)_


hasparus:
@coderabbitai review

---

PR #715, opened 2026-07-28
Write-in answers split on semicolons, not commas

hasparus:
## Problem

The write-in companion input merged in #701 splits typed answers on every comma (`merge_custom` → comma split), but the answers people actually type are full of commas — the accessibility field's real options look like *"dostosowany do osób w spektrum autyzmu (m.in. przyciemnione światło, ściszone dźwięki)"*. Type one of those as a custom value and it is silently stored as two (or four) fragments.

## Change

Semicolon becomes the write-in separator — the smaller of the two options discussed (the other being a dedicated chips/tags input component, which this doesn't preclude):

- `merge_custom` splits the companion value on the named `WRITE_IN_SEPARATOR`; `split_stored` joins unmatched values back with it for redisplay, so comma-bearing answers round-trip intact (pinned by a unit test).
- `split_answers` became `split_imported_answers` with a mandatory `known` — the Google Forms **import** path deliberately keeps commas: Forms joins checkbox answers with `", "`, and its known-options matching already protects comma-bearing options there.
- The field hint lives in one shared partial now (`components/write-in-hint.html`) instead of four copy-pasted spans, and says "semicolon separated" ("po średniku" in the PL catalog).
- No data migration: the separator only exists at the input/redisplay boundary; stored answers are plain lists.

## Known trade-off

A single text input needs *some* textual delimiter, so the inverse edge exists: a stored write-in that legitimately contains `"; "` (possible for values saved under the old comma regime, or single-value fields) will be split into two values on its next edit-and-save round trip. Semicolons are rare enough in these answers that we accept this rather than backfill; commas — the case users actually hit — now survive.

## Verification

- Unit tests: a comma-bearing write-in survives as one value, and round-trips through `split_stored` → `merge_custom`.
- Updated the propose/session-edit integration tests and the `write-in-fields` e2e spec to the semicolon contract; the facilitator edit page asserts the comma-bearing write-in renders back into its companion input.
- Full `test:py` suite green; `mise run check` green; `messages-check` fresh.

If we later want the nicer UX, a chips input component can replace the raw text input without touching the storage contract introduced here.

## Summary by CodeRabbit

* **New Features**
  * Write-in values in multi-value fields now use semicolons as separators.
  * Commas within individual write-in values are preserved.
  * Updated interface hints and Polish translations reflect the new format.

* **Bug Fixes**
  * Improved handling and round-tripping of write-in answers containing commas.

* **Tests**
  * Added and updated coverage for semicolon-separated values and comma preservation.


hasparus:
@coderabbitai review

---

hasparus:
Addressed in cb3724ec: the POST test now asserts the response via `assert_response` before checking stored state, the e2e spec reaches the custom input through its accessible label (`Or type a custom value` — the seed's only `allow_custom` field, so it's unique), and the `custom_values` docstring became a comment.

Two suggestions deliberately skipped:

- **Keyword-only fixture params** — the 3+-params-keyword-only rule governs production code; every test in this file (and the suite) takes pytest fixtures positionally, and only non-test helpers use `self, *,`. Changing one signature would break the file's convention without pytest gaining anything.
- **Asserting the comma value renders in the GET test** — repo testing policy sends rendered-markup claims to `tests/e2e` and forbids adding them to integration tests we touch; the POST round-trip test pins the same value at the storage level instead.

---

PR #722, opened 2026-07-28
Extract shared status-page hero partial

hasparus:
## Problem

Four pages hand-rolled the same hero shell — icon in a rounded tile, heading, muted lead: `login_required.html`, `auth_error.html` (its third copy triggered the rule of three in the #713 reviews), `404_dynamic.html`, and `500_dynamic.html` (which also duplicated inline SVGs and untranslated button labels).

## Change

- New `components/status_hero.html`: icon tile + optional error-code chip + heading + lead. A non-empty `error_code` selects the compact variant (code beside the tile, tighter heading/lead) — the real existing distinction between the auth pages and the error pages, driven by one input instead of class-string params.
- All four pages converted. Action-button rows stay at the call sites — they differ structurally (login-button component vs plain links vs `history.back()`).
- Every partial input is passed explicitly with `only` at each call site (the strict template checker requires it; `error_code=""` where there's no HTTP status).
- Bonus on the 500 page: hero and buttons now use `{% icon %}` + `{% translate %}` like the 404 page (msgids already had PL translations).

Rendered output is byte-identical for the base-extending pages, except the error-page `h1` gains a redundant `text-foreground` (already set on `body`; zero visual change).

Net −23 lines.

## Verification

- Full `test:py` suite: 3291 passed (page tests render through the partial, so the strict checker validated every input).
- `mise run lint`, `tingle check` (no new debt), `messages-check` (catalog fresh, no new msgids) all green.
- Screenshots of login-required, auth-error, and 404 captured on the branch — pages render identically through the partial.

---
_Generated by [Claude Code](https://claude.ai/code/session_015rbYdbfV2KQwnQpsLL32LT)_

## Summary by CodeRabbit

- **UI Improvements**
  - Updated 404 and 500 error pages with a consistent status layout and clearer visual presentation.
  - Refreshed login-required and authentication error screens using the same polished status design.
  - Improved error-page actions with consistent icons and translated “Go Back” and “Go to Home” labels.
  - Adjusted layouts for error codes, headings, and supporting messages to improve readability across screen sizes.
- **Localization**
  - Updated translation references for the “Go Back” message used on error pages.

### DIFF
--- a/src/ludamus/locale/pl/LC_MESSAGES/django.po
+++ b/src/ludamus/locale/pl/LC_MESSAGES/django.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2026-07-28 19:18+0200\n"
+"POT-Creation-Date: 2026-07-29 00:14+0200\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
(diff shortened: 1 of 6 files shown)


hasparus:
Re the nitpick asking for template assertions on `status_hero.html` (compact classes, rendered code, icon output): deliberately skipped — repo testing policy sends rendered-markup claims to `tests/e2e` and forbids adding them to integration tests ("assert status, redirect, context, and state in Python; assert rendered HTML in tests/e2e"). The partial is already exercised through all four include sites by the existing page integration tests (35 of them render it, and the strict template-variable checker validates every input on each render), and this PR's review verified old-vs-new output render-identical on all four pages. Adding class-string assertions in Python would test the template engine, not the pages.

---

PR #728, opened 2026-07-29
refactor(time-slots): migrate panel view to request.services

hasparus:
Migrates `gates/web/django/chronology/panel/views/time_slots.py` (9 `request.di.uow.time_slots` call sites) to the `request.services` shape, per docs/agents/services-migration.md.

## What moved where

- **`pacts/event.py`** — new `PanelTimeSlotsServiceProtocol` (placed here instead of the contested `pacts/legacy.py`/`pacts/chronology.py`; time slots belong to the event noun). No new DTOs — `TimeSlotDTO` already exists.
- **`mills/panel_time_slots.py`** (new) — `PanelTimeSlotsService`; constructor takes `TimeSlotRepositoryProtocol` + `TransactionProtocol` (keyword-only). Owns the transactional boundary: `create`/`update` fold the validate-then-write sequence into one `atomic()` block (previously the view validated and wrote without a transaction), reusing `PanelService.validate_time_slot` from `mills/legacy.py` rather than duplicating the rules. `update`/`delete` scope the request-supplied pk to the current event via `read_by_event` inside the transaction, so a foreign pk raises `NotFoundError` with no side effects.
- **`pacts/services.py` / `inits/services.py`** — `panel_time_slots` leaf added to `ServicesProtocol` and `Services` (the `time_slots` repo leaf already existed in `inits/repositories.py`).
- **View** — shrank to glue: form parsing, redirects, flash messages. Service validation errors come back as a `list[str]` the view translates onto the form, same as before. Duplicated datetime-combining code extracted into `_slot_times`.
- **`tests/unit/test_panel_time_slots_service.py`** (new) — 12 unit tests mocking the repo protocol + transaction directly, including foreign-pk 404-without-side-effects cases for read/update/delete.

Pure refactor: no template, URL, copy, or behavior change, so no screenshots (nothing visual changed).

## Deviations from the recipe

- The service lives in its own `mills/panel_time_slots.py` instead of an existing noun module because open PRs edit `mills/timeslots.py`, `mills/chronology.py`, and `mills/legacy.py`.
- `PanelService.delete_time_slot` in `mills/legacy.py` is now dead, but that file is off-limits for the same reason, so it gets a commented vulture `ignore_names` entry in `pyproject.toml` instead of deletion. Follow-up: remove the method and the ignore entry once #719/#625/#626 land.

## Tingle

`tingle check --base origin/main`: **no new debt, 24 paid off** — `request-di-uow` −9, `old-subdomain-loc` −15.

## Test status (sandbox, honest)

`mise` is unavailable in this sandbox, so the underlying commands ran through the poetry venv directly:

- `pytest -n auto tests/integration tests/unit`: **3370 passed, 9 skipped** (includes the 55 existing time-slot view integration tests, unweakened, and the 12 new unit tests)
- mypy `src`: clean; pylint `src` and `tests`: 10.00/10; ruff check: clean; import-linter: 7/7 contracts kept; vulture: clean; black/ruff format: clean on changed files; codespell: clean
- Not run (tool missing in sandbox, config untouched): hk, taplo, ast-grep, djlint, impeccable, e2e

---

## Summary by CodeRabbit

* **New Features**
  * Improved event time-slot management with consistent validation for ordering, event boundaries, and overlapping slots.
  * Time slots can now be safely created, edited, listed, and deleted through the panel.
  * Deletion is prevented when a time slot has associated proposals.
  * Added Polish translations for time-slot validation messages.

* **Bug Fixes**
  * Improved handling of missing or event-mismatched time slots.
  * Overnight time slots now calculate start and end times consistently.

* **Tests**
  * Added coverage for time-slot validation, updates, deletion rules, and error handling.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -463,6 +463,10 @@ ignore_names = [
     # Dead since panel_time_slots service owns deletion; remove from
     # mills/legacy.py (with this entry) once the PRs editing that file land
     "delete_time_slot",
+    # Dead since panel_time_slots owns validation (typed-enum copy lives
+    # there); remove from mills/legacy.py (with this entry) alongside
+    # delete_time_slot
+    "validate_time_slot",
     # Unused but stays
     "deleted_at",
     "modification_time",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -460,6 +460,9 @@ ignore_names = [
     "get_navbar",
     # Pydantic
     "model_config",
+    # Dead since panel_time_slots service owns deletion; remove from
+    # mills/legacy.py (with this entry) once the PRs editing that file land
+    "delete_time_slot",
     # Unused but stays
     "deleted_at",
     "modification_time",
--- a/src/ludamus/mills/panel_time_slots.py
+++ b/src/ludamus/mills/panel_time_slots.py
@@ -1,0 +1,68 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from ludamus.mills.legacy import PanelService
+from ludamus.pacts.event import PanelTimeSlotsServiceProtocol
+
+if TYPE_CHECKING:
+    from datetime import datetime
+
+    from ludamus.pacts.legacy import EventDTO, TimeSlotDTO, TimeSlotRepositoryProtocol
+    from ludamus.pacts.services import TransactionProtocol
+
+
+class PanelTimeSlotsService(PanelTimeSlotsServiceProtocol):
+    def __init__(
+        self,
+        *,
+        transaction: TransactionProtocol,
+        time_slots: TimeSlotRepositoryProtocol,
+    ) -> None:
+        self._transaction = transaction
+        self._time_slots = time_slots
+
+    def list_for_event(self, event_id: int) -> list[TimeSlotDTO]:
+        return self._time_slots.list_by_event(event_id)
+
+    def read(self, *, event_id: int, pk: int) -> TimeSlotDTO:
+        return self._time_slots.read_by_event(event_id, pk)
+
+    def create(
+        self, *, event: EventDTO, start_time: datetime, end_time: datetime
+    ) -> list[str]:
+        with self._transaction.atomic():
+            existing = self._time_slots.list_by_event(event.pk)
+            errors = PanelService.validate_time_slot(
+                start_time, end_time, event, existing
+            )
+            if not errors:
+                self._time_slots.create(event.pk, start_time, end_time)
+            return errors
+
+    def update(
+        self, *, event: EventDTO, pk: int, start_time: datetime, end_time: datetime
+    ) -> list[str]:
+        with self._transaction.atomic():
+            # Scope the pk to the panel's event before writing; a foreign pk
+            # raises NotFoundError with no side effects.
+            self._time_slots.read_by_event(event.pk, pk)
+            existing = [
+                slot
+                for slot in self._time_slots.list_by_event(event.pk)
+                if slot.pk != pk
+            ]
+            errors = PanelService.validate_time_slot(
+                start_time, end_time, event, existing
+            )
+            if not errors:
+                self._time_slots.update(pk, start_time, end_time)
+            return errors
+
+    def delete(self, *, event_id: int, pk: int) -> bool:
+        with self._transaction.atomic():
+            self._time_slots.read_by_event(event_id, pk)
+            if self._time_slots.has_proposals(pk):
+                return False
+            self._time_slots.delete(pk)
+            return True
--- a/src/ludamus/mills/panel_time_slots.py
+++ b/src/ludamus/mills/panel_time_slots.py
@@ -2,14 +2,49 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING
 
-from ludamus.mills.legacy import PanelService
-from ludamus.pacts.event import PanelTimeSlotsServiceProtocol
+from ludamus.pacts.event import PanelTimeSlotsServiceProtocol, TimeSlotValidationError
 
 if TYPE_CHECKING:
+    from collections.abc import Sequence
     from datetime import datetime
 
-    from ludamus.pacts.legacy import EventDTO, TimeSlotDTO, TimeSlotRepositoryProtocol
+    from ludamus.pacts.legacy import (
+        DateTimeRangeProtocol,
+        EventDTO,
+        TimeSlotDTO,
+        TimeSlotRepositoryProtocol,
+    )
     from ludamus.pacts.services import TransactionProtocol
+
+# TODO(hasparus): fold-back plan for the legacy strangler, blocked while
+# open PRs hold mills/legacy.py:
+#   1. delete PanelService.validate_time_slot from mills/legacy.py (its logic
+#      moved to _validate_time_slot below) and drop its vulture entry,
+#   2. delete PanelService.delete_time_slot and its vulture entry,
+#   3. fold this module into the event noun module so pacts/event.py and
+#      mills/event.py mirror each other and the name stops colliding with
+#      mills/timeslots.py.
+
+
+def _validate_time_slot(
+    *,
+    start: datetime,
+    end: datetime,
+    event: EventDTO,
+    existing_slots: Sequence[DateTimeRangeProtocol],
+) -> list[TimeSlotValidationError]:
+    errors: list[TimeSlotValidationError] = []
+
+    if start >= end:
+        errors.append(TimeSlotValidationError.START_NOT_BEFORE_END)
+
+    if start < event.start_time or end > event.end_time:
+        errors.append(TimeSlotValidationError.OUTSIDE_EVENT_DATES)
+
+    if any(start < slot.end_time and end > slot.start_time for slot in existing_slots):
+        errors.append(TimeSlotValidationError.OVERLAPS_EXISTING_SLOT)
+
+    return errors
 
 
 class PanelTimeSlotsService(PanelTimeSlotsServiceProtocol):
@@ -30,11 +65,15 @@ class PanelTimeSlotsService(PanelTimeSlotsServiceProtocol):
 
     def create(
         self, *, event: EventDTO, start_time: datetime, end_time: datetime
-    ) -> list[str]:
+    ) -> list[TimeSlotValidationError]:
+        # atomic() keeps the write consistent but does not serialize the
+        # check-then-insert: two concurrent requests can both read the same
+        # slots, both pass validation, and insert overlapping slots. Full
+        # enforcement needs a DB exclusion constraint on the slot range.
         with self._transaction.atomic():
             existing = self._time_slots.list_by_event(event.pk)
-            errors = PanelService.validate_time_slot(
-                start_time, end_time, event, existing
+            errors = _validate_time_slot(
+                start=start_time, end=end_time, event=event, existing_slots=existing
             )
             if not errors:
                 self._time_slots.create(event.pk, start_time, end_time)
@@ -42,7 +81,8 @@ class PanelTimeSlotsService(PanelTimeSlotsServiceProtocol):
 
     def update(
         self, *, event: EventDTO, pk: int, start_time: datetime, end_time: datetime
-    ) -> list[str]:
+    ) -> list[TimeSlotValidationError]:
+        # Same unserialized check-then-write race as in create().
         with self._transaction.atomic():
             # Scope the pk to the panel's event before writing; a foreign pk
             # raises NotFoundError with no side effects.
@@ -52,8 +92,8 @@ class PanelTimeSlotsService(PanelTimeSlotsServiceProtocol):
                 for slot in self._time_slots.list_by_event(event.pk)
                 if slot.pk != pk
             ]
-            errors = PanelService.validate_time_slot(
-                start_time, end_time, event, existing
+            errors = _validate_time_slot(
+                start=start_time, end=end_time, event=event, existing_slots=existing
             )
             if not errors:
                 self._time_slots.update(pk, start_time, end_time)
--- a/src/ludamus/pacts/event.py
+++ b/src/ludamus/pacts/event.py
@@ -1,8 +1,11 @@
-from typing import Protocol
+from typing import TYPE_CHECKING, Protocol
 
 from pydantic import BaseModel
 
-from ludamus.pacts.legacy import EventDTO, PanelStatsDTO
+from ludamus.pacts.legacy import EventDTO, PanelStatsDTO, TimeSlotDTO
+
+if TYPE_CHECKING:
+    from datetime import datetime
 
 
 class EventPanelContextDTO(BaseModel):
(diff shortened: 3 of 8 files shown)


src/ludamus/pacts/event.py
```diff
@@ -14,3 +17,15 @@ class EventPanelContextDTO(BaseModel):
...
+
+class PanelTimeSlotsServiceProtocol(Protocol):
+    def list_for_event(self, event_id: int) -> list[TimeSlotDTO]: ...
+    def read(self, *, event_id: int, pk: int) -> TimeSlotDTO: ...
+    def create(
+        self, *, event: EventDTO, start_time: datetime, end_time: datetime
+    ) -> list[str]: ...
+    def update(
+        self, *, event: EventDTO, pk: int, start_time: datetime, end_time: datetime
+    ) -> list[str]: ...
```
hasparus: `create`/`update` returning `list[str]` promotes a stringly-typed error contract into brand-new public API surface. The strings are magic values: the view calls `_(error)` on a runtime variable, which `makemessages` cannot extract — and indeed none of the three messages (`"Start must be before end."`, `"Time slot must be within event dates."`, `"Time slot overlaps with an existing slot."`) exists as a msgid in `locale/pl/LC_MESSAGES/django.po`. Polish organizers see English errors today, and nothing in the type system stops a service-side rewording from silently orphaning the catalog entry.

The behavior is pre-existing (the old `_validate_time_slot` did the same), so this is not a blocker for the migration PR — but this protocol is the one place where the contract gets frozen, and it's much cheaper to fix now than after more services copy the shape. Suggestion: return a typed error (e.g. a `SlotValidationError` StrEnum in `pacts`), and have the view map enum → literal `_("...")` strings so extraction works and the boundary is explicit. `validate_time_slot` in `mills/legacy.py` would need its return type widened, so if that file is off-limits, at minimum fold this into the tracked follow-up rather than letting the `list[str]` shape become the precedent.

---

src/ludamus/mills/panel_time_slots.py
```diff
@@ -0,0 +1,68 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from ludamus.mills.legacy import PanelService
```
hasparus: This import points the wrong way: a new, clean mill now depends on `mills/legacy.py`, and it reaches through a UoW-based service class just to call a pure staticmethod. After this PR, `PanelService.validate_time_slot` has exactly one caller — this module — so the logic's real home is here (or `specs/`, since it's a pure invariant), with nothing left in legacy.

I understand `mills/legacy.py` is frozen while #719/#625/#626 are open, so the import is the least-bad option *today*. But the PR's stated follow-up only covers `delete_time_slot` + the vulture entry. The full fold-back debt is bigger, and it lives only in PR prose that nobody will find in six months:

1. move `validate_time_slot` out of `mills/legacy.py` into this module (or specs) and drop this import,
2. delete `delete_time_slot` and its `pyproject.toml` vulture entry,
3. fold this module into the noun module per the pacts↔mills mirror rule — the protocol sits in `pacts/event.py` but the impl in a one-off `panel_time_slots.py`, which also collides confusingly with the existing `mills/timeslots.py`.

Please record the whole list as a TODO comment in this module (or a tracking issue referenced here), so the temporary placement stays discoverable from the code itself.

---

src/ludamus/mills/panel_time_slots.py
```diff
@@ -0,0 +1,68 @@
...
+        self, *, event: EventDTO, start_time: datetime, end_time: datetime
+    ) -> list[str]:
+        with self._transaction.atomic():
+            existing = self._time_slots.list_by_event(event.pk)
+            errors = PanelService.validate_time_slot(
+                start_time, end_time, event, existing
+            )
+            if not errors:
+                self._time_slots.create(event.pk, start_time, end_time)
+            return errors
```
hasparus, line 80: Non-blocking, pre-existing: wrapping list-validate-insert in `atomic()` is the right structure (and an improvement — the old view wrote with no transaction at all), but be aware it doesn't actually enforce the overlap invariant. Without row locking or a DB constraint, two concurrent creates can both read the same `existing`, both pass validation, and both insert overlapping slots. The `atomic()` block can read as if it guarantees the invariant, so a one-line comment here (or a follow-up Postgres exclusion constraint on the slot range) would keep the next reader from over-trusting it. Same applies to `update` below.

---

hasparus (commented):
Thermo-nuclear code quality review. Verdict: **COMMENT** — no defect found; the findings below are boundary-contract and debt-tracking issues, not blockers.

I verified the diff against the migration recipe (`docs/agents/services-migration.md`) and the GLIMPSE rules: the view shrank to glue, the service takes the specific repo protocol + `TransactionProtocol` keyword-only, transactions moved out of the view, request-supplied pks are scoped to the event inside the service with foreign-pk-raises-without-side-effects unit tests, and the duplicated datetime-combining block in create/edit POST was extracted into `_slot_times` instead of being copied a third time. The integration suite is untouched as the regression guard, per the recipe. That's the migration done properly, and the tingle delta is genuinely negative.

Three findings, in priority order:

1. **`pacts/event.py:25-30` — the new protocol freezes a stringly-typed error contract.** `list[str]` of magic English strings, translated via `_(variable)` so the msgids never reach the catalog (they're absent from `django.po` — Polish users get English errors). Pre-existing behavior, but this PR is the moment the shape becomes public API; a typed error enum in `pacts` with view-side literal `_()` mapping would fix both the boundary and the i18n hole. Cheap now, expensive after the next few services copy the pattern.

2. **`mills/panel_time_slots.py:5` — clean mill importing `mills.legacy` for a pure staticmethod it is now the sole caller of.** Defensible while `legacy.py` is frozen by open PRs, but the fold-back debt (move `validate_time_slot`, delete `delete_time_slot` + vulture entry, fold the module into the noun module per the pacts↔mills mirror rule — it also name-collides with the existing `mills/timeslots.py`) is currently recorded only in PR prose. Please put the full list in a TODO in the module or a tracking issue.

3. **`mills/panel_time_slots.py:34-41` — `atomic()` doesn't serialize the overlap check.** Pre-existing race, structurally improved by this PR (there was no transaction before), just don't let the block read as a guarantee; a comment or an eventual exclusion constraint closes it.

The deviations the PR flags (own module, vulture entry) are honestly stated and justified given the open PRs; my only ask is that the follow-up be tracked in code, not just in the PR description.

---

src/ludamus/pacts/event.py
hasparus: Fixed in 6f347131: the protocol now returns `list[TimeSlotValidationError]` (a StrEnum in `pacts/event.py`), and the view maps each code to a literal `_("...")` via match + `assert_never`, same as the anonymous-enrollment pattern — the three messages are now extractable and translated in the Polish catalog. `mills/legacy.py` didn't need widening: since `validate_time_slot` had no other callers left, its logic moved into `mills/panel_time_slots.py` and the legacy staticmethod joined `delete_time_slot` on the vulture whitelist until the PRs holding that file land.

---

src/ludamus/mills/panel_time_slots.py
hasparus: Done in 6f347131: item 1 of your list is already paid — the validation logic moved into this module (as `_validate_time_slot`, now returning the typed enum) and the legacy import is gone; `PanelService.validate_time_slot` sits on the vulture whitelist next to `delete_time_slot` until the PRs holding `mills/legacy.py` land. The remaining fold-back plan (delete both legacy methods + vulture entries, fold this module into the event noun module) is now a TODO at the top of the module instead of PR prose.

---

src/ludamus/mills/panel_time_slots.py
hasparus: Added in 6f347131: `create()` now carries a comment stating that `atomic()` keeps the write consistent but does not serialize the check-then-insert, so concurrent requests can still produce overlapping slots without a DB exclusion constraint, and `update()` points back to it. The constraint itself stays a follow-up since it needs a migration on the legacy model.

---

pyproject.toml
```diff
@@ -460,6 +460,13 @@ ignore_names = [
     "get_navbar",
     # Pydantic
     "model_config",
+    # Dead since panel_time_slots service owns deletion; remove from
+    # mills/legacy.py (with this entry) once the PRs editing that file land
+    "delete_time_slot",
+    # Dead since panel_time_slots owns validation (typed-enum copy lives
+    # there); remove from mills/legacy.py (with this entry) alongside
+    # delete_time_slot
+    "validate_time_slot",
```
hasparus, line 469: Nit, non-blocking — I checked whether these two entries were counter-gaming a counter and they aren't: `tingle.toml` has no metric over vulture's `ignore_names`, so nothing was dodged here, and the branch actually *took* a `todo-comments` +1 for the fold-back note rather than avoiding it. Both names are genuinely dead in `mills/legacy.py` (grep confirms `PanelService.validate_time_slot` and `.delete_time_slot` have zero callers at this tip), and the comments name the removal condition. That's the right call while that file is held by open PRs.

The one real cost: `ignore_names` matches by *name*, globally. Until step 1/2 of the module TODO lands, a genuinely-dead `validate_time_slot` or `delete_time_slot` introduced anywhere else in the tree will be silently whitelisted too. Small window and well-documented, so I wouldn't hold the PR on it — just worth knowing the entries are broader than the two methods they're written for.

---

tests/unit/test_panel_time_slots_service.py
```diff
@@ -0,0 +1,177 @@
...
+        time_slots.create.assert_called_once_with(_EVENT_ID, start, end)
+
+    def test_create_returns_errors_without_writing(self, service, time_slots):
+        time_slots.list_by_event.return_value = [_slot(pk=1)]
+        start = datetime(2026, 6, 1, 11, 0, tzinfo=UTC)
+        end = datetime(2026, 6, 1, 13, 0, tzinfo=UTC)
+
+        errors = service.create(event=_event(), start_time=start, end_time=end)
+
+        assert errors == [TimeSlotValidationError.OVERLAPS_EXISTING_SLOT]
```
hasparus, line 94: Nit, non-blocking — now that the errors are a typed enum, the suite covers `OVERLAPS_EXISTING_SLOT` and `START_NOT_BEFORE_END` but never `OUTSIDE_EVENT_DATES`, and nothing exercises the accumulating case (`start >= end` *and* outside the event window returns two codes, since `_validate_time_slot` appends rather than short-circuits). One extra case with a start before `event.start_time` would close both gaps and pin the list-return contract that the view's `_add_slot_errors` loop depends on.

Worth noting the enum makes this cheap insurance rather than essential: `assert_never` in `_slot_error_message` already makes a fourth member a mypy error, which is the failure mode that used to be silent.

---

hasparus (commented):
## Round 2 — thermo-nuclear review: **APPROVE**

> Submitted as a COMMENT purely because GitHub refuses `APPROVE` on a self-authored PR ("Can not approve your own pull request"). The verdict is **approve** — nothing here is withheld.

I re-read the PR at its tip (`6f34713`) as the review of record, not just the delta, and verified each round-1 claim against the code rather than the replies. All three are genuinely resolved. Two minor nits inline; neither blocks.

### Round-1 findings, verified

**(a) Stringly-typed errors frozen into the protocol — resolved, properly.**
`TimeSlotValidationError` is a `StrEnum` in `pacts/event.py` and the protocol returns `list[TimeSlotValidationError]`, so the contract is typed at the boundary where it gets frozen. The view maps it with `match` + `assert_never` in `_slot_error_message`, which is the same shape as `gates/web/django/chronology/anonymous.py:58` — a house pattern, not a one-off. The literal `_("...")` calls are now extractable, and I confirmed all three msgids are actually present in `locale/pl/LC_MESSAGES/django.po`, with no duplicate-msgid collisions introduced (the pre-existing dupes are `Category`/`Name`/`Session`/etc., untouched).

The Polish is correct per CLAUDE.md: both slot messages use **"przedział czasowy"**, not the forbidden "blok czasowy" that would collide with *track*. "Przedział czasowy musi mieścić się w zakresie dat wydarzenia." and "Przedział czasowy nakłada się na istniejący przedział." both read naturally. The catalog entries also sit in makemessages' natural order for that file (the two helper messages at lines 53–57 land before "Time slot created successfully." at line 213), which is what a real regeneration produces rather than a hand-edit.

**(b) Fold-back debt only in PR prose — resolved.**
The three-step plan is now a `TODO(hasparus)` at the top of `mills/panel_time_slots.py`, discoverable from the code. Item 1 is already paid, not just promised: the `from ludamus.mills import PanelService` import is gone from the view and never appears in the new mill, and the validation lives here as `_validate_time_slot`. Good — the new mill no longer points at `legacy.py` at all.

**(c) `atomic()` doesn't serialize check-then-insert — resolved.**
The comment in `create()` says exactly the right thing (concurrent readers both pass, needs a DB exclusion constraint), and `update()` points back to it instead of duplicating the paragraph. Deferring the constraint itself is correct scoping for a migration PR.

### Behavior-change audit on the legacy move

I diffed the moved logic against `PanelService.validate_time_slot` line by line. The only change is `for … : append; break` → `any(...)`, which is equivalent: same single append, same short-circuit, same error ordering, so the form renders identically. `_slot_times` is a byte-identical extraction of the two previously-duplicated inline blocks, including the `end_time += timedelta(days=1)` midnight-rollover branch guarded on `end_date == start_date`. `update` preserves the `slot.pk != pk` self-exclusion. `delete` is `read_by_event` → `has_proposals` → `delete`, matching the old view + `PanelService.delete_time_slot` pairing, now inside a transaction. No behavior change.

### Judged as a whole

The authz story is the part I'd have blocked on and it's right: every request-supplied pk is scoped to `current_event` via `read_by_event` **inside the service and inside the transaction**, per the panel object-scope rule, not in the view — and the foreign-pk-404s-without-side-effects cases are tested for read, update, *and* delete. The view drops to 309 lines of pure glue with zero `request.di.uow` left. The transactional boundary is a real improvement: create/update previously validated and wrote with no transaction at all.

On structure: I looked for a code-judo move that would delete complexity rather than relocate it, and I don't think one is available inside this PR's blast radius. The obvious one — collapse `mills/panel_time_slots.py` into the event noun module and delete the two dead `legacy.py` methods — is blocked by open PRs holding that file, and it's step 2/3 of the TODO. The duplicated-rules window (the dead copy in `legacy.py` alongside the live one here) is the honest cost of that, and it's recorded in both the module TODO and the vulture comments rather than left implicit. That's the strangler-fig tradeoff done correctly.

The `list[…]` return-errors shape rather than raising is pre-existing and was already litigated in round 1; now that it's typed I don't want to re-open it.

All 17 checks green, including `tingle`, `impeccable`, `test-postgres`, and codecov patch.

Clean, correct increment. Ship it.

---

tests/unit/test_panel_time_slots_service.py
hasparus: Added in 3d89a02: `test_create_returns_outside_event_dates_for_slot_before_event` covers `OUTSIDE_EVENT_DATES` on its own (08:00–09:30, before the event's 09:00 start, no overlap), and `test_create_accumulates_every_broken_rule` uses a reversed 08:00–07:00 range to assert both `START_NOT_BEFORE_END` and `OUTSIDE_EVENT_DATES` come back in one list, pinning the order the view's `_add_slot_errors` loop walks.

---

pyproject.toml
hasparus: Left as-is: I checked vulture 2.16 and there is no narrower form to move to — `ignore_names` is matched by `_match(name, self.ignore_names)` against the bare definition name with no path or qualified-name component, and a whitelist file is no tighter (it just adds the same bare name to the used-names set). `exclude` only works at whole-file granularity, which would be broader, and `ignore_decorators` doesn't apply to plain methods. So the global-by-name cost you describe is real but unavoidable short of splitting the run into two configs, which is a bigger change than the two entries it would scope; steps 1–2 of the module TODO delete them outright.

---

PR #729, opened 2026-07-29
refactor(tracks): migrate panel view to request.services

hasparus:
Migrates `gates/web/django/chronology/panel/views/tracks.py` off the legacy `request.di.uow.*` surface per the [services-migration recipe](docs/agents/services-migration.md). Pure refactor — no visual or behavioral change.

## What moved where

- **`pacts/tracks.py`** (new) — `TracksPanelServiceProtocol` plus `TrackFormContextDTO` / `TrackEditContextDTO` read aggregates. Kept out of `pacts/legacy.py` / `pacts/chronology.py` to avoid conflicting with open PRs #719/#625/#626.
- **`mills/tracks.py`** (new) — `TracksPanelService`: constructor takes `TransactionProtocol` + the specific `tracks` / `spaces` / `spheres` repo protocols (keyword-only, never the UoW). Owns the transactional boundaries for create/update/delete. Same conflict-avoidance reason for not extending `mills/chronology.py`.
- **`pacts/services.py`** — `tracks_panel` property on `ServicesProtocol`.
- **`inits/services.py`** — `tracks_panel` leaf on `Services` (repos already existed in `inits/repositories.py`, no changes there).
- **View** — shrunk to glue: form parsing, flash messages, redirects, template selection. All 11 `request.di.uow` call sites (`tracks` ×9, `spaces` ×1, `spheres` ×1) replaced with `request.services.tracks_panel`.
- **`tests/unit/test_tracks_panel_service.py`** (new) — 9 unit tests mocking the specific repo protocols + transaction, including foreign-id (track slug, space/manager pks) tests asserting no side effects.

## Authz

Object scoping moved from the view into the service: track slugs resolve via the event-scoped `read_by_slug` (foreign slug ⇒ `NotFoundError`, no writes), and submitted space/manager pks are intersected with the event's spaces and the sphere's managers inside `create`/`update`.

## Tingle

`tingle check --base origin/main`: **no new debt, 27 paid off** — `request-di-uow` −11, `old-subdomain-loc` −18 net, `todo-comments` **+1**, `type-object` +1.

An earlier revision of this section claimed `todo-comments` −1. That was wrong: the −1 came from deleting the pre-existing `# TODO(fancysnake): Extract common view boilerplate` marker at the top of the view, and the boilerplate it flags was never extracted — the file still has six copies of the `get_event_context` / `redirect("panel:index")` prologue and three copies of the not-found redirect, and the three sibling panel views carry the identical marker. The marker is restored, so the honest accounting is that this PR takes on one new TODO (the `pacts/tracks.py` follow-up: fold `pacts/tracks.py` + `mills/tracks.py` into the event noun modules once #719/#625/#626 land) and pays off nothing in that metric.

The pre-existing `# pylint: disable=duplicate-code` at the top of the view file is retained (net 0): the remaining duplication is cross-file view boilerplate shared with `personal_data_fields.py` / `facilitators.py`, which this PR must not touch.

## Test/check status (sandbox — no mise; ran via poetry venv on Python 3.14)

- `pytest tests/integration tests/unit`: **3367 passed, 9 skipped** (includes the 29 existing tracks integration tests, unmodified, as the regression guard)
- `ruff check src tests`: pass
- `mypy src`: pass
- `pylint src` and `pylint --disable=duplicate-code tests`: 10.00/10
- `lint-imports` (GLIMPSE contracts): 7 kept, 0 broken
- `black --check`, `codespell`, `vulture` (with the task's unused-variable filter): pass
- Could **not** run in this sandbox: `ast-grep` (tool not installed), `lint:impeccable` (npm fetch blocked), `hk`/`djlint`/`taplo` (no template/toml/shell changes anyway), e2e.

No screenshots: no template or UI change, context shape is identical (integration tests assert it key-for-key).

---


src/ludamus/gates/web/django/chronology/panel/views/tracks.py
```diff
@@ -18,32 +17,25 @@
...
-    return list(submitted & valid_pks)
+def _submitted_track_data(
+    *, request: PanelRequest, form: TrackForm, default_is_public: bool
+) -> TrackUpdateData:
+    return TrackUpdateData(
+        name=form.cleaned_data["name"],
+        is_public=form.cleaned_data.get("is_public", default_is_public),
+        space_pks=_submitted_pks(request, "space_pks"),
+        manager_pks=_submitted_pks(request, "manager_pks"),
+    )
```
hasparus: `default_is_public` is a dead parameter. `TrackForm.is_public` is `BooleanField(required=False)`, and this helper is only called after `form.is_valid()` passes — at that point `cleaned_data["is_public"]` is always present (`False` when the checkbox is unchecked), so the `.get()` default can never fire. The old code had the same latent `.get(..., True/False)` pattern inline, but this PR promotes it into API surface: the `True`-for-create / `False`-for-update asymmetry now reads like a real behavioral difference for a reviewer to puzzle over, and it isn't one.

Drop the parameter and use `form.cleaned_data["is_public"]` directly — the helper collapses to `_submitted_track_data(request, form)` and one boolean mode disappears from both call sites.

---

src/ludamus/gates/web/django/chronology/panel/views/tracks.py
```diff
@@ -135,76 +121,70 @@ def get(self, _request: PanelRequest, slug: str, track_slug: str) -> HttpRespons
...
+                track_slug=track_slug,
+                data=_submitted_track_data(
+                    request=self.request, form=form, default_is_public=False
                 ),
-            ),
-        )
+            )
+        except NotFoundError:
+            messages.error(self.request, _("Track not found."))
+            return redirect("panel:tracks", slug=slug)
```
hasparus: This method now carries two identical `except NotFoundError` blocks (lines 161–163 and 185–187) where the old code had one. That's the cost of pushing the track read into the service — correct per the authz rule — but the duplication is avoidable. One `try` around both branches collapses them:

```python
try:
    if not form.is_valid():
        edit_context = service.get_edit_context(
            event_pk=current_event.pk, sphere_id=sphere_id, track_slug=track_slug
        )
        context["active_nav"] = "tracks"
        ...
        return TemplateResponse(self.request, "panel/track-edit.html", context)

    service.update(
        event_pk=current_event.pk,
        sphere_id=sphere_id,
        track_slug=track_slug,
        data=_submitted_track_data(request=self.request, form=form),
    )
except NotFoundError:
    messages.error(self.request, _("Track not found."))
    return redirect("panel:tracks", slug=slug)
```

Same behavior, one exception path instead of two, and the "missing track always redirects the same way" invariant becomes structurally obvious instead of copy-pasted.

Minor, related: in this invalid-form branch `get_edit_context` fetches `selected_space_pks` / `selected_manager_pks` (two repo reads) that lines 170–173 immediately discard in favor of the submitted pks. Two wasted queries on an error path is tolerable; just flagging that `TrackEditContextDTO` is serving two use cases (GET initial render vs POST re-render) and only fully fits one.

---

src/ludamus/pacts/tracks.py
```diff
@@ -0,0 +1,47 @@
...
+    ) -> TrackFormContextDTO: ...
+    def get_edit_context(
+        self, *, event_pk: int, sphere_id: int, track_slug: str
+    ) -> TrackEditContextDTO: ...
+    def create(
+        self, *, event_pk: int, sphere_id: int, data: TrackUpdateData
+    ) -> None: ...
+    def update(
+        self, *, event_pk: int, sphere_id: int, track_slug: str, data: TrackUpdateData
+    ) -> None: ...
```
hasparus: The new protocol's input contract lies a little: `create` takes a `TrackUpdateData`. It works because the create payload happens to be the update payload plus `event_pk`, but a brand-new boundary shouldn't inherit a legacy write-TypedDict's name for a concept it doesn't quite mean — and this is the one place the conflict-avoidance argument doesn't apply, because `pacts/tracks.py` is a new file no open PR touches.

Suggestion: define the submission payload here, e.g.

```python
class TrackFormData(TypedDict):
    name: str
    is_public: bool
    space_pks: list[int]
    manager_pks: list[int]
```

and have `create`/`update` take `data: TrackFormData`. The service already rebuilds the write payload in `_scoped`, so translating to the legacy `TrackCreateData` / `TrackUpdateData` at the repo boundary costs nothing extra, the protocol's inputs stop depending on `pacts.legacy`, and when the legacy module is eventually broken up there's one less tendril to cut. (The DTO returns still referencing `pacts.legacy` types are unavoidable for now — this is only about the input side, where the fix is free.)

---

hasparus (commented):
Thermo-nuclear code quality review. Verdict: no genuine defect found — behavior, authz scoping, and transaction ownership all check out against the diff and the panel object-scope rule, and the migration follows the services-migration recipe faithfully (protocol + DTOs in pacts, service owning `atomic()`, view shrunk to glue, unit tests matching the `test_mills.py` mock pattern, existing integration tests — including the foreign-pk-dropping case in `test_track_create_page.py` — as the regression guard). The scoping logic actually got *stronger* in the move: `sorted()` makes the persisted pk lists deterministic where the old set-intersection order was arbitrary.

Three findings, all quality-level, in the line comments:

1. **`_submitted_track_data`'s `default_is_public` is dead code promoted into API surface.** `BooleanField(required=False)` guarantees `cleaned_data["is_public"]` exists after validation, so the `.get()` default never fires — the create/update `True`/`False` asymmetry advertises a behavioral difference that does not exist. Drop the parameter.

2. **`TrackEditPageView.post` duplicates the `except NotFoundError` block twice** where the old code had it once. One `try` wrapping both branches restores a single exception path. Related minor: the invalid-form branch pays for two repo reads (`selected_*_pks`) it immediately discards.

3. **The new protocol reuses the legacy `TrackUpdateData` as the `create` payload.** `pacts/tracks.py` is a new file no open PR touches, so the conflict-avoidance rationale doesn't cover it — define the form payload there (`TrackFormData`) and keep the legacy TypedDicts at the repo boundary, where `_scoped` already rebuilds the payload anyway.

One structural note on the flagged deviation: `pacts/tracks.py` / `mills/tracks.py` as fresh top-level modules is a reasonable dodge around #719/#625/#626, and the pacts/mills mirror holds between them. But GLIMPSE slices pacts/mills by noun, and "tracks" is event-noun territory — this only stays acceptable as a *temporary* shape. Please leave a tracked follow-up (issue or TODO in the migration doc) to fold these into the noun's modules once the conflicting PRs land; "avoid a merge conflict" has a way of becoming permanent drift once the PR that justified it is forgotten. Retaining the pre-existing `pylint: disable=duplicate-code` at net-zero tingle is fine as argued.

None of this blocks: the findings are deletions and renames, not redesigns, and the PR pays off 33 debt units while adding none.

---

src/ludamus/gates/web/django/chronology/panel/views/tracks.py
hasparus: Right — the default could never fire after `is_valid()`. Dropped the parameter and both call sites now read `_submitted_track_data(request=..., form=...)`, with `cleaned_data["is_public"]` used directly (a9956ca2).

---

src/ludamus/gates/web/django/chronology/panel/views/tracks.py
hasparus: Done in 4354e3f2: one `try` now wraps both branches (the invalid-form re-render moved to a `_rerender_edit_form` helper to keep ruff's try-clause size limit happy), and the re-render path calls a new `get_edit_form_context` on the mill that returns `TrackEditFormContextDTO` (track + picker choices) without the two selected-pks reads — `get_edit_context` builds on it for the GET render.

---

src/ludamus/pacts/tracks.py
hasparus: Done in 72294ef0: `pacts/tracks.py` now defines `TrackFormData` and the protocol's `create`/`update` take it; the service translates to the legacy `TrackCreateData`/`TrackUpdateData` only at the repo boundary, so the new pact no longer imports from `pacts.legacy` on the input side. Also added the follow-up TODO from the review body there — fold `pacts/tracks.py` + `mills/tracks.py` into the event noun modules once #719/#625/#626 land.

---

src/ludamus/gates/web/django/chronology/panel/views/tracks.py
```diff
@@ -1,5 +1,4 @@
 # pylint: disable=duplicate-code
-# TODO(fancysnake): Extract common view boilerplate
```
hasparus, line 2: Non-blocking, but this deletion should be reverted: the TODO it removes is still true, and removing it happens to keep a debt counter flat.

The boilerplate was not extracted. At the tip this file still has 6 copies of

```python
context, current_event = self.get_event_context(slug)
if current_event is None:
    return redirect("panel:index")
```

and 3 copies of the `messages.error(..., _("Track not found.")) / redirect("panel:tracks", slug=slug)` pair. `# pylint: disable=duplicate-code` on line 1 is retained, and the PR body itself says "the remaining duplication is cross-file view boilerplate ... which this PR must not touch" — which is an explicit acknowledgement that the condition the TODO flags is unchanged. The three sibling files (`proposals.py`, `session_fields.py`, `time_slots.py`) still carry the identical marker, so dropping it only here makes the debt look narrower than it is.

The counter angle: `tingle.toml`'s `todo-comments` metric matches `#\s*(TODO|FIXME|XXX|HACK)` over `python-src`. This commit (416bd0bc) removes one TODO; 72294ef0 adds the `# TODO(hasparus): Fold this module...` in `pacts/tracks.py`. Net delta at the tip is 0, not the `todo-comments −1` the PR body claims — that number was accurate when written and is now stale. I don't think the offset was deliberate, but it is the shape CLAUDE.md warns about ("Don't game a counter to keep a number flat"), and the honest accounting is: this PR takes on one new TODO and pays off nothing in that metric. Keep the `fancysnake` line, update the body's tingle section, and the new TODO stands on its own justification — which is a good one.

---

hasparus (commented):
## Round 2 — thermo-nuclear review

**Verdict: APPROVE.** Submitted as a comment only because GitHub refuses a formal approval from the PR author's own account — treat this as a green light, not withheld approval. All three round-1 findings are genuinely resolved; I verified each against the code at `4354e3f2` rather than taking the replies at face value. One non-blocking nit inline (a deleted TODO); everything else is clean.

### Verification of the round-1 fixes

**(a) Dead `default_is_public` — resolved.** `TrackForm.is_public` is `BooleanField(required=False, initial=True)` (`forms.py:608`), so `cleaned_data["is_public"]` is always populated after `is_valid()`. `_submitted_track_data` now reads it directly and takes only `request` / `form`. The `True`-for-create / `False`-for-update phantom asymmetry is gone from both call sites.

**(b) Single `try` — behavior-preserving, and I checked the edges.** Old flow read the track *before* validating the form; new flow validates first. Enumerating the four cases:

| form | track | old | new |
|---|---|---|---|
| valid | exists | update → success redirect | `service.update` → success redirect |
| valid | missing | not-found redirect | `read_by_slug` inside `update` raises → not-found redirect |
| invalid | exists | re-render 200 | `_rerender_edit_form` → re-render 200 |
| invalid | missing | not-found redirect | `get_edit_form_context` → `read_by_slug` raises → not-found redirect |

Identical in all four, including the non-obvious invalid-form-plus-missing-track cell, which only works because the re-render path still goes through the event-scoped read. `TemplateResponse` is lazy, so nothing renders inside the `try` and no template exception can be swallowed by the `except NotFoundError`. `update` raises from inside `atomic()`, so the foreign-slug case rolls back with no side effects — asserted by `test_update_foreign_track_slug_raises_without_side_effects`.

`_rerender_edit_form` existing to satisfy ruff's try-clause size limit is a fine outcome — the helper is cohesive on its own terms ("build the edit re-render response"), not a lint dodge.

**(b, cont.) Discarded reads — genuinely eliminated.** `get_edit_form_context` does `read_by_slug` + `get_form_context` and stops; `get_edit_context` layers the two `list_*_pks` calls on top for the GET render only. The POST error path no longer issues them, and `test_get_edit_form_context_skips_assignment_reads` pins it with `assert_not_called`. The three-level dataclass chain (`TrackFormContextDTO` → `TrackEditFormContextDTO` → `TrackEditContextDTO`) earns its keep here: each level is a real read aggregate with a distinct caller, the inheritance de-dupes the shared fields, and all fields are non-default so the MRO ordering is safe. It also matches the existing `PersonalDataFieldFormContextDTO` / `PersonalDataFieldEditContextDTO` precedent in `pacts/submissions.py`.

**(c) `TrackFormData` translation — lossless.** Field-by-field against `pacts/legacy.py:360-372`: `TrackCreateData` is `{event_pk, name, is_public, space_pks, manager_pks}`, `TrackUpdateData` is the same minus `event_pk`. `TrackFormData` carries all four non-`event_pk` fields and `create` supplies `event_pk` from its own argument. Nothing dropped, nothing silently defaulted. The protocol's input side no longer imports from `pacts.legacy`; the legacy TypedDicts are constructed only at the two repo call sites in `mills/tracks.py`. Incidental improvement worth noting: `_scoped` returns `sorted(set(...) & valid)` where the old `_scoped_pks` returned an unordered `list(set & set)` — the write payload is now deterministic.

### Whole-PR judgement at the tip

- **Authz.** Object scoping sits in the service, not the view, per the panel object-scope rule. Track slugs resolve through the event-scoped `read_by_slug`; submitted space/manager pks are intersected with `list_by_event(event_pk)` / `list_managers(sphere_id)` inside `create` and `update`, so a foreign pk is dropped rather than trusted. `delete` scopes by `event_pk` only, which is correct — it takes no request-supplied space/manager ids. Foreign-id tests assert no writes on all three paths.
- **Query counts.** No regression. The old view always fetched choices via `_track_get_choices` on both POST branches; now the valid path fetches them once inside `_scoped` and the invalid path once via `get_form_context`. Same two reads, never both.
- **Layering.** `mills/tracks.py` takes specific repo protocols + `TransactionProtocol` keyword-only, never the UoW; owns `atomic()` for all three writes. Zero `di.uow` references remain in the view (down from 11). `pacts/__init__.py` untouched, imports go to defining modules, `lint-imports` green.
- **Tests.** Mills → unit tests, per the layer rule. The 29 pre-existing tracks integration tests are unmodified and serve as the regression guard, which is the right shape for a pure refactor. No markup assertions added.
- **File size / spaghetti.** The view *shrank*; no new conditionals were bolted onto unrelated flows. Nothing near the 1k-line threshold.
- **CI.** 16/16 green — `test` (3367 passed), `test-postgres`, `checks`, `impeccable`, `tingle`, CodeQL, codecov patch + project.

`list_tracks` being a straight pass-through to the repo is a thin wrapper, but it's the right thin wrapper: it's what gets the last `di.uow` reference out of the view, and inventing logic to justify it would be worse. I'd also mildly prefer `_scoped` to build its valid-pk sets from `get_form_context` rather than re-issuing the same two repo calls in its own words, but that's taste, not a defect.

This is a correct, clean strangler-fig increment. Ship it once the inline TODO nit is addressed (or consciously declined).

---

src/ludamus/gates/web/django/chronology/panel/views/tracks.py
hasparus: Agreed — the deletion wasn't earned. Restored `# TODO(fancysnake): Extract common view boilerplate` verbatim on line 2 in a4c6719a; the boilerplate it flags is untouched (still 6 prologue copies, 3 not-found redirects, `duplicate-code` disable retained), and the actual extraction stays out of scope since it spans the sibling panel views.

Corrected numbers, `tingle check --base origin/main` at the new tip: **no new debt, 27 paid off** — `request-di-uow` −11, `old-subdomain-loc` −18, `todo-comments` **+1**, `type-object` +1. PR body updated with those, plus a note that the new TODO stands on its own justification rather than being offset by a deletion.

---

PR #730, opened 2026-07-29
refactor(panel): migrate cfp, discounts, index views to request.services

hasparus:
Migrates the three small panel view files off the legacy `request.di.uow.*` surface (7 call sites) onto `request.services.*`, per `docs/agents/services-migration.md`. Pure refactor — no template, URL, copy, or behavior change, so no screenshots.

## What moved where

**`views/cfp.py`** (4 uses of `di.uow.proposal_categories` + `PanelService(uow)`)
- New `ProposalCategoriesService` (`mills/proposal_categories.py`) behind `ProposalCategoriesServiceProtocol` (`pacts/proposal_categories.py`), wired as `request.services.proposal_categories`.
- `get_page_context(event_pk)` returns a `ProposalCategoriesPageDTO` aggregate (categories + stats) for the list page; `create` runs in a transaction; `delete_by_slug(event_pk, slug)` resolves the category through the event-scoped `read_by_slug` (foreign slug → `NotFoundError` before any write), then checks `has_proposals` and deletes inside one atomic block, replacing the legacy `PanelService.delete_category` path.

**`views/discounts.py`** (2 uses of `di.uow.facilitators`)
- The facilitator roster folded into the existing `DiscountsService` (`mills/discounts.py`) rather than getting a service of its own: the page is one join of facilitators with their discounts, and `DiscountsExportService` needed the identical join. `list_roster` returns `DiscountRosterEntryDTO` rows, and page and sheet export now share one `_roster()` helper instead of building the same dict twice. The facilitator side reads through `FacilitatorRepositoryProtocol` from `pacts/legacy.py` — imported, not edited, so it stays clear of open PRs #625/#626/#719.
- The event-scoping of request-supplied ids moved from the view helpers into the service: `read_scoped` / `read_scoped_facilitator` raise `NotFoundError` for a foreign id (unit-tested, no side effects), replacing the post-hoc `event_id ==` checks in `_scoped_discount` / `_scoped_facilitator`.

**`views/index.py`** (1 use of `di.uow.events`)
- No new service: the existing `request.services.sphere_panel.list_events(sphere_id)` (`SpherePanelService`, `mills/multiverse.py`) is exactly `events.list_by_sphere`. Creating a parallel service for one read would be the anemic-delegation smell the migration doc warns about.

## Service-slicing choice

One service per domain noun rather than a shared "panel" service: GLIMPSE slices mills by noun, and "panel" is a page group (gates axis), not a noun — a shared panel service would become a catch-all. Proposal categories are an entity of their own, so they get a small pacts/mills module pair (flat files, matching the existing `discounts`/`bookmarks`/`venues` pattern). The facilitator roster is not a separate noun here — it is the discounts join — so it stays inside `DiscountsService`, and the index view reuses an existing sphere-noun service.

## Deviation from the constraints (flagged)

`PanelService.delete_category` in `mills/legacy.py` became dead code once cfp.py stopped calling it, and `lint:vulture` fails on it. I removed just that 13-line method — a file I was asked not to touch because #625/#626/#719 edit it. The alternative was a vulture ignore entry in `pyproject.toml`, which would game a linter to keep dead code. The deletion is surgical (one method, no other lines), but flagging it for the rebase-conflict risk. Also: the new repo protocol can't be declared as a base class on the concrete repository (that would mean editing `links/`), so it's satisfied structurally — mypy still verifies the wiring in `inits/services.py`.

## Tingle

`tingle check --base origin/main`: **no new debt, 29 occurrences paid off** — `request-di-uow` −8, `legacy-loc` −14, `old-subdomain-loc` −8 net, `todo-comments` +1 (the TODO to fold `mills/proposal_categories.py` into `mills/submissions/` once #625/#626 free that package). (Plain `tingle check` in the sandbox compares against a stale local `main` and reports noise from other merged PRs.)

## Test status (sandbox, honest)

Numbers re-measured after merging `main` (merge commit `1a095287`), which brought in the sibling migrations #728 (`panel_time_slots`), #729 (`tracks_panel`) and #732 (`encounters`). `mise` is unavailable in this sandbox (install failed), so the `mise run test:py` / `mise run check` tasks were reproduced piecewise with poetry + the env from `mise.toml`:

- `pytest tests/unit`: **846 passed** (includes 10 new unit tests for the two services, MagicMock-per-repo-protocol pattern, no UoW mocks).
- `pytest tests/integration`: **2586 passed, 9 skipped, 0 failed** — the 11 sandbox-env failures reported earlier no longer reproduce. All 87 integration tests covering the three migrated views pass.
- Lint: ruff, black --check, mypy (`Success: no issues found in 252 source files`), pylint src + tests (10.00/10), import-linter (7 contracts kept), vulture, codespell, djlint, taplo, `tingle check --base origin/main` — all pass.
- Could not run: `lint:hk`, `lint:ast-grep` (mise-managed binaries absent), `lint:impeccable` (fails on an npm git fetch in the sandbox, unrelated to the diff).

---

## Summary by CodeRabbit

* **New Features**
  * Improved proposal category management, including safer creation and deletion behavior.
  * Enhanced discount roster views and exports with facilitator and discount information.
  * Added event-scoped access checks for discounts and facilitators.

* **Bug Fixes**
  * Prevented deletion of proposal categories that contain proposals.
  * Improved handling of unavailable or out-of-scope categories, discounts, and facilitators.
  * Updated panel navigation to select the first available event correctly.

* **Tests**
  * Expanded coverage for discount scoping, roster data, transactions, and proposal category operations.


src/ludamus/mills/facilitator_roster.py
```diff
@@ -0,0 +1,24 @@
...
+        self._facilitators = facilitators
+
+    def list_by_event(self, event_pk: int) -> list[FacilitatorListItemDTO]:
+        return self._facilitators.list_by_event(event_pk)
+
+    def read_scoped(self, *, event_pk: int, facilitator_id: int) -> FacilitatorDTO:
+        facilitator = self._facilitators.read(facilitator_id)
+        if facilitator.event_id != event_pk:
+            raise NotFoundError
+        return facilitator
```
hasparus: This whole slice — `pacts/facilitator_roster.py`, this service, the `ServicesProtocol` property, the `inits` leaf, and a test file — exists to wrap two reads. Half of it (`list_by_event`) is a pure repo pass-through, which the migration doc explicitly warns against: *"Don't promote a bare repository to `request.services`. Repositories stay on `request.di.uow` until they're being wrapped by an orchestrating service that earns its place."* Only `read_scoped` carries any logic.

There's a code-judo move here: the canonical event-scoped facilitator service already exists. `FacilitatorPanelService` (`mills/submissions/facilitator_panel.py`) opens with *"Every method takes the panel's event and scopes to it"* — `read_scoped` is exactly its idiom (cf. its `set_flag`/`assign_organizer` doing `read_by_event_and_slug` first). Adding `list_all(event_id)` + `read_scoped(...)` there (~8 lines plus protocol lines in `pacts/submissions.py`) deletes this module, `pacts/facilitator_roster.py`, both wiring appends, and folds the tests into the existing suite.

It would also stop the facilitator repo contract from growing a third home. It now lives in `pacts/legacy.py` (`FacilitatorRepositoryProtocol`, which `DiscountsExportService` already imports directly), in `pacts/submissions.py` (panel repos bundle), and here — and the new `FacilitatorRosterRepositoryProtocol` copies the `@staticmethod` decorations from the legacy protocol, leaking a `links` implementation detail into a fresh contract that will drift from its siblings.

Not a defect — the code is correct and tested — but it's four new surfaces for two trivial reads when one existing surface would do.

---

src/ludamus/mills/proposal_categories.py
```diff
@@ -0,0 +1,43 @@
...
+    ProposalCategoriesServiceProtocol,
+)
+
+if TYPE_CHECKING:
+    from ludamus.pacts.legacy import ProposalCategoryDTO
+    from ludamus.pacts.proposal_categories import ProposalCategoriesRepositoryProtocol
+    from ludamus.pacts.services import TransactionProtocol
+
+
+class ProposalCategoriesService(ProposalCategoriesServiceProtocol):
```
hasparus, line 19: Placement: proposal-category service logic already has a home — `ProposalCategorySettingsService` in `mills/submissions/proposal_category_settings.py`, protocol in `pacts/submissions.py`. This PR puts the CRUD half of the same entity in a new top-level module pair, so the concept now spans two slicing axes, with two near-identical class names (`ProposalCategoriesService` vs `ProposalCategorySettingsService`) living in different slices. A reader hunting for "who deletes a category" has no reason to guess which of the two modules owns it.

GLIMPSE slices pacts/mills by noun, then verb — `proposal_categories` is an entity of the event noun (legacy `submissions` maps to it), not a noun itself. `mills/submissions/proposal_categories.py` next to `proposal_category_settings.py`, with the protocol + `ProposalCategoriesPageDTO` in `pacts/submissions.py`, keeps the mirror on one axis and makes the eventual submissions→event rename a single move instead of a scavenger hunt.

I'll grant the counterargument: top-level entity modules have precedent here (`discounts.py`, `bookmarks.py`, `session_modal.py`), and avoiding growth in a legacy-named bucket is a real tingle consideration. So this is a judgment call, not a blocker — but same-concept-two-homes is precisely the drift worth containing while the module is 43 lines old and free to move.

The service itself is good: scoped `read_by_slug` before any write, transaction owned here rather than in the view, and wrapping `delete` in the same `try` as the read actually makes the view more robust than the old code (a racing delete now gets "Category not found" instead of a 500).

---

src/ludamus/gates/web/django/chronology/panel/views/discounts.py
```diff
@@ -78,10 +78,11 @@ def _scoped_facilitator(
     *, request: PanelRequest, event_pk: int, facilitator_id: int
 ) -> FacilitatorDTO | None:
     try:
-        facilitator = request.di.uow.facilitators.read(facilitator_id)
+        return request.services.facilitator_roster.read_scoped(
+            event_pk=event_pk, facilitator_id=facilitator_id
+        )
     except NotFoundError:
         return None
```
hasparus, line 84: Moving the `event_id` scoping check out of the view and into `read_scoped` is the right call per the panel object-scope authz rule ("do it in the service, not the view"), and the foreign-event rejection is unit-tested. Good.

It does leave this file asymmetric, though: `_scoped_facilitator` is now a thin NotFoundError→None adapter over a service that scopes, while `_scoped_discount` ten lines up still does the `discount.event_id == event_pk` comparison in the gate over an unscoped `services.discounts.get(pk)`. Two sibling helpers, same job, opposite layers. Worth a small follow-up giving `DiscountsService` the same `read_scoped` treatment so both view helpers collapse to the identical shape (or disappear).

---

hasparus (commented):
Thermo-nuclear code quality review. Verdict: no defects found — behavior is preserved (in one spot improved), authz scoping moved to the right layer, transactions now live in services, and the migration recipe is followed. The pushback below is structural: this PR spends more architecture than the two views need.

**The core issue: four new surfaces for what existing services already own.** Both new module pairs slice by *entity* at the top level while the concepts already have homes in the codebase:

- `facilitator_roster` wraps two reads, one of which is a pure repo pass-through — the migration doc's "don't promote a bare repository to `request.services`" case. `FacilitatorPanelService` is the documented canonical event-scoped facilitator service and would absorb both methods for ~8 lines, deleting two modules, a protocol, wiring, and a test file. (Details on the inline comment.)
- `proposal_categories` lands next to an existing `ProposalCategorySettingsService` that lives on a different axis (`mills/submissions/`), leaving the same entity with two homes and two near-identical service names. (Inline comment; judgment call given the repo's existing top-level entity modules, but worth containing now.)

Neither blocks merge — the code is correct, small, and tested — but if the point of the strangler-fig migration is to end up with *fewer* concepts, this PR adds concepts it didn't have to.

**On the self-flagged deviations, verified against the actual diffs:**

- **(a) Deleting `PanelService.delete_category` from `mills/legacy.py` — right call, keep it.** I checked: after this PR no reference to `delete_category` remains anywhere in src or tests. I also diffed the three open PRs against `mills/legacy.py`: #625 deletes `FacilitatorMergeService` (lines ~507–548), #626 touches `ProposeSessionService` (~387–420), #719 doesn't touch the file at all. None of those hunks overlap or neighbor the ~444–457 region this PR removes, so the rebase cost is a clean auto-merge. Dead code plus legacy-LOC paydown wins.
- **(b) Structural (non-inherited) protocol conformance** — acceptable here. The concrete repos already inherit the fat legacy protocols, and mypy verifies conformance to the new narrow ones at the `inits/services.py` construction sites. Multiple protocol bases on the repos would be noise. (The concern mostly dissolves anyway if the roster methods fold into `FacilitatorPanelService`.)
- **(c) Read-only DTO imports from `pacts/legacy.py`** — fine; pacts→pacts, no new layering edge, and consistent with `mills/discounts.py` doing the same.

Also checked and fine: `index.py` reusing `sphere_panel.list_events` instead of minting another service is exactly the restraint the other two slices could have shown; `delete_by_slug` gained a transaction the old path never had; unit tests match the canonical `test_mills.py` fixture pattern and assert the no-side-effects path for foreign ids.

What the PR does well: the views shrank to glue, the foreign-event facilitator check moved from the gate into a tested service invariant, and the dead-code deletion was verified rather than deferred.

---

src/ludamus/mills/facilitator_roster.py
hasparus: Slice deleted in c807ffdf — folding into `FacilitatorPanelService` would touch `mills/submissions/facilitator_panel.py` + `pacts/submissions.py`, both held by open #625 (the conflict this PR was scoped to avoid), so the two reads moved into `DiscountsService` instead, their only consumer: `list_roster` joins facilitators with their discounts (the join the view was doing) and `read_scoped_facilitator` carries the scoping authz, which also removes the pass-through and the `@staticmethod`-leaking protocol. Folding the facilitator reads into `FacilitatorPanelService` stays the follow-up once #625 lands.

---

src/ludamus/mills/proposal_categories.py
hasparus: Agreed on the destination, but declining the move in this PR: both `mills/submissions/` and `pacts/submissions.py` are held by open #625/#626, and staying out of their way is exactly why this module went top-level next to `discounts.py`/`bookmarks.py`. Added a TODO in the module recording the fold into `mills/submissions/` once those PRs land (7e99f23).

---

src/ludamus/gates/web/django/chronology/panel/views/discounts.py
hasparus: Done in this PR rather than a follow-up (c807ffdf): `mills/discounts.py` was uncontested by the open PRs, so `DiscountsService.read_scoped` now owns the `event_id` check and `_scoped_discount` collapses to the same thin NotFoundError→None adapter as `_scoped_facilitator`, with foreign-id unit tests asserting the rejection writes nothing.

---

src/ludamus/mills/discounts.py
```diff
@@ -10,26 +15,52 @@
...
+            discount.facilitator_id: discount
+            for discount in self._discounts.list_by_event(event_pk)
+        }
+        return [
+            DiscountRosterEntryDTO(
+                facilitator=facilitator,
+                discount=discounts_by_facilitator.get(facilitator.pk),
+            )
+            for facilitator in self._facilitators.list_by_event(event_pk)
+        ]
```
hasparus: `list_roster` is a real join and earns its place — but it's now the second copy of that join in this file. `DiscountsExportService.export_to_sheet` (L108-115) builds the identical `{facilitator_id: discount}` dict over the same two repos and walks `facilitators` the same way. Before this PR the join existed once in mills and once in the view; now it's twice in mills.

The judo move is a module-level pure helper both call:

```python
def _roster(
    *,
    discounts: DiscountRepositoryProtocol,
    facilitators: FacilitatorRepositoryProtocol,
    event_pk: int,
) -> list[DiscountRosterEntryDTO]: ...
```

`list_roster` becomes a one-liner, and `export_to_sheet` collapses to a pure row-formatter over `entries` — its `discounts` dict, its `.get`, and its separate `facilitators` local all disappear, with `return len(entries)` closing it out. Same behavior, one definition of "the roster", and the export rows stop being able to drift from what the page shows.

Non-blocking — a follow-up is fine.

---

hasparus (commented):
## Round 2 — verdict: APPROVE

_(Submitted as a comment only because GitHub refuses a formal approval from the PR author's own account. Treat this as an approving review — no blocking findings.)_

All three round-1 findings verified against the code at `7e99f23`, not against the replies. They hold up.

**(a) Facilitator roster slice — genuinely resolved, not renamed.** `mills/facilitator_roster.py`, `pacts/facilitator_roster.py`, the `ServicesProtocol` property, the `inits` leaf and the test file are all gone from the tree. What replaced them is not the same code under a new name: the pass-throughs are *deleted* — `DiscountsService.list_by_event` and `.get` no longer exist on the service surface (`pacts/discounts.py`), and no caller reaches for them. `list_roster` is orchestration by the honest definition: it owns the `{facilitator_id: discount}` join the view was hand-rolling, and `_discounts_context` drops from two repo calls plus a dict comprehension to a single loop over typed entries. The `@staticmethod`-copying protocol went with the slice.

**(b) `mills/proposal_categories.py` placement — defensible decline.** I checked the constraint rather than taking it on faith: #625 changes `src/ludamus/pacts/submissions.py` (among ~40 files), which is where the protocol and `ProposalCategoriesPageDTO` would have to land. That's a real conflict against a large open PR, and the destination is recorded in-module in the house `# TODO(owner):` style. A 46-line module with a TODO is cheap to move later; a merge conflict in a 40-file PR is not. Correct call — I'd have made the same one.

Worth noting the service is a genuine improvement over what it replaced, not just a relocation: `has_proposals` + `delete` are now inside one `atomic()` (they weren't in `PanelService.delete_category`), the event-scoped `read_by_slug` gates the write, and `PanelService` shed a method rather than gaining one.

**(c) `_scoped_discount` authz — enforced in the service, tested.** `DiscountsService.read_scoped` raises `NotFoundError` on an event mismatch, and both view helpers are now byte-for-byte the same shape: a `try`/`except NotFoundError` → `None` adapter with no comparison logic left in the gate. The asymmetry round 1 flagged is gone. `test_read_scoped_rejects_foreign_event_discount` asserts the rejection *and* that `repo.updated` / `repo.soft_deleted` stay empty, which is the no-side-effects half the panel authz rule actually asks for; `test_delete_by_slug_propagates_not_found_without_side_effects` does the same for categories. The pre-existing cross-event integration tests in `tests/integration/web/panel/test_discounts_page.py` still cover it end to end.

### Judging the tip as a whole

Read as its own PR rather than as a delta, this is a clean strangler-fig increment. Four views move off `request.di.uow` onto `request.services`, the `di.uow` surface shrinks rather than grows, no new legacy dependency appears, `DiscountsService(DiscountsServiceProtocol)` and the per-slice repo protocol in `pacts/proposal_categories.py` both follow the conventions already set by `pacts/discounts.py`, and the DTO types line up correctly (`FacilitatorListItemDTO` for the list read, `FacilitatorDTO` for the single read). No file goes near the 1k line boundary, no new branching lands in an existing flow, no casts or optionality added to paper over a boundary. CI is green including `tingle` and `impeccable`.

One non-blocking note inline: `list_roster` duplicates the facilitators × discounts join that `export_to_sheet` already does ten lines below it. Extracting the shared helper would make the export a pure row-formatter. Follow-up, not a merge blocker.

Minor: the facilitator-reads → `FacilitatorPanelService` follow-up is recorded only in the review thread above, while the analogous proposal-categories fold got an in-module TODO. Review threads evaporate; a one-line TODO next to `read_scoped_facilitator` would keep the two follow-ups discoverable by the same mechanism. Entirely optional.

Round-1 findings resolved, no new blocking issues. Ship it.

---

src/ludamus/mills/discounts.py
hasparus: Done in be42eb0c. Checked the two joins first — same repos, same unfiltered `list_by_event(event_pk)`, same facilitator-order walk, same `.get` for a missing discount, and `len(facilitators) == len(entries)` — so unifying them is behavior-preserving, and both existing unit suites pass unchanged as the regression guard.

`_roster(*, discounts, facilitators, event_pk)` is now a module-level private helper; `list_roster` is a one-liner over it and `export_to_sheet` is a pure row formatter over `entries` — its `discounts` dict, its `.get`, and its separate `facilitators` local are gone, closing with `return len(entries)`.</body>
</invoke>


---

tests/unit/test_proposal_categories_service.py
```diff
@@ -0,0 +1,93 @@
...
+        description="",
+        durations=[],
+        end_time=None,
+        max_participants_limit=0,
+        min_participants_limit=0,
+        name=name,
+        pk=pk,
+        slug=slug,
+        start_time=None,
+    )
```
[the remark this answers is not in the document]
hasparus: Valid — fixed in 920c2b3e: `_category` now takes `*, pk=1, name="Talk", slug="talk"`. All five call sites already passed keywords, so the separator was the whole change; defaults and the DTO construction are untouched, and the 15 unit tests in `tests/unit/test_proposal_categories_service.py` + `tests/unit/test_discounts_service.py` still pass.

---

tests/unit/test_discounts_service.py
```diff
@@ -58,6 +92,29 @@ def soft_delete(self, pk):
         self.soft_deleted.append(pk)
 
 
+class FakeFacilitators:
+    def __init__(self, *, list_items=(), facilitators=()):
+        self._list_items = list(list_items)
+        self._facilitators = {f.pk: f for f in facilitators}
+
+    def list_by_event(self, _event_id):
+        return list(self._list_items)
```
hasparus, line 101: Nit, non-blocking — the only thing I found worth writing down.

`FakeFacilitators.list_by_event` discards the event id and returns everything, while its sibling `FakeRepo.list_by_event` (L74-75) actually filters on `d.event_id == event_pk`. So `_roster` is only half-guarded: `test_list_roster_pairs_facilitators_with_their_discounts` would still pass if the helper passed the wrong id — or no id — to the facilitator side.

There's no bug today (`_roster` passes the same `event_pk` to both repos, and the real repo scopes), so this is a fidelity gap, not a defect. Storing `list_items` keyed by event and filtering, or just asserting the received id in the roster test, closes it and makes the two fakes symmetric.

---

hasparus (commented):
## Verdict: APPROVE — safe to merge

(Submitted as a comment only because GitHub refuses `APPROVE` on a self-authored PR. The verdict is approval.)

Final round. I verified the merge rather than trusting it, and it holds up. One non-blocking test-fidelity nit inline; nothing that should stop this landing.

### Merge integrity (`1a095287`) — clean, nothing lost

The strongest check available: `git diff origin/main <tip>` for the two conflict-prone files is **additions-only**.

- `pacts/services.py`: +1 import, +1 property (`proposal_categories`). Zero deletions. Every sibling leaf — `panel_time_slots` (#728), `tracks_panel` (#729), `encounters` (#732), `print_materials` / `printables_reminder` (#726), `discounts_export` — is present and untouched.
- `inits/services.py`: +1 import, +1 `cached_property`, plus the intended `DiscountsService(...)` constructor change to keyword args. Zero deletions.

Nothing from `main` was dropped, duplicated, or reverted. `cfe1011f` is an ancestor of the tip.

### Semantic integrity against merged `main`

- `_roster` still type-checks against main's untouched contracts: `FacilitatorRepositoryProtocol.list_by_event -> list[FacilitatorListItemDTO]` matches `DiscountRosterEntryDTO.facilitator`, and `read -> FacilitatorDTO` (which carries `event_id`; the list item does not) is the right DTO for `read_scoped_facilitator`. Main changed neither `pacts/legacy.py`, `pacts/discounts.py`, nor `links/db/django/repositories/` since the merge base.
- `ProposalCategoriesRepositoryProtocol` mirrors `ProposalCategoryRepository` exactly, including which members are `@staticmethod` and which is not (`create` is the instance method on both). No circular import: `pacts/legacy.py` does not import `pacts/discounts.py`.
- **`PanelService.delete_category`: zero callers** — confirmed by grepping the whole merged tree, not just the diff. `PanelService` itself stays alive for `delete_session_field` (`views/session_fields.py:294`), and no stale vulture entry references the removed method.
- `920c2b3e`: all five `_category(...)` call sites already passed keywords. Nothing broke.

### Base has moved — merges clean anyway

`main` advanced past this branch while I was reviewing: **#733 (event-settings migration), #735, #740** — 13 commits. I ran the merge in memory (`git merge-tree`): **no conflicts**, and the resulting tree contains *both* `event_settings` and `proposal_categories` in `pacts/services.py` and `inits/services.py`, with `delete_category` still gone. The stale base is harmless; no rebase gymnastics needed.

Worth noting: #733 landed `mills/event_settings.py` as a **top-level module with a documented fold-back**, i.e. exactly the pattern this PR was asked to justify in round 1. Combined with #625 and #626 both still being open (I checked — 82 and 28 files respectively, still held), the placement decision and its TODO are now corroborated by main's own precedent. Correctly declined; I'm not reopening it.

### Code quality

The round-2 `_roster` extraction is the right shape — one definition of "the roster", `list_roster` a one-liner over it, `export_to_sheet` reduced to a pure row formatter closing with `return len(entries)`. Page and export can no longer drift.

Two things I'll credit explicitly because they're easy to miss:

- Scoping lives in the service, not the gate. `read_scoped` / `read_scoped_facilitator` / `delete_by_slug` all reject a foreign id before any write, and each has a unit test asserting no side effects. That satisfies the panel object-scope rule properly rather than nominally.
- `index.py` reuses the existing `sphere_panel.list_events` instead of minting a parallel service for one read. Resisting a new service is the right call here.

No file approaches 1k lines (`mills/proposal_categories.py` is 46, `mills/discounts.py` 143). No new branching in shared paths. No casts, no new optionality, no wrappers.

### PR body accuracy

Two small drifts, prose only:

1. "Pure refactor — no template, URL, copy, or behavior change" isn't literally true. `delete_by_slug` wraps `has_proposals` + `delete` in one atomic block where the legacy `PanelService.delete_category` had no transaction, and the view's `try` now covers the delete too, so a racing delete yields "Category not found." instead of a 500. Both are improvements, and you documented them in a review thread — the body just overstates the "no behavior change" claim.
2. The test counts (846 / 2586) were measured at `1a095287`. They predate #733 and #740 landing on main. Not wrong, just stale.

Neither needs a new commit.

### CI

11 of 12 checks green at `920c2b3e` — `tingle`, `checks`, `test-postgres`, `impeccable`, CodeQL, varlock, frontend-analysis all passing. Only the long `test` job was still running when I looked; merge on green.

Approving. The merge is clean, the scoping is sound, and the resolved findings stayed resolved.

---

PR #732, opened 2026-07-30
Migrate notice_board views to request.services

hasparus:
Migrates `src/ludamus/gates/web/django/notice_board/views.py` off the legacy `request.di.uow.*` surface onto `request.services.encounters`, following `docs/agents/services-migration.md`. Pure refactor — no visual change, so screenshots are not applicable.

## What moved where

- **pacts:** new `src/ludamus/pacts/encounter.py` — `EncounterServiceProtocol`, `EncounterDetailContextDTO` (bundles encounter + creator + attendee `UserDTO`s + RSVP stats), and `RSVPOutcome` enum. `ServicesProtocol` in `pacts/services.py` gains the `encounters` property.
- **mills:** new `src/ludamus/mills/encounter.py` — `EncounterService`, constructor takes `TransactionProtocol` + the specific `EncounterRepositoryProtocol`, `EncounterRSVPRepositoryProtocol`, and `UserRepositoryProtocol` (keyword-only). It absorbs `build_index`/`build_detail` from the legacy UoW-based service, plus the create/update/delete, RSVP, and share-code reads the views used to do inline. Writes run inside `transaction.atomic()`; views no longer touch transactions.
- **inits:** `encounters` + `encounter_rsvps` repo leaves in `inits/repositories.py`; `encounters` service leaf in `inits/services.py`.
- **gates:** views shrink to form parsing, flash messages, and redirects. The RSVP flow&#39;s full/throttled/duplicate branching now lives in the service and comes back as `RSVPOutcome`; the view only maps outcomes to messages. `notice_board/helpers.py` (`build_attendee_list`) is deleted — the service returns attendee `UserDTO`s and the view maps them to `UserInfo`.

## Authz

The creator-ownership check for edit/delete moved from the view into the service (`read_owned`/`update_owned`/`delete_owned` raise `NotFoundError` for a foreign pk, before any write). Unit tests cover that a foreign id is rejected with no side effects. Detail/QR/ICS remain intentionally public by share code, unchanged from before.

## Tingle delta

`tingle check --base origin/main`: **no new debt, 47 paid off** — `request-di-uow` −11, `old-subdomain-loc` −36 net (new code lives in `encounter.py` modules per the noun-migration naming).

## Test &amp; check results (run in the CC-web sandbox via the poetry venv, matching mise task definitions)

- `pytest -n auto tests/integration tests/unit`: **3378 passed, 9 skipped** (same skips as main)
- new unit tests: `tests/unit/test_encounter_service.py`, 19 passed — mocks the specific repo protocols + transaction, no UoW mocks
- `ruff check --no-fix scripts src tests`: pass
- `black --check src tests`: pass
- `mypy src`: pass (247 files)
- `pylint src` and `pylint --disable=duplicate-code tests`: 10.00/10, **both exit 0** — re-verified at the current tip. Correction: this line previously claimed a pass on the strength of the 10.00/10 score alone. That score prints even when pylint fails, and at the time it was written `pylint src` was exiting 8 on `R0801` (duplicate-code) — the `mills/encounter.py` / `mills/legacy.py` overlap that 216d7e10 removed. The claim is accurate now, but it was not when first posted.
- `lint-imports --config=pyproject.toml`: 7 contracts kept, 0 broken
- `vulture` (per `lint:vulture` task): pass
- `codespell scripts src tests/unit tests/integration`: pass
- `tingle check --base origin/main`: pass (see above)

Not run (sandbox limitations, honestly reported): `lint:ast-grep` and `lint:hk` (binaries unavailable in this sandbox), `lint:impeccable` (fails identically on a clean `origin/main` checkout with an npm/GitFetcher sandbox error), `lint:djlint`/`taplo` (no template or TOML changes), e2e.

Known follow-up: the UoW-based `EncounterService` in `mills/legacy.py` is now unused, but `mills/legacy.py` is held by other open PRs and the migration doc says not to slim the legacy surface piecewise — it can be removed when the strangler-fig migration completes.

---


src/ludamus/mills/encounter.py
```diff
@@ -0,0 +1,158 @@
...
+        rsvp_count = self._rsvps.count_by_encounter(encounter.pk)
+        if encounter.max_participants > 0 and rsvp_count >= encounter.max_participants:
+            return RSVPOutcome.FULL
+        if self._rsvps.recent_rsvp_exists(ip_address):
+            return RSVPOutcome.THROTTLED
+        if self._rsvps.user_has_rsvpd(encounter.pk, user_id):
+            return RSVPOutcome.ALREADY_SIGNED_UP
+        with self._transaction.atomic():
+            self._rsvps.create(encounter.pk, ip_address, user_id)
+        return RSVPOutcome.CREATED
```
hasparus: The `transaction.atomic()` here wraps only the single `create` INSERT — which is already atomic on its own — while all three invariant checks (capacity, throttle, duplicate) run *outside* it. The wrapper implies a protection the method doesn't have: two concurrent POSTs can both pass the capacity check and overshoot `max_participants`. That TOCTOU is pre-existing (the old view had it too), so it's not a regression — but the whole point of this migration is that the service now owns the transactional boundary, and this was the natural moment to either widen it or stop implying it.

Concretely: either drop the no-op `atomic()` wrappers around single writes (here, in `create`, and in `cancel_rsvp` — `update_owned`/`delete_owned` are the only ones where the block actually groups a check with a write), or make the capacity check real in a follow-up (a `select_for_update` on the encounter row inside the transaction, or a DB-level guard). As written, the code reads as transactionally safe and isn't.

---

src/ludamus/mills/encounter.py
```diff
@@ -0,0 +1,158 @@
...
+    from ludamus.pacts.legacy import (
+        EncounterData,
+        EncounterDTO,
+        EncounterRepositoryProtocol,
+        EncounterRSVPRepositoryProtocol,
+    )
+    from ludamus.pacts.services import TransactionProtocol
+
+
+class EncounterService(EncounterServiceProtocol):
```
hasparus, line 29: This class is a near-verbatim duplicate of `mills/legacy.py:148` — `build_index` is character-for-character identical modulo `self._uow.encounters` → `self._encounters`. I verified the legacy `EncounterService` (and its `EncounterDetailResult` DTO) now has zero callers at this PR's head, so keeping it isn't a correctness problem, and I accept the stated constraint that `legacy.py` is held by #625/#626.

But there's a live footgun while both exist: `mills/__init__.py` does `from ludamus.mills.legacy import *` with no `__all__`, so `from ludamus.mills import EncounterService` still resolves — silently — to the dead UoW-based class, while the real one lives at `ludamus.mills.encounter.EncounterService`. Anyone (or any agent) reaching for the obvious import path gets the corpse. Please file a tracked follow-up to delete `EncounterService` + `EncounterDetailResult` from `legacy.py` the moment #625/#626 land, and link it from this PR so the duplication has an expiry date rather than becoming permanent.

---

src/ludamus/gates/web/django/notice_board/views.py
```diff
@@ -301,8 +296,9 @@ def get(self, request: RootRequest, share_code: str) -> TemplateResponse:
         current_user_id = request.user.pk if request.user.is_authenticated else None
```
hasparus: This is exactly `request.context.current_user_id` — the middleware sets it from `request.user.pk` for authenticated requests and leaves it `None` otherwise, which is precisely what this conditional recomputes. The rest of the PR standardized on `request.context.current_user_id`; finishing the job here deletes the branch:

```python
current_user_id = request.context.current_user_id
```

---

src/ludamus/gates/web/django/notice_board/views.py
```diff
@@ -263,7 +258,9 @@ def post(self, request: AuthenticatedRootRequest, pk: int) -> HttpResponse:
         if header is not None:
             data["header_image"] = header
 
-        uow.encounters.update(pk, data)
+        request.services.encounters.update_owned(
+            pk=pk, user_id=request.context.current_user_id, data=data
+        )
```
hasparus: Minor: a successful edit POST now reads the same encounter row three times — `_get_encounter` (`read_owned`), the re-check inside `update_owned`, and `Encounter.objects.get` inside the repo's `update`. The repos are all `@staticmethod` with no identity map, so these are real queries. The in-transaction re-check is in the right place (the service is the authz boundary); the redundant one is the view's pre-read. If `update_owned` returned the updated `EncounterDTO`, the happy path could skip `_get_encounter` and fetch only on the form-invalid branch. Not blocking — noting it since the double ownership check can read as confusion about which one is load-bearing.

---

hasparus (commented):
Thermo-nuclear code quality review. Verdict: **COMMENT** — no genuine defect found; the findings below are quality observations, none blocking.

I verified the migration against the old code path rather than taking the description on faith:

- **RSVP behavior preserved** — the `RSVPOutcome` mapping keeps the exact message strings, the exact levels (error/error/warning/success), and the service checks full → throttled → already-signed-up in the same order the old view did. Confirmed.
- **`current_user_id` swap is sound** — the middleware builds `AuthenticatedRequestContext` with `current_user_id=request.user.pk`, and `AuthenticatedRootRequest` types it as `int` (vs the old `int | None` dance). This is the same value with a better type. Confirmed.
- **Legacy `EncounterService` is dead** — zero callers at PR head for it, `EncounterDetailResult`, and the deleted `build_attendee_list`. But it's still importable via the `mills/__init__.py` star-import facade under the same name as the new class — see the inline comment for the follow-up ask.
- **Ownership scoping is right** — foreign-pk edit/delete now 404 via `NotFoundError` raised in the service, with no-side-effect unit tests for both, matching the object-scope authz rule. Moving the check from the view into `read_owned` is a genuine improvement.

Inline comments, in priority order:

1. `mills/encounter.py` `rsvp()` — decorative `atomic()` around a single INSERT while the capacity/throttle/duplicate checks sit outside the transaction; the code implies a safety it doesn't provide (pre-existing race, but the service owning the boundary was the moment to address or stop implying it).
2. `mills/encounter.py` — `build_index`/`build_detail` duplicate the now-dead `mills/legacy.py` service almost verbatim, and `from ludamus.mills import EncounterService` silently resolves to the dead one. Needs a tracked expiry, not just an intention.
3. `views.py` detail — `request.user.pk if ... else None` is a hand-rolled `request.context.current_user_id`; one branch can go.
4. `views.py` edit POST — triple read of the same row; consider `update_owned` returning the DTO.

One follow-up worth a ticket rather than an inline nit: `build_index` and `build_detail` carry over the legacy N+1s (`count_by_encounter` per encounter, `read_by_id` per attendee/creator). Faithful migration, not a regression, but now that the logic lives in one testable service it's cheap to fix with batch repo methods.

What the PR does well: it deletes `helpers.py` outright by moving attendee assembly behind a clean DTO (gravatar/`UserInfo` presentation correctly stays in the gate), and the wiring, protocol, and unit tests match the canonical `CFPPersonalDataFieldService` recipe exactly.

---

src/ludamus/mills/encounter.py
hasparus: Fixed in 216d7e10: `rsvp()` now runs the encounter read, capacity count, throttle, and duplicate checks inside the same `atomic()` block as the insert, with a comment stating what the block still does not buy — without a `select_for_update` row lock on the encounter (which needs a repo method in the frozen `pacts/legacy.py`), concurrent signups can still overshoot `max_participants`. The no-op wrappers around the single writes in `create()` and `cancel_rsvp()` are gone; `update_owned`/`delete_owned` keep theirs since they group a check with a write.

---

src/ludamus/mills/encounter.py
hasparus: Recorded in 216d7e10 as a TODO on the new class: it names both symbols to delete from `mills/legacy.py` once #625/#626 land, and spells out the footgun — `from ludamus.mills import EncounterService` resolves to the dead UoW class through the `mills/__init__.py` star-import. I left `mills/__init__.py` alone: it's the pre-existing legacy facade the glimpse skill allows, and dropping the star-import would change resolution for every other legacy symbol, which is the fold-back's job rather than this PR's. I also reshaped `build_index` around a single item-builder helper and made `is_full`/`spots_remaining` derived properties on the detail DTO, which removes the character-for-character overlap — that duplication was failing `lint:pylint` (R0801) on this branch, and the `checks` job is green locally now.

---

src/ludamus/gates/web/django/notice_board/views.py
hasparus: Done in 216d7e10 — the conditional is gone, the view now reads `current_user_id = request.context.current_user_id`.

---

src/ludamus/gates/web/django/notice_board/views.py
hasparus: Taken in 216d7e10: `update_owned` now returns the updated `EncounterDTO` (read back inside the same transaction, after the write), so the happy path drops the view's pre-read — `_get_encounter` only runs on the form-invalid branch, where the template needs the encounter anyway. That leaves the in-transaction ownership check as the only load-bearing one; the view maps its `NotFoundError` to 404. No repo methods were added, so nothing in `links/db/django/repositories/notice_board.py` or `pacts/legacy.py` moved.

---

tests/unit/test_encounter_service.py
```diff
@@ -0,0 +1,355 @@
...
+
+        outcome = service.rsvp(
+            share_code=encounter.share_code,
+            user_id=OTHER_USER_ID,
+            ip_address="10.0.0.1",
+        )
+
+        assert outcome == RSVPOutcome.CREATED
+        assert rsvps.create.call_args == call(encounter.pk, "10.0.0.1", OTHER_USER_ID)
+        transaction.atomic.assert_called_once_with()
```
hasparus: Nit, non-blocking: `transaction.atomic.assert_called_once_with()` proves the block was entered, but not that the checks now run *inside* it — which is the entire content of 216d7e10. `transaction`, `encounters`, and `rsvps` are independent `MagicMock()`s with no shared parent, so nothing records call ordering between `transaction.atomic()` and `encounters.read_by_share_code` / `rsvps.count_by_encounter` / `rsvps.recent_rsvp_exists`. Move all four reads back outside the `with` and every test in this class still passes.

If you want the fix to stay fixed, attach the collaborators to a common parent (`parent = MagicMock(); parent.attach_mock(transaction, "transaction"); parent.attach_mock(rsvps, "rsvps")`, …) and assert `parent.mock_calls` puts `transaction.atomic().__enter__` ahead of the reads. The behaviour is right today; it just isn't pinned against the next refactor.

---

src/ludamus/gates/web/django/notice_board/views.py
```diff
@@ -263,7 +257,12 @@ def post(self, request: AuthenticatedRootRequest, pk: int) -> HttpResponse:
         if header is not None:
             data["header_image"] = header
 
-        uow.encounters.update(pk, data)
+        try:
+            encounter = request.services.encounters.update_owned(
+                pk=pk, user_id=request.context.current_user_id, data=data
+            )
+        except NotFoundError as exc:
```
hasparus, line 264: Nit, non-blocking: this `NotFoundError` → 404 is the authz branch for the edit POST, and 216d7e10 is what created it — before this commit the 404 came from `_get_encounter` above the form validation, so the mapping is new code on the write path.

`tests/integration/web/notice_board/test_edit_page.py` exercises a foreign/unknown pk only on GET (`test_not_creator`, `test_not_found`); all six POSTs in that file target an encounter built with `EncounterFactory(creator=user, …)`. So these two lines are never hit at the gates layer.

`test_update_owned_foreign_encounter_has_no_side_effects` covers the service side and the behaviour is correct — I traced it: `resolve_uploaded_file_field` is pure, so a foreign pk with a valid form still reaches `update_owned`'s in-transaction `read_owned` before any write. But CLAUDE.md's panel object-scope rule asks for a test that a foreign id 404s *without side effects*, and the view-level mapping is now the untested half. One `authenticated_client.post(self._url(foreign.pk), valid_payload)` asserting 404 plus an unchanged row would close it.

---

hasparus (commented):
**Round 2 — verdict: APPROVE.** All four round-1 findings verified as genuinely fixed against the code at `216d7e10`, not against the replies. Two non-blocking nits are attached as line comments; neither holds up the merge.

_(Submitted as a COMMENT only because GitHub refuses the APPROVE event on a self-authored PR — "Can not approve your own pull request". The verdict is approval.)_

## Verification of the round-1 fixes

### (a) RSVP transaction boundary — resolved

`mills/encounter.py:133-146`: the share-code read, capacity count, throttle check, and duplicate check now all sit inside the same `atomic()` block as the insert. Behaviour is unchanged as claimed — all four repo calls are plain reads, so re-running them inside the block can't produce a different outcome, and the early `return`s inside the `with` commit an empty transaction rather than rolling anything back. The comment naming the residual race is accurate: without a row lock nothing serializes two concurrent signups.

Dropping the two wrappers is safe. `create()` is one `Encounter.objects.create(**data)`. `cancel_rsvp()` is a read followed by `delete_by_user` with no invariant tying them — if the encounter vanishes in between, the delete matches zero rows, which is the same outcome as it already being gone. `update_owned`/`delete_owned` correctly keep theirs, since those group the ownership check with the write.

Worth recording for future readers of this file: `edges/settings.py:201` sets `ATOMIC_REQUESTS: True` on the Postgres config, so in production the whole request is already one transaction and these `atomic()` blocks are savepoints inside it. That doesn't make the change wrong — the sqlite path has no such wrapper, and a service should own its own boundary regardless — but it does mean the round-1 concern was less severe than it read, and the honest half of the new comment (no `select_for_update`) is the half that actually matters.

### (b) Legacy duplication — resolved, and the de-dup is behaviour-preserving

I checked the `build_index` fold against `mills/legacy.py:189-231` case by case:

- `mine` comes from `list_upcoming_by_creator`, which filters `creator_id=creator_id`, so the derived `is_mine` is always `True` and `organizer_name` always `""` — matching the old hardcoded values.
- `rsvpd` comes from `list_upcoming_rsvpd`, which `.exclude(creator_id=user_id)`, so the derived `is_mine` is always `False` — matching the old hardcoded `False`. The `my_ids` filter is belt-and-braces given that exclude, but `test_build_index_merges_mine_and_rsvpd_sorted` deliberately feeds `mine` back through `list_upcoming_rsvpd` to pin the dedup, so it's covered contract rather than vestigial code.
- `past` derives `is_mine` exactly as the old comprehension did.
- Ordering survives: `[*mine, *rsvpd]` preserves the old append order and `list.sort` is stable, so ties resolve identically. Query counts are unchanged — one `count_by_encounter` per item, `_resolve_creator_name` only for non-mine.

The derived properties can't drift, and they agree with the old computation on both branches. For `limit > 0`, `spots_remaining == 0` ⟺ `limit - rsvp_count <= 0` ⟺ `rsvp_count >= limit`, which is the old `is_full`. For `limit <= 0`, `spots_remaining` is `None`, `None == 0` is `False`, and the old expression short-circuited to `False` too. The only construction site is `build_detail`, and `EncounterDetailPageView` reads both properties in Python (`views.py:326-327`) rather than serialising the model, so nothing loses them.

Declining to touch `mills/__init__.py` is the right call for this PR. The TODO records the footgun and names both symbols, which is what the round-1 comment asked for.

### (c) `request.context.current_user_id` — done

`views.py:298`. The conditional is gone.

### (d) Triple read — resolved, no read-your-write or staleness problem

`update_owned` reads back inside the same `atomic()` after the write, so the returned DTO reflects the update under any isolation level. The reordering also doesn't move a side effect ahead of the authz check: `resolve_uploaded_file_field` (`pacts/legacy.py:68-74`) is pure — it inspects the `ClearableFileInput` tri-state and returns, no storage write — so a foreign pk with a valid form still 404s from `update_owned`'s in-transaction `read_owned` before anything is written. Form-invalid keeps its own `_get_encounter` because the template needs the row anyway.

## CI claim

Verified, exactly as stated. Run [30502246099](https://github.com/zagrajmy/ludamus/actions/runs/30502246099) on `11f8a1db` failed the `checks` job with `Process completed with exit code 8` from `lint:pylint src`, reporting three `R0801` hits: `ludamus.mills.encounter:[46:54]`↔`ludamus.mills.legacy:[197:205]`, `ludamus.mills.encounter:[55:63]`↔`ludamus.mills.legacy:[206:214]`, and `ludamus.pacts.encounter:[15:22]`↔`ludamus.pacts.legacy:[545:552]`. Pylint still printed `10.00/10` on that run, which is presumably why the PR body's check list claims `pylint src` passed — the score line is not the exit code. Worth correcting in the description so the next reader doesn't trust it.

At `216d7e10` the `checks` job is green, so all three are gone.

**Current CI:** 11 of 12 checks green — `checks`, `tingle`, `impeccable`, `test-postgres`, `frontend-analysis`, `varlock-scan`, CodeQL ×3, staging dispatch. `test` is still in progress, but its `mise run test:py:cov` step has already completed successfully; only the Playwright e2e step is outstanding.

## Whole-PR read at the tip

The shape is right for a strangler-fig step. `views.py` shrinks 437 → 429 lines and the views are down to form parsing, message mapping, and redirects. The RSVP full/throttled/duplicate branching moved into the service and comes back as an `RSVPOutcome` the view pattern-matches on — a real reduction in the number of concepts a reader holds, not a relocation of them. Repo protocols are injected individually rather than through the UoW, ownership checks sit in the service where the panel object-scope rule wants them, and the DTO boundary is Pydantic all the way out to the template context. No file-size problem, no new ad-hoc branching, no feature logic leaking into shared paths, no casts or optionality papering over an unclear invariant. `tingle` is green and the PR pays down 47.

Ship it.

---

tests/unit/test_encounter_service.py
hasparus: You're right, and the "move all four reads back out and everything still passes" check confirmed it — I reproduced that before fixing.

Fixed in 831efa9f. The four collaborator fixtures now hang off one parent `MagicMock` (`collaborators`), so `transaction`, `encounters`, `rsvps` and `users` are children of a single mock and every call lands on one ordered `mock_calls` list. `test_rsvp_creates_signup_in_transaction` asserts that list as an exact sequence:

```
transaction.atomic
transaction.atomic().__enter__
encounters.read_by_share_code
rsvps.count_by_encounter
rsvps.recent_rsvp_exists
rsvps.user_has_rsvpd
rsvps.create
transaction.atomic().__exit__
```

I asserted the call *names* rather than full `call` objects because building the expected list with `call.transaction.atomic().__enter__()` trips ruff's `unnecessary-dunder-call`, and suppressing it wasn't on the table. The argument assertion that the old test carried is kept alongside as `rsvps.create.call_args == call(...)`, so args are still pinned; the reads' arguments stay pinned by the neighbouring throttle and duplicate tests.

Verified the way you'd want: I temporarily moved the encounter read, capacity count, throttle and duplicate checks back outside the `with` in `mills/encounter.py`, ran the file, and got

```
AssertionError: assert ['encounters....enter__', ...] == ['transaction...s_rsvpd', ...]
At index 0 diff: 'encounters.read_by_share_code' != 'transaction.atomic'
1 failed, 18 passed
```

then reverted the service — `git status` shows the commit touches only the two test files, no production change.

---

src/ludamus/gates/web/django/notice_board/views.py
hasparus: Agreed — closed in 831efa9f with `test_not_creator_post_has_no_side_effects` in `tests/integration/web/notice_board/test_edit_page.py`, right next to the GET-only `test_not_creator`.

It snapshots the row as an `EncounterDTO`, POSTs a valid payload (title, start_time, max_participants) to another user's encounter via the `encounter` fixture, asserts `assert_response_404(response)`, then re-reads the row and compares it to the snapshot — so both halves of the panel object-scope rule are covered: 404 and nothing written. Comparing whole DTOs rather than individual fields means a write to any field the form can reach fails the test, not just the three in the payload. No markup assertions; that stays with e2e.

Confirmed it lands on the branch you flagged rather than the older `_get_encounter` 404 — running just this test with `--cov` on the view module reports lines 266-267 (the `messages.success` / `redirect` happy path) as missed while the `except NotFoundError` / `raise Http404` pair is covered, which is only possible if the request reached `update_owned` and came back through the mapping.

`pytest -n auto tests/integration tests/unit` is 3378 passed, 9 skipped; ruff, black, mypy, both pylint invocations (exit 0), import-linter and vulture are clean, and `tingle check --base origin/main` reports no new debt.

---

PR #733, opened 2026-07-30
refactor(event-settings): migrate panel view to request.services

hasparus:
Migrates `gates/web/django/chronology/panel/views/event_settings.py` (all 16 `request.di.uow` call sites) to the service layer, per [docs/agents/services-migration.md](docs/agents/services-migration.md). Pure refactor — no visual or behavioral change.

## What moved where

- **`pacts/event_settings.py`** (new) — `EventSettingsServiceProtocol`, `EventSettingsRepos` bundle (events, event_settings, event_proposal_settings, proposal_categories, session_fields), `EventDisplaySettingsContextDTO`, `ProposalSettingsUpdateData` write TypedDict, `EventSlugTakenError`. Named around the event-settings noun to stay off `pacts/event.py`, which #726 and #728 were rewriting at the time. Both have since merged, and none of the still-open siblings (#625/#626/#719/#730) touch either fold target, so a TODO now books the fold into `pacts/event.py` / `mills/event.py` as a mechanical follow-up rather than listing blockers that no longer exist.
- **`mills/event_settings.py`** (new) — `EventSettingsService` with `update_general`, `get_display_context`, `update_displayed_fields`, `get_proposal_settings`, `update_proposal_settings`. Constructor takes `TransactionProtocol` + the repos bundle (keyword-only).
- **`pacts/services.py` / `inits/services.py`** — one `event_settings` leaf each; **`inits/repositories.py`** — `event_settings` and `event_proposal_settings` repo leaves.
- **View** — shrank to glue: form parsing, redirects, flash messages. No repo access, no models.

## Transaction boundary

The proposal-settings POST previously opened `request.di.uow.atomic()` in the view around four writes (description, event dates, anonymous flag, optional apply-dates-to-categories fan-out). That block now lives inside `EventSettingsService.update_proposal_settings` via the injected `TransactionProtocol` — views no longer start transactions.

## Authz / scoping

Every request-supplied slug is resolved against `current_sphere_id` inside the service (`events.read_by_slug(slug, sphere_id)`) before any write. Slug uniqueness is not pre-checked: `update_general` writes inside `TransactionProtocol.savepoint()` and, on `DatabaseConstraintError`, re-reads to see whether another event in the sphere actually holds the slug — a pre-flight read loses the race against a concurrent rename, and the same table also carries a date-range check constraint, so only a genuine slug clash is re-labelled `EventSlugTakenError`. Unit tests assert a foreign/missing slug raises `NotFoundError` with no side effects for each write path.

## Tingle delta (vs origin/main)

`tingle check --base origin/main`: **no new debt, 52 occurrences paid off** across 32 metrics. Notably `request-di-uow` −16, `old-subdomain-loc` −38 net. Two additions, both deliberate: one `dict[str, object]` context annotation in the view (`type-object`) and the fold TODO in `pacts/event_settings.py` (`todo-comments`).

## Test & check status (run through the poetry venv with `.env.test`, since mise isn't available in this sandbox)

Numbers below are from the current tip, which is merged up to date with `main` (picking up the sibling migrations #728 `panel_time_slots`, #729 `tracks_panel` and #732 `encounters`).

- `pytest -n auto tests/integration tests/unit`: **3439 passed, 9 skipped, 0 failed**
- the three event-settings integration test files (`test_event_settings_page.py`, `test_event_display_settings_page.py`, `test_event_proposal_settings_page.py`): 64 tests, all passing. They are the regression guard and were not weakened — the only edits add two cases for non-numeric field ids and retarget one "Event not found." redirect assertion to `/panel/`
- new `tests/unit/test_event_settings_service.py`: 15 passed (mocks the specific repo protocols + transaction, no UoW mocks)
- `ruff check`, `black --check`, `mypy src` (252 files), `import-linter` (7 contracts kept), `vulture`, `codespell`: all clean
- `pylint src` and `pylint tests`: 10.00/10
- Not run (sandbox limitations, honest note): `ast-grep scan` (binary not installed in the venv) and `impeccable_lint.py` (fails on an npm GitFetcher error unrelated to this diff); djlint/e2e untouched — no template changes

---


src/ludamus/mills/event_settings.py
```diff
@@ -0,0 +1,98 @@
...
+            raise EventSlugTakenError
+        self._repos.events.update(current_event.pk, data)
+
+    def _slug_taken(self, slug: str, sphere_id: int) -> bool:
+        try:
+            self._repos.events.read_by_slug(slug, sphere_id)
+        except NotFoundError:
+            return False
+        return True
+
```
hasparus, line 58: **The slug check looks race-safe now that it lives in the service, but it isn't — and the codebase already has the tool to make it real.**

`update_general` does read → `_slug_taken` → `update` with no transaction and no lock. Two managers renaming events in the same sphere to the same slug both pass `_slug_taken`, and the loser hits `event_has_unique_slug_and_sphere` (`links/db/django/models.py:349`) inside `EventRepository.update` → raw `IntegrityError` → 500, not `EventSlugTakenError`. That's identical to pre-PR behaviour so it isn't a regression, but this PR is the moment the invariant moved into a service that owns its transaction boundary, and it's the natural moment to close it.

The canonical translation already exists: `TransactionProtocol.savepoint()` re-raises `IntegrityError`/`DataError` as `DatabaseConstraintError` (`links/db/django/transaction.py:24-34`, used by `mills/crowd.py`, `mills/submissions/engine.py`). Wrapping the write and mapping the constraint failure to `EventSlugTakenError` makes the DB the single source of truth for the invariant and turns `_slug_taken` from *the* check into a nicety.

The code-judo version deletes the helper, the extra read and the branch outright:

```python
def update_general(self, *, sphere_id: int, slug: str, data: EventUpdateData) -> None:
    current_event = self._repos.events.read_by_slug(slug, sphere_id)
    try:
        with self._transaction.savepoint():
            self._repos.events.update(current_event.pk, data)
    except DatabaseConstraintError as exc:
        raise EventSlugTakenError from exc
```

One real caveat, so don't apply that blindly: `Event` also carries the `event_date_times` `CheckConstraint`, and `EventSettingsForm` doesn't enforce `publication_time <= start_time < end_time`, so a bad date range would surface through the same `DatabaseConstraintError` and get mislabelled "slug already exists". Either inspect the constraint name, or keep `_slug_taken` as the friendly fast path and add the `savepoint()` wrapper purely so the race loses with a domain error instead of a 500. Either way the current shape asserts an invariant it doesn't actually hold.

Non-blocking, but please don't leave it as-is silently — if you'd rather defer, say so in the PR body so the next reader doesn't assume the check is authoritative.

---

src/ludamus/pacts/event_settings.py
```diff
@@ -0,0 +1,59 @@
...
+class EventSettingsServiceProtocol(Protocol):
+    def update_general(
+        self, *, sphere_id: int, slug: str, data: EventUpdateData
+    ) -> None: ...
+    def get_display_context(self, event_pk: int) -> EventDisplaySettingsContextDTO: ...
+    def update_displayed_fields(
+        self, *, sphere_id: int, slug: str, selected_ids: list[int]
+    ) -> None: ...
+    def get_proposal_settings(self, event_pk: int) -> EventProposalSettingsDTO: ...
+    def update_proposal_settings(
```
hasparus, line 69: **The protocol advertises two different scoping contracts, and the bare-pk half is the one that writes.**

Three methods take `sphere_id` + `slug` and self-scope. `get_display_context(event_pk)` and `get_proposal_settings(event_pk)` take a bare pk and scope nothing. Today that's safe — both callers pass `current_event.pk` from `get_event_context`, which resolved the slug against `current_sphere_id` — so this is not an open IDOR and I'm not claiming it is.

What makes it worth fixing rather than shrugging at: both bare-pk methods are named `get_*` but **create rows**. `get_display_context` → `EventSettingsRepository.read_or_create` → `EventSettings.objects.get_or_create(event_id=...)`; `get_proposal_settings` → `EventProposalSettingsRepository.read_or_create_by_event` → same. So the protocol offers a "getter" that any future caller can hand an arbitrary int and have it materialise settings rows for a foreign event. That is exactly the shape the panel object-scope rule exists to prevent, sitting one careless caller away, in the same protocol whose three sibling methods do it right.

Two options, both cheap: take `sphere_id` + `slug` uniformly and let the service do the scoped read (costs one query on two GET paths, buys a protocol you can't misuse), or keep the pk but rename to `read_or_create_*` so the write is visible at the call site. I'd take the first — uniformity across the five methods is worth more than the query, and it means the whole service is self-guarding rather than three-fifths self-guarding.

---

src/ludamus/gates/web/django/chronology/panel/views/event_settings.py
```diff
@@ -120,35 +121,25 @@ def get(self, _request: PanelRequest, slug: str) -> HttpResponse:
...
 
     def post(self, _request: PanelRequest, slug: str) -> HttpResponse:
-        sphere_id = self.request.context.current_sphere_id
-
-        try:
-            current_event = self.request.di.uow.events.read_by_slug(slug, sphere_id)
-        except NotFoundError:
-            messages.error(self.request, _("Event not found."))
+        if self.get_current_event(slug) is None:
             return redirect("panel:index")
```
hasparus, line 129: **Verdict on self-flagged deviation (a): the pre-flight is genuinely load-bearing, but the current shape hides why and still pays for two reads.**

I checked whether this could just be deleted. It can't: drop it and a POST with a bad slug + an invalid form falls into `_render_with_form`, which calls `get_event_context(slug)`, gets `({}, None)` back, and renders `panel/settings.html` with no `current_event`, no `events`, no `stats`. So the guard is real, and this is not a gratuitous double-read. Good instinct keeping it.

The problem is that nothing in the code says that, and a future "why are we reading the event twice?" cleanup will delete it and ship a broken template. There's a judo move that removes the redundancy *and* makes the guard self-evident — `_render_with_form` already builds the context you're re-deriving:

```python
def post(self, _request, slug):
    context, current_event = self.get_event_context(slug)
    if current_event is None:
        return redirect("panel:index")

    form = EventSettingsForm(self.request.POST, self.request.FILES)
    if not form.is_valid():
        return self._render_with_form(context, form)
    ...
```

That drops `services.events.read_by_slug` on every POST, drops the second `event_panel.load_context` on the invalid-form path, and the guard now visibly produces the context the error path needs. The service keeps its own authoritative sphere-scoped read — that one is not redundant, it's the write-time scope check and it should stay.

Related, while you're in here: after this guard, the `except NotFoundError` at L138-140 is only reachable if the event vanishes mid-request, and it redirects to `panel:event-settings` for an event that no longer exists (which then bounces to `panel:index`). The other two views redirect straight to `panel:index` for the same condition. `test_error_event_not_found_during_update` pins the current target by patching `EventRepository.update`, so changing it means touching that test — your call whether the consistency is worth it, but the three views disagreeing on where "event not found" lands is the kind of thing that quietly becomes permanent.

---

src/ludamus/gates/web/django/chronology/panel/views/event_settings.py
```diff
@@ -180,39 +171,29 @@ def get(self, _request: PanelRequest, slug: str) -> HttpResponse:
...
+        context["fields"] = display.fields
+        context["filterable_field_ids"] = display.displayed_field_ids
 
         return TemplateResponse(self.request, "panel/display-settings.html", context)
 
     def post(self, _request: PanelRequest, slug: str) -> HttpResponse:
-        sphere_id = self.request.context.current_sphere_id
+        selected_ids = [
+            int(pk) for pk in self.request.POST.getlist("displayed_session_fields")
+        ]
```
hasparus: `int(pk)` on raw POST data with nothing catching `ValueError` — `displayed_session_fields=abc` from any panel manager is a 500 with no useful message. Pre-existing, but this PR moved the block and it's the one parsing step left in the view, so it's the cheapest possible fix while you're here:

```python
selected_ids = [
    int(pk)
    for pk in self.request.POST.getlist("displayed_session_fields")
    if pk.isdigit()
]
```

The service already discards any id that isn't a public field of the event, so silently dropping garbage is consistent with how the rest of that list is treated — no new swallow, just not crashing. Minor.

---

src/ludamus/pacts/event_settings.py
```diff
@@ -0,0 +1,59 @@
+from __future__ import annotations
```
hasparus, line 9: **Module placement — I accept the stated reason, but please book the fold-back.**

GLIMPSE slices `pacts`/`mills` by **noun, then verb**. `event_settings` is neither: it's an entity cut of the `event` noun, and `pacts/event.py` (38 lines) / `mills/event.py` (46 lines) are nowhere near the ~1000-line threshold that justifies a split. By the book this belongs in `event.py` — folded in, that's 97 and 144 lines respectively, still tiny. The symmetry rule is satisfied either way since you added both halves, so this is placement judgement, not a violation.

The PR body gives merge-friction with #625/#626/#719/#726/#728/#730 as the reason, and GLIMPSE does list merge friction as a legitimate split trigger, so I'm not pushing back on the decision for this PR. The concern is the aftermath: once those land, nobody comes back, and the `event` noun's contracts stay permanently spread across `pacts/event.py`, `pacts/event_settings.py` and `pacts/legacy.py`. `mills/` already shows where that ends up — `panel_time_slots.py`, `session_modal.py`, `field_values.py`, `url_recovery.py` are all entity/page-shaped modules that presumably each had a good local reason too.

Either file a follow-up to fold `event_settings` into `event.py` after the parallel PRs merge, or if you want the cut to be permanent, make it `pacts/event/{panel,settings}.py` + `mills/event/{panel,settings}.py` so the noun stays one namespace. Both satisfy the ≥2-files-per-folder rule today. Not blocking.

---

hasparus (commented):
## Thermo-nuclear review — verdict: COMMENT

A clean strangler-fig increment: the view really did shrink to glue, the `uow.atomic()` block landed in the service where it belongs, and every request-supplied id on the write paths (the slug, and the `displayed_session_fields` list intersected against the event's public fields) is scoped inside the service rather than the view — which is the shape `docs/refactors/panel-object-scope-authz.md` asks for. The 11 unit tests mock the specific repo protocols, not the UoW, and the three untouched integration files remain the regression guard.

I verified the claims rather than taking the PR body's word for them. In a worktree at `e05cd610`: `tests/unit/test_event_settings_service.py` 11 passed; the three integration files 62 passed; `ruff` and `mypy` clean on the changed modules. `di.uow` is gone from the view file. The `NotFoundError` and duplicate-slug integration tests still exercise the paths they claim to, through the new wiring.

Nothing here blocks. Two findings are more than nits, and both are places where the PR preserves complexity it could delete:

**1. The sphere-scoped slug check isn't race-safe** (`mills/event_settings.py:28-47`). Read → check → write, no transaction, no lock. The `event_has_unique_slug_and_sphere` constraint keeps the data correct, but the loser of a concurrent rename gets a raw `IntegrityError` → 500 instead of `EventSlugTakenError`. Same as pre-PR, so not a regression — but this is the moment the invariant moved into a service that owns transactions, and `TransactionProtocol.savepoint()` → `DatabaseConstraintError` already exists for exactly this translation. Doing it deletes `_slug_taken`, a read, and a branch. One caveat about the `event_date_times` check constraint in the inline comment.

**2. The protocol has two scoping contracts, and the unscoped half writes** (`pacts/event_settings.py:48-57`). `get_display_context` / `get_proposal_settings` take a bare `event_pk` while their three siblings take `sphere_id` + `slug`. Not an IDOR today — both callers pass a pk from a sphere-scoped read — but both resolve to `get_or_create`, so the protocol offers a `get_*` that materialises settings rows for any int a future caller supplies. Uniform `sphere_id` + `slug` costs two queries and removes the hazard.

**On the self-flagged deviations**, judged on merit:

- **(a) the pre-flight `get_current_event`** — justified, not a smell. I traced it: without it, bad slug + invalid form renders `panel/settings.html` with an empty context. But the code doesn't say that, and there's a judo move that removes the redundancy *and* makes the guard self-evident — hoist `get_event_context` and pass the context into `_render_with_form`. Details inline.
- **(b) `ProposalSettingsUpdateData`** — fine. One quibble not worth its own thread: `apply_dates_to_categories` is a behaviour flag, not data, so the "write shape" TypedDict is doing two jobs. An explicit keyword-only `apply_dates_to_categories: bool` would keep the TypedDict a pure payload. Taste-level; ignore if you disagree.
- **(c) the `EventSettingsRepos` bundle** — correct, mirrors `EnrollmentRepos` / `ImportRepos`. Those two are `frozen=True` and this one isn't; the codebase is already inconsistent there, so no action needed.

**On the slug-change path**, which you asked to be checked: the redirect to the new slug is preserved and the repos are stateless `@staticmethod`s, so wiring `event_settings` / `event_proposal_settings` onto both `Repositories` and the legacy `UnitOfWork` creates no identity-map divergence. Keeping the UoW leaves in place is what `services-migration.md` prescribes. I found no cache or denormalised copy keyed on event slug elsewhere in the tree, so no stale-URL fallout beyond the redirect you already handle.

**Two smaller things** in inline comments: an unguarded `int(pk)` over raw POST data (500 on hand-crafted input, pre-existing but moved by this diff), and module placement — `event_settings` as a top-level slice beside a 38-line `pacts/event.py`. I accept the merge-friction justification for now; the ask is a follow-up to fold it back, not a change here.

**Systemic, explicitly not this PR's fault:** `inits/services.py` is now 45 leaves and `inits/repositories.py` 41, against GLIMPSE's ~12-per-namespace guidance. This PR adds 1 and 2. The registries need bucketing by noun soon, and each migration PR makes that refactor slightly more expensive — worth someone scheduling before the strangler-fig finishes.

Ship it once you've decided what to do about (1) — a "deferring, tracked in #NNN" is a fine answer, I'd just rather the race be a known deferral than an assumed non-issue.

---

src/ludamus/mills/event_settings.py
hasparus: Fixed in `66a33c54`: the pre-flight read is gone and the write runs inside `TransactionProtocol.savepoint()`, so the unique index is now the authority — heeding your `event_date_times` caveat, the handler doesn't blanket-map `DatabaseConstraintError` but re-reads the target slug and only raises `EventSlugTakenError` when another event actually holds it, letting anything else propagate untouched. Unit tests cover the concurrent-collision path (including the `__cause__` chain) and both non-slug-constraint paths.

---

src/ludamus/pacts/event_settings.py
hasparus: Took your first option in `9951a261`: `get_display_context` and `get_proposal_settings` now take `sphere_id` + `slug` like their three siblings and do the scoped read themselves, so all five methods are self-guarding and no caller can hand the protocol a bare foreign int. New unit tests assert a foreign slug raises `NotFoundError` before `read_or_create` / `read_or_create_by_event` is reached, so no settings row is materialised.

---

src/ludamus/gates/web/django/chronology/panel/views/event_settings.py
hasparus: Applied the judo in `49dcea1a`: `post` now hoists `get_event_context`, redirects on `current_event is None`, and passes the context into `_render_with_form`, so the second `load_context` and the `services.events.read_by_slug` are gone and the guard visibly produces what the error path renders (a comment says so too). I also took the consistency change — the `NotFoundError` handler redirects to `panel:index` like the other two views, and `test_error_event_not_found_during_update` was updated to match; `EventSlugTakenError` still stays on the settings page since the event does exist there.

---

src/ludamus/gates/web/django/chronology/panel/views/event_settings.py
hasparus: Guarded in `49dcea1a`. I went one step past the suggested filter: rather than silently dropping non-numeric values, the view rejects the request with `_("Invalid field selection.")` (Polish added to the catalog) and redirects back with the saved selection untouched — CLAUDE.md's "useful errors, no silent swallows", and the only way to send a non-digit here is a hand-crafted request, so there's no real UX cost. Integration test covers `displayed_session_fields=abc` returning the message and leaving the stored fields unchanged.

---

src/ludamus/pacts/event_settings.py
hasparus: Booked in `d8e67ee2` — a TODO at the top of `pacts/event_settings.py` records the fold-back into `pacts/event.py` / `mills/event.py` once #625, #626, #719 and #726 land, and says why the cut exists (merge friction only, both halves nowhere near the ~1000-line threshold). Left the module where it is for this PR as you accepted; it also shows up as a `todo-comments` tick in `tingle check`, which is the intended reminder.

---

src/ludamus/mills/event_settings.py
```diff
@@ -0,0 +1,115 @@
...
+        except DatabaseConstraintError as exc:
+            # The same table also carries a date-range check constraint, so
+            # only re-label the failure when another event really holds the
+            # slug now; anything else keeps its own error.
+            if self._slug_held_by_other(
+                slug=data.get("slug"), sphere_id=sphere_id, event_pk=current_event.pk
+            ):
+                raise EventSlugTakenError from exc
+            raise
+
```
hasparus, line 48: **Round 1 (a) — verified correct, and the re-read is the better of the two options I offered.**

I traced this rather than taking the reply's word for it, because savepoint-plus-`ATOMIC_REQUESTS` is easy to get subtly wrong. All four things I was worried about check out:

**The failed statement leaves the outer transaction usable.** `edges/settings.py:201` sets `ATOMIC_REQUESTS: True`, so `savepoint()`'s inner `transaction.atomic()` (`links/db/django/transaction.py:29`) is a *nested* atomic and Django issues a real `SAVEPOINT`. On the `IntegrityError` the inner `__exit__` does `savepoint_rollback(sid)` rather than setting `needs_rollback` on the request-level atomic, so the connection is not poisoned and the subsequent query is legal — no `TransactionManagementError`.

**The re-read is outside the savepoint and can see the winner.** The `try` wraps the `with`, so by the time `_slug_held_by_other` runs the savepoint has already been rolled back. Django runs Postgres at READ COMMITTED, and the unique violation can only have fired if the other transaction *committed* (an uncommitted duplicate would have blocked on the index lock instead), so the fresh statement snapshot is guaranteed to see it. This is the case that would silently not work under REPEATABLE READ; at the isolation level this project actually runs, it does.

**Nothing gets swallowed or mislabelled on the non-slug path.** `_event_update_data` always sets `"slug"`, so the `slug is None` early-out is dead for this caller but harmless; the `except NotFoundError: return False` falls through to a bare `raise`, which re-raises the original `DatabaseConstraintError` untouched. An `event_date_times` violation therefore keeps its own error rather than being dressed up as a slug collision. The two unit tests at `tests/unit/test_event_settings_service.py` pin both non-slug branches (slug held by self, slug free), which is exactly the pair I'd have asked for.

**`__cause__` survives:** `EventSlugTakenError.__cause__` is the `DatabaseConstraintError`, whose own `__cause__` is the `IntegrityError` — two hops, both explicit, and asserted in `test_update_general_raises_when_a_concurrent_rename_takes_the_slug`.

One thing worth saying out loud: you picked the re-read over inspecting the constraint name, and that's the right call for reasons beyond the ones in my original comment. Constraint-name mapping would have put the string `event_has_unique_slug_and_sphere` — a `links`-layer schema detail — inside a `mills` service. The re-read expresses the same decision in domain terms ("does another event hold this slug?") and costs one query only on the already-failed path. Better layering, not just an equivalent workaround.

The real proof is `test_error_on_duplicate_slug`, which exercises this against a live database and asserts `event.slug == original_slug` after the failure — i.e. it verifies the savepoint rollback actually undid the write rather than just that the right exception came out. `test-postgres` is green, so this path is confirmed on the engine where `ATOMIC_REQUESTS` is set.

Non-blocking follow-up, explicitly **pre-existing and not this PR's job**: `EventSettingsForm` still has no `clean()` enforcing `publication_time <= start_time < end_time`, so a manager who submits an inverted date range trips `event_date_times` and — correctly, per the above — gets a raw `DatabaseConstraintError` and a 500. That was a 500 before this PR too, so nothing regressed; but now that the constraint failure has a named home in the service, a form-level `clean()` would turn the last 500 on this page into a field error. Worth a follow-up issue, not a change here.

---

src/ludamus/gates/web/django/chronology/panel/views/event_settings.py
```diff
@@ -180,39 +176,31 @@ def get(self, _request: PanelRequest, slug: str) -> HttpResponse:
...
 
         return TemplateResponse(self.request, "panel/display-settings.html", context)
 
     def post(self, _request: PanelRequest, slug: str) -> HttpResponse:
-        sphere_id = self.request.context.current_sphere_id
+        raw_ids = self.request.POST.getlist("displayed_session_fields")
+        if not all(raw_pk.isdigit() for raw_pk in raw_ids):
+            messages.error(self.request, _("Invalid field selection."))
+            return redirect("panel:event-display-settings", slug=slug)
+        selected_ids = [int(raw_pk) for raw_pk in raw_ids]
```
hasparus, line 192: **Nit, non-blocking: `isdigit()` doesn't quite imply `int()` succeeds.**

The reject-with-a-message shape is better than the silent filter I suggested, and I agree with the reasoning — the only way to send a non-digit here is a hand-crafted request, so there's no UX cost to being strict, and it satisfies "useful errors, no silent swallows" properly. The integration test checks both halves that matter: the message comes back *and* `displayed_session_fields` is unchanged in the DB.

The guard just doesn't close quite the whole gap. `str.isdigit()` is true for Unicode `Numeric_Type=Digit`, which is a superset of what `int()` parses:

```
'²'.isdigit()   -> True     int('²')  -> ValueError
'²'.isdecimal() -> False
'٣'.isdigit()   -> True     int('٣')  -> 3        # fine either way
```

So `displayed_session_fields=²` still walks past the guard on line 189 and 500s on line 192 — the same crash the guard exists to prevent, just needing a weirder payload. `isdecimal()` is exactly `int()`'s accepted set for this purpose and is a one-character change.

Two reasons this is a nit and not a finding: it takes a deliberately hostile request from someone who already has panel-manager rights, and `isdigit()`-then-`int()` is the established convention across ~20 call sites in this codebase (`panel/views/tracks.py:45`, `panel/views/proposals.py:586`, `panel/views/base.py:60`, `chronology/views.py:780`, …), all carrying the identical latent `ValueError`. You matched the house pattern, which is the defensible choice. Flip this one to `isdecimal()` if you feel like it; sweeping the rest is its own PR and I'm not asking for it here.

---

hasparus (commented):
**VERDICT: APPROVE** — posted as a comment only because GitHub refuses `APPROVE` on a self-authored PR ("Can not approve your own pull request"). Treat this as an approval; nothing below blocks merge.

## Round 2 — approve

All five round-1 findings are genuinely resolved. I verified each against the code at `d8e67ee2` rather than against the replies, and two of them I went looking for holes in specifically. Two nits ride along; neither blocks.

### What I checked, and what held

**(a) TOCTOU slug check — correct, and the design choice is better than either option I offered.** Full trace in the inline comment on `mills/event_settings.py`. Summary: the nested `atomic()` under `ATOMIC_REQUESTS` issues a real savepoint, so the constraint failure rolls back only the inner block and leaves the request transaction usable; the re-read sits outside the savepoint and, at READ COMMITTED with the loser only failing after the winner committed, is guaranteed to see the conflicting row; the non-slug path re-raises the original error bare, so `event_date_times` can't be mislabelled or swallowed; and `__cause__` chains two hops to the `IntegrityError`. Choosing a domain re-read over constraint-name mapping keeps a `links`-layer schema string out of a `mills` service — that's better layering, not just an equivalent dodge. `test_error_on_duplicate_slug` proves it against a live DB by asserting the slug is unchanged after the failure, i.e. that the rollback actually happened, and `test-postgres` is green.

**(b) Scoping contract — uniform, and I confirmed no `get_*` can still create a row for an unscoped id.** All five protocol methods now take `sphere_id` + `slug`, and every one opens with `read_by_slug(slug, sphere_id)` before touching anything else. The two that call `read_or_create` / `read_or_create_by_event` reach them only through a pk derived from that scoped read, so the "getter that materialises rows for a foreign event" shape is gone — there is no longer a bare `int` on the protocol surface for a future caller to misuse. I also swept the rest of the service for request-supplied ids: `selected_ids` is intersected against public fields of the scoped event, which is the only other user-controlled input, and it's scoped. That satisfies `docs/refactors/panel-object-scope-authz.md` end to end. The four `assert_not_called()` tests are the right assertion — they pin *no side effect*, not just the exception.

**(c) Pre-flight redundancy — applied, and the redirect unification is a real fix rather than a test edit.** `post` hoists `get_event_context`, redirects on `None`, and hands the context to a keyword-only `_render_with_form`; the second `load_context` and the `services.events.read_by_slug` are both gone, and the comment says why the guard exists so the next cleanup doesn't delete it. On the test: `test_error_event_not_found_during_update` was pinning a genuinely wrong destination — it patches `EventRepository.update` to raise, and the old code redirected to `panel:event-settings` for an event that had just been reported missing, which then bounces to `panel:index` anyway. Changing the assertion to `/panel/` records where the user actually lands and matches the other two views. Nothing meaningful was un-pinned. `EventSlugTakenError` correctly still redirects to the settings page — the event exists there, so staying put is right — and `test_error_on_duplicate_slug` still holds that line.

**(d) `int(pk)` guard — the error path is real and the saved selection survives.** Message plus redirect, no silent drop; the integration test asserts both the message and that `displayed_session_fields` still contains the original field. `"Nieprawidłowy wybór pola."` is correct Polish and touches none of the reserved terms in CLAUDE.md's glossary. The catalog diff against the merge base is *exactly* the four added lines and nothing else — no flag churn, no `POT-Creation-Date` drift, no collateral on the 11 `python-brace-format` entries you were worried about. The hand-insertion worked, and `messages-check` (in the green `checks` job) independently confirms extraction is current with nothing untranslated or fuzzy. One residual nit inline: `isdigit()` is a hair wider than `int()` accepts.

**(e) Fold-back booked.** TODO is at the top of `pacts/event_settings.py` with the reason and the blocking PRs named, and it registers as a `todo-comments` tick so it can't quietly evaporate. That's what I asked for.

### Judging the tip as a whole

This is a clean strangler-fig step and it leaves the codebase smaller, not just rearranged. The view drops 38 lines and every `request.di.uow` reference in it; the service that replaces them is 115 lines with no branching worth the name. No file goes anywhere near 1k. No new special-case branch got bolted onto an unrelated flow — the one new conditional (the field-id guard) is a parse-boundary check at the top of a handler, which is where it belongs. The DI wiring is boring and follows the existing `cached_property` pattern. The unit suite is genuinely good: it tests the failure branches, the side-effect absence, and the exception chain, not just the happy path.

The nits, both explicitly optional: `isdigit()` → `isdecimal()` on the field-id guard (inline; you matched a ~20-site house convention, so this is a preference, not a defect I'm asking you to fix), and a pre-existing 500 on inverted event dates that a form-level `clean()` would close (inline; pre-dates this PR, belongs in a follow-up). Separately, `require_event_context` in `event/panel/views/base.py:51` already packages the hoisted guard you hand-rolled — but the other four handlers in this file hand-roll it too, so switching just one would make the file less consistent, not more. Leave it.

### CI at `d8e67ee2`

11 of 12 green — `checks` (lint + `messages-check`), `test-postgres`, `tingle`, `impeccable`, `frontend-analysis`, `varlock-scan`, CodeQL ×4, staging dispatch. `test` was still `in_progress` when I looked. `test-postgres` is the one that matters for the savepoint work and it passed, and `messages-check` passing is the independent check on the catalog. Worth a glance at `test` before merge, but nothing in this diff is SQLite-sensitive.

Nice work on (a) in particular — the re-read landed on the right side of the layering question, and the tests pin the parts that would actually break.

---

src/ludamus/gates/web/django/chronology/panel/views/event_settings.py
hasparus: Flipped to `isdecimal()` in `e6a9b31d`. You're right that this is the same class of bug the round-1 finding asked me to fix, so the guard wasn't actually finished — `'²'.isdigit()` being true meant the payload still reached `int()` on line 192 and 500'd exactly as before, just needing a stranger request to get there.

I confirmed that rather than assuming it: with the guard on `isdigit()`, the new case fails with `ValueError: invalid literal for int() with base 10: '²'` raised from line 192; with `isdecimal()` it takes the `_("Invalid field selection.")` path and leaves the saved selection untouched. The existing rejection test is now parametrized over `("abc", "²")` so both halves stay pinned.

On the wider pattern: I left the other sites alone, per your note. For whoever picks that up, the sweep is `isdigit()` → `isdecimal()` wherever the result feeds `int()` — you named `panel/views/tracks.py:45`, `panel/views/proposals.py:586`, `panel/views/base.py:60` and `chronology/views.py:780`, and it's around twenty call sites in total. Worth doing as one mechanical PR once the parallel branches land, since several of those files are currently held by open PRs and touching them here would have meant conflicts for a nit. Two things make it more than a find-and-replace: a handful of those sites feed the value somewhere other than `int()`, where `isdigit()` may genuinely be what's wanted, and the ones that silently filter rather than reject will keep dropping `²` quietly instead of crashing — a behaviour change worth deciding on deliberately rather than inheriting from a sed.

---

src/ludamus/mills/event_settings.py
hasparus: Thanks for tracing the savepoint path rather than taking my word for it — the `ATOMIC_REQUESTS` nesting and the READ COMMITTED visibility argument are the two things I was least sure of when I wrote it, and it's genuinely useful to have them written down. Your point about constraint-name mapping leaking `event_has_unique_slug_and_sphere` into `mills` is the better articulation of why the re-read wins; I'd reached for it on cost grounds and hadn't framed it as a layering property.

On the `EventSettingsForm.clean()` follow-up: agreed on the substance, and I'm deliberately not doing it here. This PR is a `request.di.uow` → `request.services` migration, and the inverted-date-range 500 is pre-existing and unchanged by it — the constraint fired the same way before, it just didn't have a named home in a service. Adding form validation would mean new `clean()` logic, new Polish strings in the catalog, and its own tests for each ordering pair, none of which the migration needs and all of which would make the diff harder to review against the "same behaviour, different plumbing" claim it's making.

Worth noting for the follow-up that it's three orderings, not one: `publication_time <= start_time`, `start_time < end_time`, and the transitive case where publication is after end. A `clean()` that reports the specific violated pair as a field error is more useful than one blanket non-field message, and the `event_date_times` constraint should stay as the backstop either way — the form can't cover a write that doesn't come through it. Happy to open the issue and reference this thread if you'd like it tracked rather than left in review history; say the word and I'll file it.

---

src/ludamus/pacts/event_settings.py
```diff
@@ -0,0 +1,71 @@
+# TODO(hasparus): Fold this module into pacts/event.py, and
+# mills/event_settings.py into mills/event.py, in a follow-up. GLIMPSE slices
+# pacts and mills by noun then verb, and `event_settings` is an entity cut of
+# the `event` noun, not a verb; folded in, both files stay far below the
+# ~1000-line threshold that would justify the split. The split existed only to
+# keep this branch off files #726 and #728 were rewriting; both have merged, so
```
hasparus, line 6: Nit, non-blocking, and the only inaccuracy I found anywhere in this PR: the attribution here is wrong for #726.

The TODO says the split existed "to keep this branch off files #726 and #728 were rewriting". #728 checks out — `d44ef412` really did touch `pacts/event.py` and `mills/event.py`, which is exactly why staying off them was the right call at the time. #726 didn't. As merged (`870a0b2c`) it touched `mills/printing.py`, `mills/timeslots.py`, `pacts/printing.py`, the print gates, templates, the catalog and its tests — neither fold target appears in its file list. I can't rule out that an earlier revision of that branch did, but the commit that landed didn't.

The conclusion is unaffected — the fold targets are uncontested now and the move is mechanical either way — so this is a one-word fix to "#728 was rewriting" whenever this file is next touched, not something to respin for. Worth correcting because a TODO outlives its PR, and the next reader tracing "why is this module here?" will look up #726 and find nothing. The same sentence is in the PR body, so it's a paired edit.

---

hasparus (commented):
**APPROVE.** The merge is clean, verifiably so, and nothing blocking remains. One nit accompanies the approval and one observation for the strangler-fig follow-up; neither should hold the merge.

_(Filed as a COMMENT only because GitHub refuses a formal approval on a self-authored PR — "Can not approve your own pull request". The verdict is APPROVE.)_

## The merge is clean, and I can prove it rather than assert it

`origin/main` is `2ca9e64f`, and there are **zero** commits in `main` that aren't in this branch — `f41b2332` merged the current tip of `main`, not a stale one. That makes the merge verification unusually cheap and unusually conclusive: `git diff origin/main origin/claude/uow-event-settings` and `git diff $(git merge-base ...) origin/claude/uow-event-settings` produce **byte-identical output** — the same 10 files, +671/-88. Every file in the repo outside those 10 is identical to `main`.

That single fact settles all four merge questions at once, because a dropped hunk from either side would have to show up as a difference against `main`:

- **`pacts/services.py`** — diff vs `main` is exactly `+1` import line and `+2` for the `event_settings` property. Every sibling leaf `main` gained (`tracks_panel`, `encounters`, `panel_time_slots`, ...) is present and appears exactly once. Nothing dropped, nothing duplicated.
- **`inits/services.py`** — same shape: `+2` imports, `+13` for the `event_settings` cached property. The `EventSettingsRepos` bundle names five repos, all of which exist on `Repositories`.
- **`inits/repositories.py`** — `+8`, the two new repo leaves, on top of everything `main` has.
- **`django.po`** — `+4` lines against `main`, one entry, zero deletions. So nothing was lost from either side by construction. I ran `msgfmt --check` over the merged catalog: it passes. The five apparently-duplicated msgids (`Category`, `Name`, `Session`, `All categories`, `Schedule`) are disambiguated by `msgctxt` and all predate this branch.

On the `"Apply"` / `"Timetable with descriptions"` reasoning: **verified, and for a stronger reason than the merge author gave.** They're absent from the merged catalog, but they're equally absent from `main` — and, decisively, `git grep` finds neither string anywhere under `src/` at this tip. #726 removed the source strings, so the catalog is *correct* to have dropped them; had the auto-merge resurrected them, `makemessages` would strip them again. Nothing to fix.

## Semantic integrity post-merge

- **Savepoint dependency holds.** `DatabaseConstraintError` (`pacts/services.py:75`) and `TransactionProtocol.savepoint()` (`:88`) both survive the merge unchanged, and `links/db/django/transaction.py` still maps `IntegrityError`/`DataError` onto the domain error. `mills/event_settings.py` is now one of six consumers of that contract (`crowd.py`, `submissions/engine.py`, `submissions/import_log.py`, `panel/views/proposals.py`), so it's riding an established path, not a private one. The round-1 tracing of `ATOMIC_REQUESTS` nesting and READ COMMITTED visibility is unaffected by anything `main` brought in.
- **No symbol collides with post-#728 `pacts/event.py`.** That module exports `TimeSlotValidationError`, `EventPanelContextDTO`, `EventPanelServiceProtocol`, `PanelTimeSlotsServiceProtocol`; `pacts/event_settings.py` exports `EventSlugTakenError`, `EventSettingsRepos`, `EventDisplaySettingsContextDTO`, `ProposalSettingsUpdateData`, `EventSettingsServiceProtocol`. Disjoint — which also means the booked fold is a straight concatenation with no rename needed. `EventSettingsRepos` sits close to `links`' `EventSettingsRepository` and `pacts/legacy`'s `EventSettingsRepositoryProtocol`, but those are different names in different layers, not a shadow.
- **Nothing `main` brought in contradicts this branch.** `cfe1011f` (enrollment), #726 (print), #728/#729/#732 (sibling migrations) touch no event-settings view, service, template or repo. The one shared file with real merge risk was the catalog, covered above.
- **`@dataclass` DTOs are house style, not drift.** I checked before flagging `EventDisplaySettingsContextDTO` as off-pattern and it isn't — `pacts/submissions.py` alone has 12, plus `chronology.py`, `printing.py`, `tracks.py`. Withdrawn before it was raised.

## Prior rounds

Both round-1 findings I re-checked are still fixed at this tip and neither was disturbed by the merge: the pre-flight slug read is gone in favour of the index-plus-re-read (`mills/event_settings.py:31-56`), and all five protocol methods take `sphere_id` + `slug` and self-scope. The round-2 `isdecimal()` fix is in place at `views/event_settings.py:189`, parametrized over `("abc", "²")`. The deferred `EventSettingsForm.clean()` nit stays deferred and I'm not reopening it — it's pre-existing, it's a 500 that was a 500 before this PR, and it needs its own strings and tests.

## Quality of the change itself, judged as it will land

The view is genuinely glue now — 16 `request.di.uow` call sites in, zero out, and no repo or model reference left in the file. That's the whole point of the migration and it landed without leaving a residue. The transaction boundary moved from the view into `update_proposal_settings`, so the four-write fan-out is atomic under the service's own `TransactionProtocol` rather than the view reaching for `uow.atomic()`.

Two things I'd normally push on, and why I'm not:

- **The double read in `EventSettingsPageView.post`.** `get_event_context` reads, then the service reads again. That's the exact shape a future cleanup deletes. The comment at `:124-126` explains why it can't be, the guard visibly produces the context the invalid-form path renders, and the service's read is the authoritative write-time scope check that has to stay. Both reads earn their place.
- **The `_slug_held_by_other` re-read branch.** It's a conditional bolted onto an error path, which is normally a smell. Here it's the thing that keeps the `links`-layer constraint name `event_has_unique_slug_and_sphere` out of a `mills` service, and it's pinned by four unit tests covering slug-taken, concurrent-rename (including the `__cause__` chain), slug-held-by-self, and slug-free. Right call.

Test coverage is proportionate: 15 unit tests in `tests/unit/test_event_settings_service.py`, one per method plus a foreign-slug no-side-effects case for each of the five, which is exactly what the panel object-scope rule asks for. No file goes anywhere near 1000 lines (`mills/event_settings.py` 115, `pacts/event_settings.py` 71).

## PR body accuracy

Rewritten post-merge, so I spot-checked the numbers rather than trusting them. All hold: 16 `di.uow` call sites removed (counted at `2ca9e64`, zero remain); 15 new unit tests; 64 integration tests across the three event-settings files (36 + 13 + 15 — the 13 is the parametrized rejection test expanding, which is why a naive `def test_` count gives 63); `#732` is `notice_board` → `services.encounters`, so calling it "encounters" is right; `#625/#626/#719/#730` are all still open and I confirmed via the files API that #730 touches neither fold target. CI independently corroborates the lint claims the body could only run partially in the sandbox — all 16 checks green on `69e26185`, including `tingle`, `impeccable`, `checks`, `test` and `test-postgres`.

One inaccuracy, inline on `pacts/event_settings.py`: the "#726 and #728 were rewriting" attribution is right for #728 and wrong for #726. Cosmetic, and the conclusion it supports is still true.

## One observation, not a request

`uow.event_settings` (`links/db/django/uow.py:35`, protocol at `pacts/legacy.py:1435`) now has no callers anywhere in `src/` or `tests/` — this view was its last consumer. Leaving it is **correct**: `docs/agents/services-migration.md:76` says the legacy leaves "stay until the last view migrates. Don't slim them piecewise," and `vulture` is green. Noting it only so whoever finally closes the strangler fig knows this one is already collectable. `uow.event_proposal_settings` is *not* — `chronology/views.py:247,521` still use it.

## Merge recommendation

**Merge it.** The merge resolution is provably lossless on all four files flagged for scrutiny, no symbol or invariant `main` brought in conflicts with what this branch adds, both prior rounds' findings are fixed and stayed fixed, and CI is green on the exact tip. The one nit is a wrong PR number in a comment; fix it on the way past or in the fold-back follow-up, but it is not worth another round trip.


---

PR #734, opened 2026-07-30
fix(email): accept the ?tls= parameter, and report how production resolves EMAIL_URL

hasparus:
## The bug

Production has never delivered a single email. `EMAIL_URL` is
`smtp://resend:<key>@smtp.resend.com:587/?tls=True` — the URL Resend hands out —
and it resolves to **`EMAIL_USE_TLS = False`**:

- `django-environ` takes TLS from the *scheme* (`smtp+tls://`, `smtps://`) and
  files every other query parameter under `OPTIONS`. `?tls=True` uppercases to
  `TLS`, which is not in its two-item allow-list, so it lands in `OPTIONS`.
- `settings.py` never reads `OPTIONS`.

So Django opened a plaintext connection to port 587, Resend rejected it, and
`send_mail(fail_silently=True)` in `DjangoUserNotifier._deliver` swallowed the
error. Silent since 1 July.

`?tls=` is [dj-email-url](https://github.com/migonzalvar/dj-email-url) syntax,
which many providers document. Nobody typo'd anything; two libraries disagree.

## The fix

`edges/email.py` wraps `env.email_url()` — django-environ still parses — and:

- maps `?tls=` / `?ssl=` onto `EMAIL_USE_TLS` / `EMAIL_USE_SSL`, parsing real
  booleans, so `?tls=False` is off rather than a truthy `"False"` string
- **refuses** a parameter it cannot read, instead of dropping it — the failure
  mode that cost us six weeks now stops the boot
- adds the missing `EMAIL_USE_SSL` line, so `smtp+ssl://` no longer parses fine
  and connects in the clear

`consolemail://` and `filemail://` keep working, so dev and the staging inbox
are untouched. `.env.schema` documents the accepted forms.

## The check

`workflow_dispatch`-free now: the job runs on PRs touching it, resolves
`secrets.EMAIL_URL` through the same wrapper, and prints backend, host, port,
TLS/SSL and user. The key appears only as a length and a `re_` prefix test. It
already caught the live value:

```
backend: smtp
host: smtp.resend.com port: 587
tls: False ssl: False
user: resend
password: length 36 resend-shaped: True
PROBLEM: no TLS
```

## After merging

Redeploy production and the existing secret starts working as written — no
secret rotation needed, because `?tls=True` is now read.

Two follow-ups, deliberately not here: the `production` environment has no
deployment-branch policy, so any branch can read its secrets; and waitlist
promotion did not fire when session capacity was raised on the Skytower test
fixture, which is a separate bug from this one.

### DIFF
--- a/.env.schema
+++ b/.env.schema
@@ -63,6 +63,12 @@ MEMBERSHIP_API_CHECK_INTERVAL=15
 
 # Email transport URL. @sensitive (may embed smtp://user:pass@host), so it
 # carries no default here — settings.py falls back to consolemail://.
+#   consolemail://                        print to stdout (development)
+#   filemail:///app/logs/emails           write to a directory (staging inbox)
+#   smtp://user:pass@host:587/?tls=True   STARTTLS, the form Resend hands out
+#   smtp+tls://user:pass@host:587         the same thing, spelled as a scheme
+# TLS is off unless the scheme or ?tls= says otherwise, and a parameter this
+# project cannot read stops the boot rather than being dropped.
 # @sensitive @optional
 EMAIL_URL=
 # @optional
--- a/.env.schema
+++ b/.env.schema
@@ -126,3 +126,6 @@ GS_BUCKET_NAME=
 GS_CREDENTIALS_JSON=
 # @optional
 GS_LOCATION=media/
+
+# @sensitive @optional
+KOBOLD_APP_PASSWORD=
--- a/.github/workflows/prod-email-check.yml
+++ b/.github/workflows/prod-email-check.yml
@@ -1,0 +1,58 @@
+name: Prod email check
+
+on:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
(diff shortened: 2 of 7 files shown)


hasparus:
Shelved. Fixing the live secret directly instead — `smtp+tls://` on 587 needs no code and unblocks production today.

Left as a draft rather than closed, because the diagnosis is worth keeping: `?tls=True` is dj-email-url syntax, django-environ takes TLS from the scheme, and the parameter is dropped into `OPTIONS` where nothing reads it. If we ever want the provider's URL to work verbatim, the wrapper here does it.

Two things to fix before reviving it:
- `email_config` raises `ImproperlyConfigured` at settings-import time, so a bad parameter would take the whole site down to protect notification email. It should drop unreadable parameters, not refuse to boot.
- The check job runs on PRs touching its own file, so after merge it would never run again. It needs `schedule:` to be worth keeping, since the secret changes outside the repo.

PR #739, opened 2026-07-31
Edits/speed up pylint

fancysnake:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency when displaying and managing nested venue trees.
  * Preserved venue details, parent relationships, actions, and tree navigation across venue operations.
  * Added coverage confirming timetable session cards support dragging and expose duration data correctly.

* **Documentation**
  * Added guidance for organizing reusable test setup and assertions.

* **Refactor**
  * Standardized panel access checks and shared test setup across administrative workflows.
  * Simplified and consolidated test data construction for more consistent validation.


fancysnake:
## Deep code-quality review

Short version: **the test refactor is excellent and I'd merge it as-is. The tooling and DTO changes riding along with it are not, and one of them silently removes a repo-wide quality gate.** I'd split this PR.

---

### What's good, and it's most of the diff

The test dedup is the right kind of refactor, and it's rare to see one this disciplined:

- **−5,500 lines with no assertion loss.** Occurrence counts of `contains=` (163) and `context_data=` (549) across `tests/` are *identical* to `main`. Nothing was quietly dropped under cover of "dedup".
- **`panel_context(event, active_nav="cfp")` factors out only the sidebar keys** and leaves page-specific context spelled out literally at each call site. That respects the "build the full expected context, `ANY` only for forms" rule instead of collapsing everything into an opaque helper.
- **`test_proposal_status_actions.py` is the code-judo move of the PR** — four near-identical files (29 tests) become one parametrized module (32 cases). More coverage, a quarter of the code, and the `StatusAction` dataclass makes the accept/hold/reject/pending differences readable in one screen.
- Named assert helpers (`assert_not_a_manager`, `assert_scheduled_proposal_refused`) read better than the `assert_response(...)` blocks they replace.
- Every touched file *shrank*. Nothing crossed 1k lines. Full suite green: 3472 passed, 9 skipped.

Now the parts I want changed.

---

### 🔴 Blocker: `tingle check || true` disables the debt gate repo-wide — and it is silencing this PR

```toml
[tasks."lint:tingle"]
run = ["tingle check || true"]
```

Running the gate on this branch:

```
type-object (regex_count): +2
  + src/ludamus/pacts/venues.py:46
  + src/ludamus/pacts/venues.py:46
check failed: metrics grew by a net +2 (policy: sum)
```

Those `+2` are the `value: object` and `-> object` on the `_drop_related_manager` validator this same PR adds. **The PR takes on the debt and disables the detector in the same diff.**

Even setting aside that specific coupling: `|| true` doesn't skip the check for this branch, it removes it for everyone forever, buried in a PR titled "speed up pylint". Every future branch silently loses it. If the intent was "tingle is a warning, not a merge blocker", that's a policy decision that belongs in its own PR with `tingle report` still visible — not a shell-level `|| true` that makes the task structurally incapable of failing.

Please drop this line. Fix the `object` signature or accept the red on this branch; don't neuter the check for the repo.

### 🔴 Blocker: the `SpaceNodeDTO` change is a net loss — revert it

Commit b4d4594 says *"build SpaceNodeDTO from the model like every other DTO"*. But `SpaceTreeRepository._node` — the **other** builder for the same DTO, in the same class, ~180 lines below — still constructs all 14 fields explicitly. So the class now has *two* idioms for one DTO where it previously had one. The stated consistency win didn't materialize, and here's what it cost:

**1. The DTO contract now lies.**
```python
depth: int = 1
is_leaf: bool = True
```
`depth` and `is_leaf` are tree facts a single `Space` cannot know — the comment says so outright. Making them optional means any future call site that forgets them gets a silently-wrong "childless root" instead of a validation error. That's an explicit boundary traded for a silent fallback.

**2. `_drop_related_manager` is a magic validator serving one caller.**
```python
@field_validator("children", mode="before")
def _drop_related_manager(cls, value: object) -> object:
    return value if isinstance(value, list) else []
```
This exists solely because `Space.parent` has `related_name="children"`, so `from_attributes` grabs the reverse manager. It lives on the *shared* DTO and silently coerces any non-list to `[]` on **every** construction path, forever — a whole category of "why is my children list empty" bugs, to save one repository function. And it's `object -> object`, which is the type-escape-hatch tingle just flagged.

**3. `model_copy(update=...)` skips validation entirely.** The four fields that actually vary per node are now the four that no longer get validated.

**4. It's slower on a recursive tree walk** — two model instantiations plus a dict merge per node, replacing one constructor call.

The code-judo move here is subtraction: the original explicit constructor was 14 boring lines that were direct, validated, honest about the contract, matched `_node` right below it, and cost zero debt. Revert the DTO half of b4d4594 and **the tingle blocker above dissolves with it.**

If you do want the `from_attributes` consistency later, the honest shape is a split — a plain `SpaceDTO` for the columns a `Space` actually owns, and `SpaceNodeDTO(SpaceDTO)` adding the tree fields as required. But that's a separate PR, and I'd want to see it convert `_node` too.

---

### Weakened duplicate detection on `src`

```diff
-min-similarity-lines=4
+min-similarity-lines=10
```
```diff
-{ task = "lint:pylint", args = ["src"] },
-{ task = "lint:pylint", args = ["--disable=duplicate-code", "tests"] },
+{ task = "lint:pylint" },
```

Collapsing to one invocation forces one threshold across both trees. Net effect: `tests` goes from *duplicate-code entirely off* to *on at 10* (a real improvement), while `src` goes 4 → 10 (a real regression, in the tree where `duplicate-code` actually earns its keep).

The speedup here came from `-j 0`, not from the threshold. Keeping the two invocations and adding `-j 0` to each gets you the parallelism *and* keeps `src` at 4:

```toml
{ task = "lint:pylint", args = ["-j", "0", "src"] },
{ task = "lint:pylint", args = ["-j", "0", "--disable=duplicate-code", "tests"] },
```

Did you measure how much the 4 → 10 bump contributes on its own? If it's marginal, `src` shouldn't pay for it.

### `ars` typo silently drops `-x` from devcheck

```toml
{ task = "test:py", ars = [ "-x" ] },
```

`mise task info devcheck` confirms mise ignores the unknown key without complaint — the run line is a bare `task: test:py`. So `devcheck` now runs the whole suite with no fail-fast, and lost the old `-vv` too. One-character fix, but worth catching before it wastes someone's afternoon.

### Smaller tooling notes

- **`lint:pylint:diff` description** is copy-pasted from `lint:pylint` and claims "duplicate-code disabled for tests" — no longer true of either task.
- **`lint:pylint:diff` diffs against `main` (two-dot)**, i.e. worktree vs main's tip, so files changed *on main* since your branch point get linted too. `main...` (merge-base) is what you want. (Empty changesets are fine — bare `pylint` exits 0 with "No files to lint".)
- **`.pylintrc` still has `jobs=1`** while the task passes `-j 0`. Set it once in `.pylintrc` and drop the flag.
- **`.taplo.toml`: removing `inline_table_expand = false`** has nothing to do with pylint speed, and you can see the churn it caused right in this diff — the `format:black` args block went from one line to four. If it's needed, say why; if it's incidental, put it back.

---

### Recommendation

Split this. The test refactor is genuinely good work and should land on its own merits. Before it does:

1. Drop `|| true` from `lint:tingle`.
2. Revert the `SpaceNodeDTO` / `venues.py` change (this also clears #1).
3. Fix `ars` → `args`.
4. Restore `min-similarity-lines=4` for `src` via the two-invocation form, or show that the bump is load-bearing for the speedup.

fancysnake:
## Review triage

Went through every item in the deep review above. Two were already fixed in 6eff220 (pushed ~27 min after the review landed); the rest are validated below with measurements.

### ✅ Done — fixed in 6eff220

- **`tingle check || true`** — reverted, `lint:tingle` is back to `run = ["tingle check"]`. The gate is intact for the repo.
- **`ars` → `args` in `devcheck`** — fixed, `devcheck` fail-fasts again.

### 🔴 P1 — blocks merge

- **`SpaceNodeDTO` `from_attributes` change turns the `tingle` CI check red.** Confirmed locally:
  ```
  type-object (regex_count): +2
    + src/ludamus/pacts/venues.py:46
  check failed: metrics grew by a net +2 (policy: sum)
  ```
  and the `tingle` job is the one red check on this PR. The `object -> object` signature on `_drop_related_manager` is the whole delta.

  The supporting critique holds up: `SpaceTreeRepository._node` still builds all 14 fields explicitly, so the class now has two idioms for one DTO — the stated consistency win didn't land. `depth: int = 1` / `is_leaf: bool = True` are defaults a single `Space` can't honestly supply, `model_copy(update=...)` skips validation on exactly the four fields that vary, and the validator coerces silently on every construction path to serve one caller. Reverting the `venues.py` half of b4d4594 clears the red check and costs nothing else in the PR.

### 🟡 P2 — worth fixing here

- **`min-similarity-lines` 4 → 10 on `src`.** Accepted, and the suggested two-invocation form is better on *both* axes than assumed. Measured on this branch with `-j 0`:

  | invocation | wall |
  |---|---|
  | `pylint -j 0 src tests` (current, threshold 10) | **2m55s** |
  | `pylint -j 0 --min-similarity-lines=4 src` | 0m40s |
  | `pylint -j 0 --disable=duplicate-code tests` | 0m29s |
  | `pylint -j 0 tests` (threshold 10) | 1m16s |

  So the split form is **1m09s** (tests dup off) or **1m56s** (tests dup at 10) — both faster than the 2m55s combined run, while `src` keeps threshold 4. The speedup wasn't the 4→10 bump at all; it was `-j 0` plus not cross-comparing 1000 files of `src`×`tests` for duplicates. `src` at 4 is clean today. Recommend reverting `.pylintrc` to `min-similarity-lines=4` and going back to two invocations.

- **`.pylintrc` `jobs=1` vs `-j 0` on the task.** Confirmed. It matters more than cosmetics: `lint:pylint:diff` does *not* pass `-j 0`, so it runs single-threaded. Set `jobs=0` in `.pylintrc` once and drop the flag from both tasks.

- **Stale descriptions.** Both `lint:pylint` and `lint:pylint:diff` still say "duplicate-code disabled for tests", which is true of neither now.

- **`lint:pylint:diff` uses two-dot `git diff ... main`.** Real, currently latent — `main` was merged into this branch, so merge-base and `main` tip resolve to the same 83 files right now. It will lint unrelated files as soon as `main` moves ahead. Use `$(git merge-base main HEAD)`. (Verified the empty-changeset case is safe: bare `pylint` prints "No files to lint" and exits 0.)

- **`devcheck` lost `-vv`.** Minor, but it was there before and the new form only restores `-x`.

### 🟠 P3 — real, but not in this PR

- **Split `SpaceDTO` / `SpaceNodeDTO(SpaceDTO)`** with tree fields required, and convert `_node` too. That's the honest shape if the `from_attributes` consistency is still wanted — but it's a separate change and needs both builders converted to be worth anything.
- **Splitting this PR** (test refactor vs tooling). The refactor is the bulk and is green; re-splitting a passing branch is churn now. Worth doing on the next tooling change instead.

### ⚪ P4 — no action

- **`.taplo.toml` removing `inline_table_expand = false`.** Cosmetic only; the visible cost is a 3-line expansion of one `args` block. Not a defect, no behaviour change — leaving as-is unless there's a reason to restore it.
- **CodeRabbit's rate-limit notice** — no findings in it, nothing to action.

*(No inline review threads exist on this PR, so there's nothing resolvable — all feedback arrived as this single top-level comment.)*

Issue #709, opened 2026-07-27
E2E: panel personal-data write-ins have no browser test


hasparus:
`templates/panel/_personal_data_fields.html` renders an organizer's view of a  <br>facilitator's personal data. It is included by three pages — facilitator  <br>create, facilitator edit, and the proposal form — and until zagrajmy/ludamus#701 its  <br>multi-value branch rendered no write-in input at all, while the inputs it did  <br>render came back without a `value`. With the write-in merge zagrajmy/ludamus#701 adds, opening  <br>a facilitator and pressing Save wiped whatever they had written in.

zagrajmy/ludamus#701 fixes the template and covers the **POST** round trip in an integration  <br>test. What no test covers is the half that makes a browser send the value  <br>back: the `value="{{ current_value|custom_values:field }}"` attribute on the  <br>three inputs. Deleting it today turns the suite green while silently  <br>reintroducing the data loss.

## Scope

A Playwright spec against the panel, in the spirit of `write-in-fields.spec.ts`:

* a facilitator with a stored multi-value personal answer that mixes a  <br>configured option with a write-in
* open the facilitator edit page, confirm the write-in comes back in the  <br>companion input rather than as an empty box
* save without touching anything, confirm the stored answer is unchanged —  <br>the regression that motivated this

The proposal form includes the same partial, so covering one page guards the  <br>attribute for all three; picking the facilitator edit page keeps the spec  <br>short.

## Notes

* Needs seed data: `bootstrap_data.py` has no personal-data field with  <br>`is_multiple` + `allow_custom`, nor a stored value for one. Add both to the  <br>panel-lab event (`frostfire-con`) rather than an event other specs read.
* Pin the spec to chromium in `playwright.config.ts` — it mutates shared seed  <br>data, so a second project's copy would race it.
* Related: zagrajmy/ludamus#704 (client TypeScript modules with no browser coverage). Same  <br>root cause — the browser-facing half of the panel is thinly tested.

PR #752, opened 2026-08-01
Fix/panel proposal silent save failure

fancysnake:


fancysnake:

The bug fixes are right — a stored write-in really was coming back unsplit, and the
silent save failure really did need a form-level error. But the implementation leaves
more structure behind than it needed to, and one piece of it couples a template to an
invariant nothing enforces. Requesting changes.

---

### 1. Every single caller of `dynamic_fields_form` now does the same unfold dance. Push it into the callee.

After this PR, `dynamic_fields_form` has exactly three call sites, and all three look
like this:

```python
dynamic_fields_form(
    prefix=prefix,
    fields=[(field, False) for field in fields],
    data=data,
    # A write-in is stored as one value; it comes back split across the
    # control and its companion input, as the proposal wizard renders it.
    initial=unfolded_initial(prefix=prefix, fields=fields, stored=stored),
)
```

Same comment, verbatim, three times. Same prefix threaded through twice. Same helper
called with the same arguments the form builder already has.

When 100% of callers do X before calling Y, X belongs inside Y. `dynamic_fields_form`
already knows the prefix and the fields — everything `unfold_custom_answers` needs.
Make `initial` slug-keyed (the form owns the prefix anyway) and unfold internally:

```python
def dynamic_fields_form(
    *,
    prefix: str,
    fields: Sequence[tuple[OrganizerFieldDTO, bool]],
    data: QueryDict | None = None,
    initial: Mapping[str, FieldValue] | None = None,  # keyed by slug
) -> forms.Form:
    ...
    # A write-in is stored as one value; it comes back split across the
    # control and its companion input, as the wizard renders it.
    stored = initial or {}
    return form_class(
        data,
        initial=unfold_custom_answers(
            stored={f"{prefix}_{d.slug}": stored.get(d.slug) for d, _ in fields},
            fields=[d for d, _ in fields],
            prefix=prefix,
        ),
    )
```

The three call sites collapse to `initial=values or {}`, `initial=values or {}`, and
`initial={field.slug: current for field, current in ctx.session_fields}`. That deletes
`unfolded_initial` entirely, three copies of the comment, three prefixing
comprehensions, the new import in each of the three modules, and — see #3 — the type
widening. It also means the next page that renders organizer fields cannot forget to
unfold, which is exactly the bug this PR is fixing.

Only three callers exist, all of them in this PR's blast radius, so the parameter's
change of meaning costs nothing.

### 2. `unfolded_initial` fabricates DTOs to satisfy a signature that never asks about them.

```python
requirements=[
    PersonalFieldRequirementDTO(field=field, is_required=False)
    for field in fields
],
```

The comment concedes the problem: *"The requirement wrapper only carries the field:
unfolding asks nothing about who has to answer."* Right — so the signature is wrong,
not the caller. `unfold_custom_answers` reads `req.field.slug`, `.field_type`,
`.options`, `.is_multiple`, `.allow_custom`. `fold_custom_answers` reads
`.field.slug`, `.offers_custom_input`, `.allow_custom`, `.is_multiple`. Neither ever
touches `is_required`.

Change both to `fields: Sequence[OrganizerFieldDTO]`; the four wizard call sites pass
`[req.field for req in requirements]`. That deletes the fake objects, and lets
`PersonalFieldRequirementDTO` go back under `TYPE_CHECKING` — this PR promoted it to a
runtime import purely to construct throwaway wrappers, which is a real (if small) cost
paid for nothing.

### 3. Don't widen a shared helper's type to accommodate a caller.

```python
-    initial: Mapping[str, FieldValue] | None = None,
+    initial: Mapping[str, FieldValue | int] | None = None,
```

`initial` values are never ints. The `| int` leaked in from `WizardData`, which carries
ints for wizard state like time-slot ids that never reach this form. The contract now
says something untrue about what a form field can be initialised with, and the next
reader has to go find out why. With #1 and #2 applied this reverts to
`Mapping[str, FieldValue]` on its own.

### 4. The template guesses which card is broken from a global signal.

```django
{# Open when the page says something is wrong: the only #}
{# form-level error an edit can raise is this card's. #}
<details ... {% if form.non_field_errors %}open{% endif %}>
```

This opens *every* facilitator's card whenever the proposal form has *any* form-level
error, on the strength of an invariant that nothing enforces and that this very PR
already erodes — `_create` adds a second non-field error (`"Please select at least one
facilitator."`) three methods away. The comment's escape hatch is the word "edit"; the
template does not know whether it is rendering an edit.

The data to do this properly is already in hand. `field_descriptors` puts `errors` and
`custom_errors` on every `FieldAnswer`. Carry a `has_errors` alongside the prefix in
`FacilitatorPersonalData` (and while you're there, that 3-tuple is at its limit — a
small frozen dataclass reads better than a 4-tuple unpacked in a `{% for %}`), then:

```django
<details ... {% if entry.has_errors %}open{% endif %}>
```

Opens the card that is actually wrong, stays correct on the create page, stays correct
when someone adds a `clean()` to the proposal form, and drops the load-bearing comment.

Also: this behaviour is markup, and per CLAUDE.md markup claims belong in `tests/e2e` —
right now the `open` attribute is asserted nowhere.

### 5. The `_create` branch is unreachable, and untested.

```python
personal_data_valid = self._personal_data_is_valid(current_event.pk)
if not form.is_valid() or not facilitator_ids or not personal_data_valid:
    ...
    if not personal_data_valid:
        form.add_error(None, PERSONAL_DATA_ERROR)
```

`_render` sets `facilitator_personal_data` to `[]` when `session is None`, which is
exactly the `_create` case. The create page never renders a personal-data card, so
`personal_data_submitted` is never posted, so `_personal_data_forms` returns `None`, so
`_personal_data_is_valid` is unconditionally `True`. The branch cannot fire through the
UI — and the test diff only covers the edit path, which is consistent with that.

If a hand-crafted POST does force it, the outcome is worse than silence: an error
saying "fix the facilitator personal data below" above a page with no personal data
below. Either drop the check on the create path, or render the cards on create too and
make the message true. Symmetry between `_create` and `_update` isn't a reason to keep
a branch that can't run.

### 6. Rendering and validation disagree on which facilitators are in play.

Pre-existing, but this PR builds user-facing error reporting on top of it, so it is now
load-bearing:

- `_get_facilitator_personal_data_post` builds bound forms for the facilitators
  **assigned to the proposal**.
- `_personal_data_forms` builds bound forms for **posted ids ∩ event facilitators**.

A posted id that is valid for the event but not assigned to this proposal gets
validated (and written by `_collect_personal_data`) while never being rendered. Fail
validation on one of those and the save is blocked by a card that isn't on the page —
the same dead end this PR set out to fix. These should be one set derived once, not two
expressions that happen to agree in the common case. The per-card fix in #4 does not
cover this.

### 7. The new test bypasses the canonical matcher.

```python
        )
        form = response.context["form"]
        assert form.non_field_errors() == [...]
```

`assert_response` with `"form": ANY` followed by a manual assertion is the pattern
`FormErrorsMatcher` exists to replace — it's already imported in
`test_facilitator_edit_page.py` and used four times in the suite. Use
`"form": FormErrorsMatcher(__all__=["Fix the facilitator personal data below before saving."])`
and drop the trailing assertion. (The rest of the tests are good: full expected
context, exact DTOs, no HTML assertions, no `ANY` on simple values.)

### Nits

- `PAPERCUTS.md` gains a duplicate of the ssh_config entry immediately below it — same
  date, same cause, same workaround.
- `PERSONAL_DATA_ERROR` is wedged between `collect_session_field_values` and
  `_PROPOSALS_PAGE_SIZE`; it belongs with the other module constants.

---

**Verdict:** changes requested. #1 and #2 together delete a helper, a runtime import, a
type widening and three copies of the same comment while making the fix
unforgettable-by-construction — that's the version of this PR that should merge. #4 and
#5 are the parts that will bite someone later.

fancysnake:
## Review triage

Went through every point against the code. Verdicts below; nothing marked done yet —
these are the accept/reject calls, fixes follow.

| # | Verdict | Priority |
|---|---------|----------|
| 1 — unfold inside `dynamic_fields_form` | accepted | P2 |
| 2 — `unfold/fold` should take fields, not requirements | accepted, subsumed by #1 | P3 |
| 3 — `\| int` widening | accepted, falls out of #1 | P2 |
| 4 — template opens every card | accepted | P2 |
| 5 — `_create` branch dead through the UI | accepted (with a correction) | P2 |
| 6 — render/validate facilitator sets disagree | accepted, deferred | P3 |
| 7 — manual assertion instead of `FormErrorsMatcher` | accepted | P2 |
| nit — PAPERCUTS duplicate | accepted | P2 |
| nit — `PERSONAL_DATA_ERROR` placement | rejected | P4 |

**#1 confirmed.** `dynamic_fields_form` has exactly three production call sites
(`panel/views/facilitators.py`, `panel/views/proposals.py`,
`chronology/views.py::SessionEditView._fields_form`) and after this PR all three call
`unfolded_initial` with the same arguments and the same comment. The wizard's own
`unfold_custom_answers` call sites build forms through `build_personal_data_form` /
`build_session_details_form` / `form_class`, not through `dynamic_fields_form`, so
folding the unfold into the callee cannot double-apply. The unit tests in
`tests/unit/gates/web/django/test_dynamic_fields.py` pass no `initial` at all.

**#3 confirmed and it is a consequence of #1.** The widening exists only because
`unfolded_initial` returns `dynamic_fields.WizardData`, which is
`dict[str, FieldValue | int]` — the `int` is wizard state (`category_id`,
`time_slot_ids`), never a form initial. Reverts on its own once #1 lands.

**#2 accepted but downgraded.** `unfold_custom_answers` reads only
`req.field.{slug,field_type,options,is_multiple,allow_custom}` and
`fold_custom_answers` only `req.field.{slug,offers_custom_input}` — neither touches
`is_required`, so the signature really is asking for more than it uses. But the
fabricated `PersonalFieldRequirementDTO`s and the runtime import both disappear with
#1 alone. What is left is a standalone signature cleanup across four wizard call
sites, worth its own change.

**#5 confirmed, with one correction.** `_render` sets `facilitator_personal_data` to
`[]` when `session is None`, and the `personal_data_submitted` hidden input lives
inside `{% if facilitator_personal_data %}` in `panel/proposal-form.html` — so the
create page never posts it and the branch cannot fire through the UI. It is not
strictly unreachable, though: a hand-crafted POST carrying
`personal_data_submitted=1` plus ids valid for the event does reach
`_personal_data_forms`, and then the page shows "fix the facilitator personal data
below" above a page with no cards below. Dropping the check on the create path is the
right call. It also explains the red codecov patch check — the two uncovered lines are
exactly this branch.

**#6 confirmed, deferring.** `_get_facilitator_personal_data_post` iterates
`sessions.read_facilitators(proposal_id)`; `_personal_data_forms` iterates
`submitted_ids & valid_pks` over the event's facilitators. The two agree for any POST
the page itself produces, so this needs a crafted request to bite today — but it is a
real divergence and the fix is one derived set, not a patch at either end. Separate
change.

**#4 confirmed.** Today the invariant happens to hold on the edit path — `_update`
adds only `PERSONAL_DATA_ERROR`, and the "select at least one facilitator" error is
`_create`-only, where no cards render. But it is enforced by nothing, and meanwhile
every facilitator's card springs open when one of them is wrong. Carrying `has_errors`
per entry is strictly better and drops the load-bearing comment.

**nit rejected:** `PERSONAL_DATA_ERROR` sits immediately above `_PROPOSALS_PAGE_SIZE`
and `SCHEDULED_FILTER` — that *is* the module-constant block. Moving it moves it
within the same block.

**CI:** `tingle` is red at `old-subdomain-loc +16`, which #1 should more than repay by
deleting `unfolded_initial` and three call-site comprehensions. Not treating it as a
gate on its own.

fancysnake:
## Review items resolved — 21871875

- [x] **#1 — unfold inside `dynamic_fields_form`.** Done. `initial` is slug-keyed
  and the form builder unfolds; `unfolded_initial` is gone, along with three
  copies of the comment, three prefixing comprehensions and the import in each
  of the three modules.
- [x] **#2 — fabricated requirement wrappers.** Done for `unfold_custom_answers`,
  which now takes the fields it actually reads; `PersonalFieldRequirementDTO` is
  back under `TYPE_CHECKING`. `fold_custom_answers` keeps the requirement
  signature for now — filed as #753.
- [x] **#3 — `| int` widening.** Reverted to `Mapping[str, FieldValue]`, as
  predicted, once #1 landed.
- [x] **#4 — the template guessing.** Done. Entries carry `has_errors` and became
  a small frozen `PersonalDataCard`; the card that is wrong is the card that
  opens. The 3-tuple's middle element went with it — every consumer read
  `descriptor.name_prefix`, nothing read the entry's own prefix. `has_errors`
  comes off `form.is_valid()` on the bound form rather than a re-scan of the
  descriptors. It is asserted in `context_data`; the rendered `open` attribute
  stays an e2e concern, in scope for #709.
- [x] **#5 — the `_create` branch.** Dropped, with a comment saying why the
  create page has no cards to point at. This also clears the codecov patch
  failure — those were its two uncovered lines.
- [ ] **#6 — the two facilitator sets.** Filed as #754 rather than fixed here.
  Choosing between "the proposal's assigned facilitators" and "the facilitators
  this POST is about to assign" is a product call, and the second answer pulls
  the create page back in.
- [x] **#7 — the canonical matcher.** Done:
  `FormErrorsMatcher(__all__=[...])` on the context, manual assertion dropped.
  The `session.title` check after it stays — that is state, not markup.
- [x] **nit — PAPERCUTS duplicate.** Removed.
- **nit — `PERSONAL_DATA_ERROR` placement.** Won't fix: it already sits directly
  above `_PROPOSALS_PAGE_SIZE` and `SCHEDULED_FILTER`, which is the module's
  constant block.

Net −21 lines. Full suite green (3529 passed, 9 skipped); mypy, ruff, pylint,
djlint, vulture clean. `tingle` still counts `old-subdomain-loc +9`, down from
+16 — what is left is the `PersonalDataCard` dataclass and the error constant,
both legacy-path additions I would rather keep than golf away.

PR #759, opened 2026-08-01
fix(print): one door card page per room and day

fancysnake:
Door cards listed every day of the event on a single A4 sheet per room. A card hangs on a door for one day, so break the page per (room, day) and repeat the event header, room name and capacity on each sheet.

## Summary by CodeRabbit

* **New Features**
  * Door cards now print as one A4 page per room and day, with entries grouped and ordered clearly.
  * Printed timetables and door cards now include event details and optional scope information on every page.
  * Improved screen previews with consistent A4 sizing, centering, spacing, and page boundaries.
  * Empty print views now display the document header alongside a clear empty-state message.

* **Bug Fixes**
  * Spaces without scheduled items no longer produce empty door cards.

* **Documentation**
  * Added documentation for a local linting issue affecting development environments.

### DIFF
--- a/PAPERCUTS.md
+++ b/PAPERCUTS.md
@@ -263,3 +263,8 @@ If you fix a papercut, remove it.
   /etc/ssh/ssh_config.d/20-systemd-ssh-proxy.conf' (symlink owned by
   nobody:nogroup); worked around with git -c credential.helper='!gh auth git-
   credential' push <https://github.com/zagrajmy/ludamus.git> HEAD:the-branch
+- 2026-08-01: mise run lint fails locally on lint:hk: oxlint can't build
+  src/ludamus/client/oxlint.config.ts because eslint-plugin-sonarjs isn't
+  installed in the local node_modules (CI is fine). Same failure on a clean
+  tree, so it's env drift, not a code problem — had to stash and re-run to prove
+  my change was innocent.
--- a/src/ludamus/mills/printing.py
+++ b/src/ludamus/mills/printing.py
@@ -19,7 +19,6 @@ from ludamus.pacts.printing import (
     AreaScheduleQueryDTO,
     AreaScheduleSessionDTO,
     AreaScheduleSpaceDTO,
-    DoorCardDayDTO,
     DoorCardDTO,
     DoorCardEntryDTO,
     DoorCardsDocumentDTO,
@@ -148,11 +147,9 @@ class PrintMaterialsService:
         for space in spaces:
             # Cards hang on doors for participants: a room with nothing
             # scheduled gets no card, and empty hours are simply not listed.
-            if not (space_items := items_by_space.get(space.pk)):
-                continue
             entries_by_day: dict[date, list[DoorCardEntryDTO]] = defaultdict(list)
 
-            for item in space_items:
+            for item in items_by_space.get(space.pk, []):
                 day = item.start_time.astimezone(query.tz).date()
                 entries_by_day[day].append(
                     DoorCardEntryDTO(
@@ -162,15 +159,15 @@ class PrintMaterialsService:
                     )
                 )
 
-            days = [
-                DoorCardDayDTO(
-                    day=day, entries=sorted(entries_by_day[day], key=_entry_start)
+            cards += [
+                DoorCardDTO(
+                    space_name=space.name,
+                    capacity=space.capacity,
+                    day=day,
+                    entries=sorted(entries_by_day[day], key=_entry_start),
                 )
                 for day in sorted(entries_by_day)
             ]
-            cards.append(
-                DoorCardDTO(space_name=space.name, capacity=space.capacity, days=days)
-            )
 
         return DoorCardsDocumentDTO(
             event_name=event.name,
--- a/src/ludamus/pacts/printing.py
+++ b/src/ludamus/pacts/printing.py
@@ -63,15 +63,12 @@ class DoorCardEntryDTO(BaseModel):
     session: PrintSessionDTO
 
 
-class DoorCardDayDTO(BaseModel):
-    day: date
-    entries: list[DoorCardEntryDTO]
-
-
+# One card is one sheet of paper: it hangs on a door for a single day.
 class DoorCardDTO(BaseModel):
     space_name: str
     capacity: int | None
-    days: list[DoorCardDayDTO]
+    day: date
+    entries: list[DoorCardEntryDTO]
 
 
 class DoorCardsDocumentDTO(BaseModel):
--- a/src/ludamus/templates/panel/print/base.html
+++ b/src/ludamus/templates/panel/print/base.html
@@ -1,5 +1,5 @@
 {% extends "base.html" %}
-{% load i18n %}
+{% load i18n tessera %}
 {% block title %}
     {{ document.event_name }}
 {% endblock title %}
@@ -16,7 +16,9 @@
         /* Document typography + white paper are scoped to the sheet so they
            never touch the site navbar/footer that now wrap this page, and so
            the sheet prints white regardless of the active light/dark theme.
-           One .sheet is one printed page. */
+           A .sheet starts a printed page; content taller than the paper still
+           spills onto an unheadered continuation page — nothing caps it yet
+           (#767). */
         .sheet {
             font-family: sans-serif;
             color: #111827;
@@ -50,26 +52,6 @@
         .empty {
             color: #9ca3af;
             font-style: italic;
-        }
-
-        .toolbar {
-            margin: 0 0 12px;
-        }
-
-        .print-btn {
-            font: 600 14px sans-serif;
-            padding: 8px 16px;
-            border: 1px solid #111827;
-            border-radius: 8px;
-            background: #111827;
-            color: #fff;
-            cursor: pointer;
-        }
-
-        .print-hint {
-            margin-left: 10px;
-            font: 14px sans-serif;
-            color: #6b7280;
         }
 
         @media screen {
@@ -116,9 +98,11 @@
         <div class="print-doc">
             {# The sheets repeat their header, so the document's one heading is here. #}
             <h1 class="sr-only">{{ document.event_name }}</h1>
-            <div class="toolbar no-print">
-                <button type="button" class="print-btn" data-action="print">{% translate "Print" %}</button>
-                <span class="print-hint">{% translate "or press Ctrl+P (⌘P on Mac)" %}</span>
+            {# Theme-aware tokens: the toolbar sits on the page, not on the white sheet. #}
+            <div class="no-print mb-3 flex flex-wrap items-center gap-3">
+                {% translate "Print" as t_print %}
+                {% tessera_button t_print button_type="button" icon="printer" data_action="print" full_width_mobile=False %}
+                <span class="text-sm text-foreground-secondary">{% translate "or press Ctrl+P (⌘P on Mac)" %}</span>
             </div>
             {% block body %}
             {% endblock body %}
--- a/src/ludamus/templates/panel/print/base.html
+++ b/src/ludamus/templates/panel/print/base.html
@@ -15,7 +15,8 @@
 
         /* Document typography + white paper are scoped to the sheet so they
            never touch the site navbar/footer that now wrap this page, and so
-           the sheet prints white regardless of the active light/dark theme. */
+           the sheet prints white regardless of the active light/dark theme.
+           One .sheet is one printed page. */
         .sheet {
             font-family: sans-serif;
             color: #111827;
@@ -72,11 +73,21 @@
         }
 
         @media screen {
-            .sheet {
+            .print-doc {
                 max-width: 210mm;
                 margin: 16px auto;
+            }
+
+            /* On paper the page break draws the boundary; on screen only this
+               does, so the preview shows where each sheet starts and ends. */
+            .sheet {
+                min-height: 297mm;
                 padding: 14mm;
                 box-shadow: 0 1px 6px rgba(0, 0, 0, 0.18);
+            }
+
+            .sheet+.sheet {
+                margin-top: 16px;
             }
         }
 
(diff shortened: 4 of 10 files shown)


fancysnake:

**Verdict: changes requested.** The behavior is right and I verified the new break
condition is logically correct (`not inner.last or not parent.last` == `not (inner.last
and parent.last)`, and a card always has ≥1 day because `build_door_cards` skips spaces
with no items). But the fix landed in the wrong layer: the template is now compensating
for a DTO whose grain no longer matches the thing being printed. There's a code-judo move
here that deletes the nested loop, the compound conditional, and a whole DTO class.

---

### 1. (blocker) The page unit changed — the data model didn't

The PR's own comment states the invariant: *"one A4 page per room and day: a card hangs on
the door for one day."* So the printed unit is now **(space, day)**. But
`DoorCardsDocumentDTO.cards` is still `list[DoorCardDTO]` where a `DoorCardDTO` is a space
holding `days: list[DoorCardDayDTO]` — one-per-space. The template papers over the
mismatch with a nested loop plus `forloop.parentloop`, which is exactly the smell: a
compound cross-loop conditional is what you write when the collection you're iterating
isn't the collection you're rendering.

Flatten the DTO so one `DoorCardDTO` *is* one card:

```python
class DoorCardDTO(BaseModel):
    space_name: str
    capacity: int | None
    day: date
    entries: list[DoorCardEntryDTO]
```

`DoorCardDayDTO` is then deleted outright. In `PrintMaterialsService.build_door_cards`
the `if not (space_items := items_by_space.get(space.pk)): continue` guard also goes —
an empty `entries_by_day` naturally contributes no cards:

```python
for space in spaces:
    entries_by_day: dict[date, list[DoorCardEntryDTO]] = defaultdict(list)
    for item in items_by_space.get(space.pk, []):
        day = item.start_time.astimezone(query.tz).date()
        entries_by_day[day].append(DoorCardEntryDTO(...))
    cards += [
        DoorCardDTO(
            space_name=space.name,
            capacity=space.capacity,
            day=day,
            entries=sorted(entries_by_day[day], key=_entry_start),
        )
        for day in sorted(entries_by_day)
    ]
```

And the template collapses to a single flat loop — no `parentloop`, no nesting, no
conditional. The name `cards` becomes *more* honest than it is today: a card is the paper
that hangs on one door for one day, and the list is now literally the print run in order.
The "one page per card" invariant lives in the data instead of being reconstructed by
template arithmetic.

Net: −1 DTO class, −1 nested loop, −1 `continue` guard, −1 compound conditional, and the
model finally describes the artefact.

### 2. (blocker) There's already a canonical page-break helper — this hand-rolls a third copy

`chronology/print.html` (the public print page) already does this with no template logic
at all:

```
print:break-after-page last:print:break-after-auto
```

Meanwhile `panel/print/door-cards.html` and `panel/print/timetable.html` both hand-roll
`<section {% if not forloop.last %}style="break-after: page;"{% endif %}>`, and this PR
makes the door-cards copy strictly harder to read. CLAUDE.md is explicit — prefer Tailwind,
use the design system, don't hand-roll. The utilities are already emitted (the public page
uses them), and the sections are direct children of `.sheet`, so `:last-child` resolves
correctly:

```html
<section class="print:break-after-page last:print:break-after-auto">
```

That's the whole fix. The conditional disappears, and it stays correct no matter how the
loops are later restructured — which is the real point, since this PR is the second time
the condition has had to be re-derived. Worth applying to `timetable.html` in the same
pass; it's the identical line.

Combined with finding 1, the template body reduces to one loop over one flat list with
zero page-break logic in it.

### 3. The doc-header is now copy-pasted markup inside a loop

Moving `.doc-header` inside the per-page `<section>` is correct — a card that doesn't name
its event is useless on a door. But that block is now byte-identical in two panel print
templates and repeated per page in one of them, and the public side already solved this
with `chronology/parts/print_header.html`. Extract `panel/print/parts/doc-header.html` and
`{% include %}` it from both. Small, but this is the moment where duplication went from
"twice" to "twice and in a loop", and the next print material makes it three.

### 4. A behavior change shipped with no test — and invalidated an existing one

The diff is one template file. Nothing asserts the new pagination anywhere.

Worse, `tests/e2e/tests/print-flow.spec.ts` has a test literally named **"door cards
render one card per room with its hours"** — the exact claim this PR reverses. It still
passes because it only checks a heading and "Capacity: 18" are visible, but the name now
documents the old behavior. Rename it to "one card per room and day" and assert the page
count: for a room with sessions on two days, two sections carrying that room's name. Per
the house rules that assertion belongs in e2e, not in an integration test.

`tests/integration/web/panel/test_timetable_print.py::test_door_cards_limited_to_time_window`
has to be updated for the flattened DTO anyway — it constructs the full nested
`DoorCardDTO(days=[DoorCardDayDTO(...)])` shape. While in there:
`test_door_cards_page_for_sphere_manager` uses `context_data={"document": ANY}` plus three
raw `in response.content` substring checks. Both are against the project's own testing
rules (no whole-document `ANY`, no HTML assertions in integration tests). Pre-existing, but
you're editing this exact behavior — it's the cheap moment to fix it.

### 5. (nit) Undocumented spacing change

`margin: 0 0 6mm 0` → `0 0 1mm 0` on the capacity line is bundled in with no mention in the
commit message. Presumably intentional now that the header repeats and vertical space is
tighter per page, but it's an unrelated visual change riding along in a pagination fix.
One line in the commit body would do.

---

**Not a concern:** file size (43 lines), and the break condition itself is correct — I
checked all four loop-position combinations.

**What I'd want to see:** flatten `DoorCardDTO` to one-card-per-(space, day), swap both
panel templates to `print:break-after-page last:print:break-after-auto`, extract the
doc-header include, and update the e2e test's name and assertion. That's a smaller diff
than the current one in the template, deletes a DTO and a service branch, and leaves the
next print material with less to copy.

fancysnake:
## Thermo-nuclear code quality review (round 2)

**Verdict: changes requested.** The four blockers from the last round are genuinely
fixed — `DoorCardDTO` is flat and one card is one sheet, `DoorCardDayDTO` is gone, the
`continue` guard is gone, both templates use `print:break-after-page
last:print:break-after-auto`, the doc-header is an include, and the tests were rewritten
to the flattened shape (including dropping the `context_data={"document": ANY}` +
`in response.content` violations). Full Python suite green locally: 3531 passed, 9
skipped. The mills diff is a straight deletion and I have no notes on it.

But the fix pulled the panel print stack *halfway* toward the one that already exists,
and stopped. What landed is a second, hand-written implementation of a sheet abstraction
this repo already ships — including two utilities lifted verbatim from it. And moving the
toolbar out of the sheet broke it in dark mode.

---

### 1. (blocker) This re-derives `chronology/print.html`'s sheet in a `<style>` block

`chronology/print.html` has had a printed-sheet abstraction the whole time:

```
box-border shrink-0 snap-start w-[210mm] max-w-full min-h-[297mm] rounded-sm bg-white
p-[14mm] font-sans text-[10pt] text-neutral-900 shadow-[...] print:m-0 print:max-w-none
print:min-h-[297mm] print:w-[210mm] print:rounded-none print:p-[14mm] print:shadow-none
print:[print-color-adjust:exact] print:break-after-page last:print:break-after-auto
```

`panel/print/base.html` now declares the same thing in CSS:

| `chronology/print.html` | `panel/print/base.html` after this PR |
|---|---|
| `bg-white` | `.sheet { background: #fff }` |
| `font-sans` | `.sheet { font-family: sans-serif }` |
| `text-[10pt]` | `.sheet { font-size: 10pt }` |
| `text-neutral-900` | `.sheet { color: #111827 }` (= neutral-900) |
| `min-h-[297mm]` | `@media screen { .sheet { min-height: 297mm } }` |
| `p-[14mm]` | `@media screen { .sheet { padding: 14mm } }` |
| `shadow-[0_1px_10px_...]` | `@media screen { .sheet { box-shadow: 0 1px 6px ... } }` |
| `w-[210mm]` | `@media screen { .print-doc { max-width: 210mm } }` |
| `print:break-after-page last:print:break-after-auto` | *copied verbatim onto `class=`* |

That last row is the tell. The PR reached into the canonical component, took the two
utilities it needed, and left the other eight to be re-declared by hand in a different
styling system — on the same element. `.sheet` is now styled by a `<style>` block **and**
by Tailwind variants, which is the worst of both: you can't read the sheet's rules in one
place, and the two definitions will drift (the shadows already have).

CLAUDE.md is explicit: *"Tailwind = component look. Partials in `templates/components/`;
`extra_class` for page layout only. No utility copy-paste into CSS."*

The judo move is to stop having two printed-sheet implementations. One
`templates/components/print/sheet.html` partial (`{% include %}` with a `body` block, or a
`{% with sheet_class=... %}` if that's simpler), used by `chronology/print.html` and both
panel templates. That deletes the entire `<style>` block from `panel/print/base.html`
except `@page` and the `@media print` chrome-hiding, deletes `.print-doc`'s screen rules
in favour of the wrapper the public page already has, and — the real payoff — the *next*
print material inherits a working sheet instead of copying a third one.

Worth pricing honestly: it's a bigger diff than what's here. If you'd rather not do it in
this PR, then at minimum don't ship a `.sheet` that's half CSS and half Tailwind — pick
one system for this element in this PR and open an issue for the merge.

### 2. (blocker) The toolbar left the white sheet and now fails contrast in dark mode

Before: `<div class="sheet">` wrapped the toolbar, and `.sheet` sets `background: #fff`.
After: the toolbar sits in `.print-doc`, which has no background, so it renders directly
on `--color-background`. In the dark theme that's `#0a0a0a`:

- `.print-hint` `#6b7280` on `#0a0a0a` → **4.10:1**. Fails AA for 14px text. It was
  4.83:1 on white — this PR flipped it from pass to fail.
- `.print-btn` is `background: #111827` with `border: 1px solid #111827` → **1.11:1**
  against the page. The button's white label is readable, but the control has no visible
  boundary at all; that's a 3:1 non-text contrast failure and it looks broken.

This is a hand-rolled button in the first place. CLAUDE.md says use the tessera design
system rather than hand-rolling components — swapping `.print-btn`/`.print-hint` for the
tessera button plus muted-text token fixes the contrast, deletes ~15 lines of CSS, and
makes the control theme-aware for free. Failing that, keep the toolbar on a surface.

### 3. "One .sheet is one printed page" is a comment, not an invariant

`base.html` now asserts it:

```css
/* ... One .sheet is one printed page. */
```

Nothing enforces it. A room with a dense day, or a timetable day with many distinct
session times, produces a `<section>` taller than 297mm; `break-after: page` fires at the
end, so the overflow lands on a second physical page **with no header** — which is the
exact defect this PR exists to fix. On screen it's invisible too: `min-height: 297mm` lets
the preview grow silently, so the one place that was supposed to "show where each sheet
starts and ends" shows a single tall rectangle that prints as two.

The codebase already has the right answer to "too much for one page": `_space_chunks` /
`MAX_TIMETABLE_SPACES_PER_PAGE` in `mills/printing.py` paginates in the service, where it
belongs. Door cards now paginate by `(space, day)` but have no row cap, and the timetable
paginates by space but not by row count. Either extend that pagination (a max-entries cap
that splits a day into "Sheet 1 of 2"), or make the header repeat on overflow
(`position: fixed` in `@media print` is the standard trick, and `<thead>` already repeats
for the timetable table). What I'd push back on is shipping a comment that promises more
than the code delivers — at minimum reword it to what's true.

### 4. `parts/doc-header.html` is the second copy of a partial that already exists

`chronology/parts/print_header.html` renders event name, event dates and `scope_name`
with the same border-bottom treatment. The new `panel/print/parts/doc-header.html` renders
event name, event dates and `scope_name` with the same border-bottom treatment. The only
real difference is the QR/logo block and the styling system.

They don't share context shapes today (`event` + `qr_svg` vs `document.*`), so this isn't
a one-line dedupe — but the partial takes parameters. Both `DoorCardsDocumentDTO` and
`PrintTimetableDocumentDTO` already carry `event_name` / `event_start` / `event_end` /
`scope_name`, so `{% include "…/print_header.html" with name=document.event_name … %}`
is a small change. Fixing the header duplication was finding 3 last round; the fix
deduplicated it within the panel and created a new duplicate across the boundary. Fold
this into finding 1 — one sheet, one header.

### 5. A full rewrite of `door-cards.html` and the inline `style=` attributes survived

Every line of the card body was rewritten in this PR, and `style="font-size: 26pt;
margin: 0 0 1mm 0;"` / `style="width: 28mm; padding: 2mm 3mm 2mm 0; …"` came along
unchanged. `chronology/print.html` renders the *same markup* as `text-[26pt] m-0 mb-[1mm]`
etc. This was the moment to convert — the diff was already touching those lines, and it
would have removed the second reason the two templates can't be merged.

### 6. Smaller notes

- `test_door_cards_page_for_sphere_manager` and `test_door_cards_limited_to_time_window`
  now assert nearly byte-identical `DoorCardsDocumentDTO`s. The first one's distinguishing
  claim ("Empty Hall gets no card") is real and is covered by the exact-equality assert,
  but a reader can't tell the two tests apart. A comment naming what each pins down, or
  merging them, would help.
- `expect(await cards.count()).toBeGreaterThan(0)` in the e2e is close to vacuous — it
  only guards the loop below from silently passing on an empty list. The seed data is
  known (three days, Miniature Painting on all of them), so assert the exact counts;
  `toBeGreaterThan(1)` on `roomHeadings` would still pass if a bug produced two sheets for
  one day.
- Landscape preview: `.print-doc { max-width: 297mm }` is ~1122px, wider than a lot of
  viewports, and there's no scroll container — the page will scroll sideways or crush the
  8pt grid. `chronology/print.html` wraps its sheets in an `overflow-x-auto` snap
  scroller. (Still an improvement on the 210mm it had before, so not a blocker.)

---

**What I'd want to see before merge:** fix the dark-mode toolbar (2) — that's a visible
regression this PR introduced, and it's small. Reword or enforce the one-sheet-one-page
claim (3). For (1)/(4)/(5), either do the merge into one shared sheet + header component,
or explicitly defer it in an issue and don't leave `.sheet` split across two styling
systems in the meantime. The behavior in this PR is right; the structure is one
copy-paste away from being right too.

PR #760, opened 2026-08-01
fix(timetable): show only leaf spaces in the overview heatmap

fancysnake:
build_heatmap rendered a column for every Space in the event, so venue and area branch nodes appeared as permanently empty rooms. Filter to leaves, matching the timetable grid and capacity_hours.

_leaves_in_tree_order moves from a TimetableService staticmethod to a module-level function now that two services use it.

### DIFF
--- a/src/ludamus/mills/timetable.py
+++ b/src/ludamus/mills/timetable.py
@@ -104,6 +104,25 @@ def _position_sessions(
     return positions
 
 
+def _leaves_in_tree_order(nodes: list[SpaceDTO]) -> list[SpaceDTO]:
+    children: dict[int | None, list[SpaceDTO]] = defaultdict(list)
+    for node in nodes:
+        children[node.parent_id].append(node)
+
+    leaves: list[SpaceDTO] = []
+
+    def walk(node: SpaceDTO) -> None:
+        if kids := children.get(node.pk, []):
+            for kid in kids:
+                walk(kid)
+        else:
+            leaves.append(node)
+
+    for root in children.get(None, []):
+        walk(root)
+    return leaves
+
+
 class TimetableService:
     def __init__(self, uow: UnitOfWorkProtocol) -> None:
         self._uow = uow
@@ -119,7 +138,7 @@ class TimetableService:
     ) -> TimetableGridDTO:
         all_nodes = self._uow.spaces.list_by_event(event_pk)
         node_name_by_pk = {node.pk: node.name for node in all_nodes}
-        leaf_spaces = self._leaves_in_tree_order(all_nodes)
+        leaf_spaces = _leaves_in_tree_order(all_nodes)
         if track_pk is not None:
             track_space_pks = set(self._uow.tracks.list_space_pks(track_pk))
             leaf_spaces = [
@@ -276,25 +295,6 @@ class TimetableService:
         )
 
     @staticmethod
-    def _leaves_in_tree_order(nodes: list[SpaceDTO]) -> list[SpaceDTO]:
-        children: dict[int | None, list[SpaceDTO]] = defaultdict(list)
-        for node in nodes:
-            children[node.parent_id].append(node)
-
-        leaves: list[SpaceDTO] = []
-
-        def walk(node: SpaceDTO) -> None:
-            if kids := children.get(node.pk, []):
-                for kid in kids:
-                    walk(kid)
-            else:
-                leaves.append(node)
-
-        for root in children.get(None, []):
-            walk(root)
-        return leaves
-
-    @staticmethod
     def _build_space_groups(
         spaces: list[SpaceDTO], name_by_pk: dict[int, str]
     ) -> list[SpaceGroupDTO]:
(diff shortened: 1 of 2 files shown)


fancysnake:

The fix is correct and the direction is right — lifting `_leaves_in_tree_order` out of
`TimetableService` to module scope is exactly where a pure tree helper belongs. But the PR
stopped one step short of the move that would have made it pay for itself.

### 1. There are now two definitions of "leaf space" rendered side by side on the same page

`build_heatmap` gets the helper plus a comment explaining that only leaves are bookable.
Twenty lines below, `capacity_hours` already carries that same rationale in prose — and its
own bespoke implementation:

```python
# Only leaf spaces are bookable rooms — a venue or area node would
# inflate capacity with slots nothing can ever fill.
spaces = self._uow.spaces.list_by_event(event_pk)
parent_pks = {s.parent_id for s in spaces if s.parent_id is not None}
rooms = [s for s in spaces if s.pk not in parent_pks]
```

`ScheduleOverviewView` calls **both** in one request, so the same page computes "which spaces
are rooms" twice, two different ways. They agree on well-formed trees and diverge when a node's
parent isn't in the returned list: the set-difference counts it as a room, the root-walk drops
it silently. `Space.parent` is a plain self-FK with no same-event constraint, so that state is
reachable — the page would then say "12 rooms" above a 11-column heatmap.

This is the code-judo move the PR left on the table:

```python
rooms = _leaves_in_tree_order(spaces)
```

Two lines and a duplicated comment deleted, one definition of "room" in the file, and the new
helper earns its promotion to module scope instead of just changing address. Please fold this in.

### 2. The helper re-implements `SpaceTreeRepository.list_tree`

`_leaves_in_tree_order` builds `children_by_parent` with a `defaultdict`, recurses, and walks
from `children[None]` — which is character-for-character what `list_tree` already does over the
same `order_by("order", "name")` queryset, and it produces `SpaceNodeDTO.is_leaf` while it's
there. `mills/venues.py::_leaf_pks` is a third walk over that same concept.

Not asking this PR to unify `links` and `mills` — `list_tree` also prefetches tracks, which the
heatmap doesn't need. But "which spaces are bookable rooms" now has three implementations and no
canonical home, which is precisely how a fourth one gets written. Worth an issue.

### 3. Duplicate per-request loads (pre-existing, cheap to fix if #1 lands)

`build_heatmap` loads spaces + agenda items, then `get_all_conflicts` →
`ConflictDetectionService._load_event_context` loads both again, then `capacity_hours` loads
them a third time. Not this PR's debt, but once both methods share the leaf helper, hoisting one
`list_by_event` into the view-facing call path is nearly free.

### 4. The test asserts the safe half only

```python
assert [s.pk for s in result.spaces] == [3, 4]
assert [c.space_pk for c in result.rows[0].cells] == [3, 4]
```

Every cell in this fixture is `EMPTY`, so the test still passes if the `space_pk_set` filtering
that follows the change breaks. Add one agenda item on space 3 and assert its cell comes back
`SCHEDULED` — that's the regression this change can actually cause. (The pk ordering assertion
also can't distinguish tree order from input order here, since the input is already sorted.)

Related, worth an explicit decision rather than an accident: an agenda item assigned to a
non-leaf space now vanishes from the map with no signal. `assign_session` blocks that via
`_require_space_in_event`, so it can only come from imports or legacy rows — and `capacity_hours`
already ignores them — but "silently invisible" is a poor failure mode for a page whose entire
job is showing what is scheduled.

---

**Verdict:** the behavior change is right and the test is real. Blocking on #1 — collapsing
`capacity_hours` onto the same helper is a smaller diff than the one already written, and leaving
two leaf definitions on one page after a PR whose whole subject is "what counts as a leaf" is the
wrong resting state. #2 and #3 are follow-ups; #4 is a small ask in this PR.

fancysnake:
Addressed in 1bbfddb.

**#1 — folded in.** `capacity_hours` now calls `_leaves_in_tree_order`; the
set-difference and its duplicated comment are gone, so the page has one
definition of "room".

One correction to the rationale: the divergence isn't reachable. Every `Space`
write runs `full_clean()` (`SpaceRepository` create and update), and
`_validate_same_event` rejects a cross-event parent; the one path that skips
`full_clean`, `_clone_subtree`, builds each clone's parent inside the target
event. So "12 rooms above an 11-column heatmap" needs raw SQL. The dedup is
still worth it on consistency grounds, just not as a bug fix.

**#4 — done.** The fixture had `agenda_items.list_by_event → []`, so every cell
was `EMPTY` and the `space_pk_set` filter was unasserted. The test now schedules
one item in room 3 and asserts `[SCHEDULED, EMPTY]`.

The "silently invisible agenda item" half is a non-issue: all three creation
paths are leaf-guarded — `assign_session` via `_require_space_in_event`,
`revert_change` replays an already-validated placement, and `accept_session`'s
form choices come from `read_space_options`, which filters
`children__isnull=True`. `Space._validate_leaf_parent` also blocks turning a
space that holds agenda items into a branch.

**#2 — #762.** It's five implementations, not three:
`_leaves_in_tree_order`, `printing._scoped_spaces`, `venues._leaf_pks`,
`SpaceTreeRepository.list_tree`'s `is_leaf`, and `read_space_options`'s ORM
filter. The issue lists them and the two candidate homes.

**#3 — #763.** Confirmed three loads each of spaces and agenda items per
overview request.

PR #764, opened 2026-08-01
fix(panel): lay the proposal form out with grid, not multicol

fancysnake:
The form was a CSS multi-column masonry. Every card carried break-inside-avoid, so a column the browser couldn't split overflowed the balanced column box: the footer was laid out after the (too short) container height and the tall column painted straight over it — theme switcher and copyright line ended up behind the facilitators card, with cards continuing below the end of the page.

Real grid columns instead: three at xl, two at lg, one below. The display:contents wrappers stay — they're only broken inside multicol — so the conditional cards keep flowing into the grid on their own. items-start keeps each card at its content height.

Lost with masonry: column height balancing. Rows align now, so a short card leaves whitespace under it. In exchange the cards no longer get shuffled between columns as the event's data changes.

The e2e regression pins three columns at 1440px, one at 480px, and that the form's box never reaches past the footer's top. Against the old markup it fails at the first assertion — multicol put the cards in two columns, not three.

## Summary by CodeRabbit

* **Improvements**
  * Updated the proposal form to use a responsive grid layout.
  * Improved spacing and alignment across proposal form sections.
  * The form now displays in three columns on desktop and collapses to one column on mobile.
  * Prevented the form from overlapping the footer.

### DIFF
--- a/src/ludamus/templates/panel/parts/proposal-session-fields.html
+++ b/src/ludamus/templates/panel/parts/proposal-session-fields.html
@@ -1,11 +1,10 @@
 {% load i18n %}
 {% load tessera %}
-<div id="proposal-session-fields"
-     class="space-y-6 break-inside-avoid mb-6">
+<div id="proposal-session-fields" class="space-y-6">
     <div class="card">
         <div class="card-body">
             <h3 class="text-lg font-semibold text-foreground mb-2">{% translate "Participants and duration" %}</h3>
-            {# Stacked, not columns: the card sits in a narrow masonry column, #}
+            {# Stacked, not columns: the card sits in a narrow grid column, #}
             {# where a two-up grid squeezes the duration steppers into a wrap. #}
             <div class="space-y-4">
                 <div>{% tessera_field form.participants_limit %}</div>
--- a/src/ludamus/templates/panel/parts/proposal-session-fields.html
+++ b/src/ludamus/templates/panel/parts/proposal-session-fields.html
@@ -1,5 +1,7 @@
 {% load i18n %}
 {% load tessera %}
+{# One grid cell on purpose: hx-swap="outerHTML" needs a real element to #}
+{# replace, so these cards stack together instead of each taking a cell. #}
 <div id="proposal-session-fields" class="space-y-6">
     <div class="card">
         <div class="card-body">
--- a/src/ludamus/templates/panel/proposal-form.html
+++ b/src/ludamus/templates/panel/proposal-form.html
@@ -45,10 +45,10 @@
         <form id="proposal-form"
               method="post"
               enctype="multipart/form-data"
-              class="p-4 lg:p-6 columns-1 lg:columns-2 xl:columns-3 gap-6">
+              class="p-4 lg:p-6 grid grid-cols-1 lg:grid-cols-2 xl:grid-cols-3 gap-6 items-start">
             {% csrf_token %}
             <div class="contents">
-                <div class="card break-inside-avoid mb-6">
+                <div class="card">
                     <div class="card-body">
                         <h3 class="text-lg font-semibold text-foreground mb-2">{% translate "Session Details" %}</h3>
                         <div class="space-y-6">
@@ -90,7 +90,7 @@
             </div>
             <div class="contents">
                 {% if all_facilitators %}
-                    <div class="card break-inside-avoid mb-6">
+                    <div class="card">
                         <div class="card-body">
                             <input type="hidden" name="facilitators_submitted" value="1">
                             <h3 class="text-lg font-semibold text-foreground mb-2">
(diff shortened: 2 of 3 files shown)


fancysnake:
## Code quality review

The fix direction is right — multicol was the wrong tool and grid is the boring, correct replacement. The template change is small and does what it says. Two things stop me approving: the regression test doesn't actually pin the regression, and the diff leaves behind two wrappers that no longer mean anything.

I reproduced the pre-fix markup shape in Chromium to check the test claims rather than reasoning about them.

---

### 1. The regression test asserts a property of the seed data, not of the change

`columnLefts` counts distinct `left` offsets of rendered `.card`s. That number is a function of how many cards the event happens to produce, not of the grid declaration.

I built the pre-fix shape — `columns:3`, two `display:contents` wrappers, five cards with `break-inside-avoid` — and Chromium put them in **three** distinct lefts. The assertion would have passed on the broken markup at that card count. It reads 2 today only because `frostfire-con` currently renders a particular set of conditional cards, and `proposal-duration.spec.ts` already warns in a comment that `panel.spec` mutates this event as the suite runs.

Assert the thing the PR changed instead. Data-independent, and it reads `none` → 0 on the old markup at every viewport:

```ts
const columns = (page: Page) =>
  page.locator("#proposal-form").evaluate((form) => {
    const { display, gridTemplateColumns } = getComputedStyle(form);
    return display === "grid" ? gridTemplateColumns.split(" ").length : 0;
  });
```

`grid-cols-3` computes to three px tracks even with one child; multicol computes to `none`. Three lines, no `.card` coupling, no helper that depends on markup the test isn't about.

### 2. The footer assertion doesn't test what its name promises

`getBoundingClientRect()` on the form returns its border box and excludes overflowing children by construction. If the overflow you describe happens, the form's own box is the one measurement that under-reports it.

Empirically, across four multicol variants (tall card first, tall card last, many-small-plus-one-tall, and the real form's `display:contents` shape) Chromium expanded the container to its tallest column every time: `form.bottom === footer.top` exactly, no card ever crossed the footer. The assertion passed in all of them. The PR body concedes as much — "against the old markup it fails at the first assertion."

Either measure what actually overlapped (lowest `.card` bottom vs footer top) or drop the assertion and the "never covers the footer" clause from the test name. Right now it's decorative, and a decorative assertion in a regression test is worse than none — it reads as coverage.

### 3. The two `<div class="contents">` wrappers are identity wrappers now, and they mislead

Neither carries an id, an hx-target, or a class beyond `contents`. They render nothing.

Under multicol they at least matched the layout: content flowed column-first, so "group 1" genuinely occupied the first column. Grid auto-placement is row-major — row 1 is now `[Session Details | Participants and duration | Facilitators]`, which splits the pair the first wrapper groups. The wrappers encode a grouping the layout no longer honours, and the PR body's justification ("they're only broken inside multicol") argues they don't *hurt*, not that they earn their keep.

Pick one:

- the grouping doesn't matter → **delete both wrapper elements**. Cards become direct grid children, rendering is byte-identical, one concept disappears from the template.
- the grouping matters → drop `contents` and make each wrapper a real grid item (`class="space-y-6"`), so the group stacks in its own column.

Keeping `contents` is the only option that means nothing.

### 4. `#proposal-session-fields` is now a silent exception to the layout

It's the one non-`contents` block in the form, so its up-to-three cards (participants/duration, session fields, orphan values) stack in a single cell while every other card gets its own. That's forced by `hx-swap="outerHTML"` needing a real element — fine — but the diff removed `break-inside-avoid`, which was the marker that said "this is one unit," and nothing replaced it.

Either give it `class="contents"` (an outerHTML swap re-sends the partial's root, so htmx keeps working) so it flows like everything else, or leave a one-line comment saying it's deliberately one cell. This is also the item that interacts with the whitespace tradeoff you disclosed: with balancing gone, this stack being the tallest thing in row 1 is what sets row 1's height.

### 5. Minor

The seven `<div class="card"><div class="card-body"><h3 class="text-lg font-semibold text-foreground mb-2">` openings are exactly the lines this diff touched. There's no card partial in `components/` though `.card`/`.card-body` are design-system classes — worth extracting while you're already editing all seven.

The `admin/login` + `e2e-manager` beforeEach is copy #10 across the suite. It's the house convention so it isn't this PR's problem, but a `loginAsManager(page)` helper is overdue.

---

### Checked and fine

- Tailwind's `grid-cols-*` emits `repeat(N, minmax(0, 1fr))` (confirmed in the built CSS), so the multicol→grid switch carries no min-content blowout risk. No `min-w-0` needed.
- No nested `.card` under `facilitator-picker`, `time-slot-picker`, `proposal-duration` or `_personal_data_fields`, so the descendant selector isn't over-counting today.
- `columns-*` appears nowhere else in the templates; `break-inside-avoid` survives only in `chronology/print.html`, which is print media and legitimate.
- `items-start` + `gap-6` correctly subsumes the per-card `mb-6`.

**Verdict:** template change is good. Blocking on #1 — a regression test that passes on the broken markup at a different card count isn't a regression test — and on #3, which is a free deletion.

Issue #690, opened 2026-07-25
Extract a partial for the POST icon-button row actions


fancysnake:
Found in review of zagrajmy/ludamus#654.

Panel table row actions are POST forms wrapping a single icon button, and the markup is copy-pasted. `panel/facilitators.html` alone has five near-identical blocks (flag, unflag, mark-as-guest, assign-organizer, step-down), each ~13 lines of:

* `<form method="post" action="{% url ... %}">` with optional `data-confirm` / `data-confirm-action`
* `{% csrf_token %}`
* `<input type="hidden" name="next" value="{{ request.get_full_path }}">`
* an icon `<button type="submit">` carrying `title` and `aria-label`

Every new row action copies the block again, and the accessibility attributes are exactly the part a copy-paste tends to drop.

## Proposal

One component in `src/ludamus/templates/components/` taking url, icon, label, confirm copy, and next. Convert the five call sites in `facilitators.html` first.

## Scope beyond facilitators.html

Looser cousins of the same shape live in `panel/proposals.html`, `panel/proposal-detail.html`, `panel/content-log.html`, `panel/timetable-log.html`, `crowd/user/parties.html`, `crowd/user/party_detail.html`, `crowd/user/safety.html`, `chronology/offer_claim.html`. They differ in whether they carry `next` and whether they confirm, so converting them is a judgement call per call site rather than a mechanical sweep — worth doing only if the partial covers them without sprouting flags.

Related: zagrajmy/ludamus#586 (Tessera button lint rule) covers `.btn` drift; this is the surrounding form, not the button styling.

---

## Update after zagrajmy/ludamus#769 (facilitator soft delete)

The five blocks in `panel/facilitators.html` are still five, but they are no longer the same five:

* `flag` / `unflag` became **delete** / **restore**, and delete carries `data-confirm` + `data-confirm-action` — so the partial has to take confirmation copy, not just treat it as optional decoration.
* **Two shapes, not three.** An earlier revision of #769 rendered a third shape for a facilitator that runs sessions — a bare `<button type="button" disabled>` outside any form, with the refusal in `title`. That is gone: `disabled` is not keyboard- or touch-reachable and the rule belonged to the service, so the row now always submits and the service answers at click time. The partial needs only "form + confirm" and "form, no confirm".

[...]

PR #769, opened 2026-08-02
Feat/facilitator soft delete

fancysnake:
## Summary by CodeRabbit

* **New Features**
  * Facilitators can be deleted and restored instead of flagged for deletion.
  * Deleted facilitators appear in a dedicated view with clear status indicators.
  * Bulk delete and restore actions report skipped facilitators and explain restrictions.
  * Restoring a matching imported facilitator reactivates the existing record.
* **Bug Fixes**
  * Facilitators linked to sessions cannot be deleted.
  * Deleted facilitators are excluded from editing, proposals, and session listings.


src/ludamus/mills/panel_facilitators.py
```diff
@@ -627,11 +633,23 @@ def set_columns(self, *, event_id: int, columns: list[str]) -> None:
             raise EmptyColumnSelectionError
         self._repos.panel_settings.update_facilitator_columns(event_id, keys)
 
-    def set_flag(self, *, event_id: int, facilitator_slug: str, flagged: bool) -> None:
+    def delete(self, *, event_id: int, facilitator_slug: str) -> None:
```
fancysnake: **This is the only mutating method in `FacilitatorPanelService` without `with self._transaction.atomic()`.** `create_facilitator`, `merge`, `set_accreditation`, `assign_organizer` all wrap; `SessionDeletionService.restore` wraps a *single* repo call. This one does read → `has_sessions` → `soft_delete`, i.e. the one place in the class that actually needs the transaction, and it's bare.

Concretely: organizer A opens the list, organizer B attaches the facilitator to a new session, A clicks delete. `has_sessions` was checked against a snapshot that no longer holds, and you get exactly the state the guard exists to forbid — a session whose facilitator is deleted and therefore filtered out of `read_facilitators_for_sessions`. A byline vanishes from the program with nothing on the session saying why.

Wrap read+check+write in `atomic()`, consistent with the rest of the class. `restore` should wrap too, for the same reason and for symmetry — a reviewer shouldn't have to work out why two of eight writes are special.

src/ludamus/links/db/django/repositories/submissions.py
```diff
@@ -1017,13 +1038,45 @@ def release(pk: int, *, organizer_id: int | None) -> bool:
             qs = qs.filter(organizer_id=organizer_id)
         return bool(qs.update(organizer=None))
 
+    @staticmethod
+    def has_sessions(pk: int) -> bool:
```
fancysnake: **`sessions__deleted_at__isnull=True` opens a hole that lets the exact failure the guard prevents happen anyway.**

Walk it: facilitator's only session gets soft-deleted from the proposals panel → `has_sessions` now returns `False` → the facilitator is deletable, and the detail page's session list (alive-only) shows nothing, so an organizer has no reason to hesitate → they delete → someone hits **Restore** in the proposals "Recently deleted" section. `ChronologyService.restore` (`mills/chronology.py`) brings the session back and touches nothing else. The session is now live with a soft-deleted facilitator attached, `read_facilitators_for_sessions` filters them out, and you have a session in the program with a missing byline and no trace of why — the precise outcome `HAS_SESSIONS` was written to prevent.

The comment above this method ("Deleted sessions no longer name their facilitator anywhere, so they must not hold a deletion up either") is only true if session deletion is terminal. It isn't; it's reversible, by design, in the same panel.

Two clean ways out — pick one, don't add a third state:
- Drop the `deleted_at` filter here. A facilitator on any session, alive or in the bin, isn't deletable. Boring, one-line, no cross-feature coupling. The cost is an organizer occasionally having to empty the session bin first, which is honest.
- Make session restore restore its deleted facilitators inside the same transaction, so restore is total rather than partial. More correct-feeling, but it wires two features together and you'd want a test asserting the pair moves as one.

Either way this needs a test at the seam: soft-delete session → delete facilitator → restore session → assert the byline is there.

src/ludamus/gates/web/django/chronology/panel/views/facilitators.py
```diff
@@ -777,16 +782,20 @@ def post(self, _request: PanelRequest, slug: str) -> HttpResponse:
...
             return redirect(f"{base}?{query}")
 
-        applied = missing = 0
+        applied = missing = refused = 0
         for facilitator_slug in slugs:
             try:
                 self._apply(action, current_event.pk, facilitator_slug)
             except NotFoundError:
                 missing += 1
+            except FacilitatorActionError:
```
fancysnake: **This blanket `except` throws away `exc.refusal`, then `_report` hardcodes one refusal's message as if it were the only one.**

The single-action path 100 lines up already solves this properly: `messages.error(self.request, _ORGANIZER_REFUSALS[exc.refusal])`. The bulk path instead counts refusals into one `int` and writes a fresh translatable string (`"%(count)d facilitator runs sessions and was kept."`) that asserts a reason it never checked. It happens to be true today because `delete` is the only bulk-reachable method raising `FacilitatorActionError`. That's an accident of the current call graph, not a property of the code — the day `set_accreditation` or a future bulk action raises `ALREADY_YOURS` or `NOT_ORGANIZER`, this reports "runs sessions" and lies to the organizer with a straight face.

Code judo: count by refusal, not by "something refused".

```python
refused: Counter[OrganizerActionRefusal] = Counter()
...
except FacilitatorActionError as exc:
    refused[exc.refusal] += 1
```

and in `_report`, one loop emitting `_ORGANIZER_REFUSALS[refusal]` per bucket. That deletes the bespoke plural string entirely (four Polish plural forms in `django.po` go with it), removes the duplicated copy, and makes the bulk path structurally incapable of misreporting a reason. The message loses the count prefix — if you want it back, `f"{n} × {msg}"` or a single generic ngettext wrapper around the per-reason text beats maintaining a parallel copy table.

src/ludamus/templates/panel/facilitators.html
```diff
@@ -126,12 +126,12 @@
                             </span>
                             <button type="submit"
                                     name="action"
-                                    value="flag"
-                                    class="btn btn-danger text-sm bulk-action-btn">{% translate "Flag for deletion" %}</button>
+                                    value="delete"
```
fancysnake: **The bulk bar is identical in both list modes, but the two lists share no valid actions except one each way.**

You made the deliberate — and correct — call that deleted and live never mix (`deleted_at__isnull=not filters.get("deleted")`, with a note saying a restore is always a deliberate visit to the bin). The action bar didn't get the memo:

- In the bin, **Delete** and **Mark as guest** resolve through alive-only reads → every row `NotFound` → *"N facilitators could not be found."* Which is false; they're right there on screen.
- In the bin, **Merge** hands the slugs to the merge basket, which loads them via `list_by_slugs` → alive-only → the selection silently shrinks, sometimes to nothing, with no explanation.
- In the live list, **Restore** hits `NotFoundError` for every row → the same misleading "could not be found".

So four buttons, and in each mode at most two do anything, with the failures reported as data errors rather than as "that action doesn't apply here". Gate the bar on `filter_deleted`: **Restore** alone in the bin, **Delete / Mark as guest / Merge** in the live list. The state model already carries the answer — the template just isn't reading it.

src/ludamus/templates/panel/facilitators.html
```diff
@@ -229,36 +229,48 @@
...
                                                 <button type="submit"
-                                                        title="{% translate "Remove deletion flag" %}"
-                                                        aria-label="{% blocktranslate with name=facilitator.display_name %}Remove deletion flag from {{ name }}{% endblocktranslate %}"
+                                                        title="{% translate "Restore" %}"
+                                                        aria-label="{% blocktranslate with name=facilitator.display_name %}Restore {{ name }}{% endblocktranslate %}"
                                                         class="text-foreground-muted hover:text-foreground inline-flex items-center">
                                                     {% icon "arrow-uturn-left" class="w-4 h-4" %}
                                                 </button>
                                             </form>
+                                        {% elif facilitator.session_count %}
```
fancysnake: **A `disabled` button carrying the refusal in a `title` isn't reachable by the people who most need it.** `disabled` removes it from the tab order, so keyboard and screen-reader users can't get to it; `title` tooltips don't fire on touch at all. The project's definition of done says keyboard-reachable semantic HTML, and this row action is the one place the rule is explained.

It also puts the business rule in a third location. "Has live sessions ⇒ not deletable" now lives in `FacilitatorRepository.has_sessions`, in `FacilitatorPanelService.delete`, and here as `{% elif facilitator.session_count %}` — with the refusal string duplicated between `_ORGANIZER_REFUSALS` and this template (`django.po` now lists the same msgid under both files). Three copies of one rule, and the template's copy is the one that silently rots when the rule changes — which, per my comment on `has_sessions`, it probably should.

Cheapest fix that's also the most honest: render the delete form unconditionally and let the service refuse. The refusal message already exists, already renders, and is already correct at the moment of the click rather than at the moment of the page render — which matters, since a session can be attached between the two. If you want to keep the affordance, `aria-disabled="true"` on a real focusable button plus a visible hint beats `disabled` + `title`.

(The same rule also has to be reflected in the *bulk* Delete button, which it isn't — another sign it doesn't belong in the template.)

src/ludamus/links/db/django/repositories/submissions.py
```diff
@@ -887,8 +907,16 @@ def list_by_event(
         event_id: int, filters: FacilitatorListFilters | None = None
     ) -> list[FacilitatorListItemDTO]:
         filters = filters or {}
-        qs = Facilitator.objects.filter(event_id=event_id).annotate(
-            session_count=Count("sessions", distinct=True),
+        # Deleted and live never mix, so a restore is always a deliberate visit
+        # to the bin.
+        qs = Facilitator.all_objects.filter(
+            event_id=event_id, deleted_at__isnull=not filters.get("deleted")
```
fancysnake: **`list_by_event` got the alive-sessions fix; `list_by_slugs`, thirty lines down in the same class, didn't.**

Here: `Count("sessions", filter=Q(sessions__deleted_at__isnull=True), distinct=True)`. There: `Count("sessions", distinct=True)`, still counting soft-deleted sessions, still on the alive-only `Facilitator.objects`. Same annotation feeding the same `FacilitatorListItemDTO.session_count`, two different answers — so the merge confirm page can show "3 sessions" for a facilitator the list page shows as "1", and the organizer decides which one to keep on the wrong number.

That divergence is what a shared annotation helper is for. Pull the `Count(...)` expression out as a module-level constant (or a small `_with_session_count(qs)`) and use it in both — one place to change when the rule moves again, and the drift becomes impossible rather than merely unlikely.

src/ludamus/mills/panel_facilitators.py
```diff
@@ -627,11 +633,23 @@ def set_columns(self, *, event_id: int, columns: list[str]) -> None:
...
+        # A facilitator running sessions cannot be deleted: their byline would
+        # vanish from the program with nothing on the session saying why. The
+        # sessions go first, or the facilitator stays.
         facilitator = self._repos.facilitators.read_by_event_and_slug(
             event_id, facilitator_slug
         )
-        self._repos.facilitators.set_flag(facilitator.pk, flagged=flagged)
+        if self._repos.facilitators.has_sessions(facilitator.pk):
```
fancysnake: **Delete and restore write no `facilitator_change_logs` entry, so the History tab now has a hole exactly where the largest state change is.**

`set_accreditation` logs a `ContentFieldChange` for flipping one enum value. Deleting a facilitator — removing them from the program entirely — logs nothing, and neither does bringing them back. An organizer opening the History tab on a restored facilitator sees an unbroken record with the deletion missing from it, which is worse than no history: it reads as "nothing happened here".

I'll grant the old `set_flag` didn't log either, so this is inherited rather than introduced. But `flagged` was a sticky note and this is a deletion; the project's definition of done asks new paths to log meaningful events, and "who deleted this facilitator and when" is the single most meaningful event this feature produces. `deleted_at` records *when* and nothing else — not *who*, and a restore erases even that.

Both writes need the transaction anyway (see my comment on `delete`) — add the log entry inside it and you get who/when/what atomically for free.

src/ludamus/pacts/panel.py
```diff
@@ -285,7 +285,7 @@ def search_candidates(
     ) -> list[FacilitatorListItemDTO]: ...
     def list_fields(self, event_id: int) -> list[OrganizerFieldDTO]: ...
     def detail_context(
-        self, *, event_id: int, facilitator_slug: str
+        self, *, event_id: int, facilitator_slug: str, include_deleted: bool = False
```
fancysnake: **`include_deleted: bool = False` is a mode flag plumbed through the protocol, the service, and one call site — for a single page.**

Lower conviction than my other comments; the flag genuinely does work and the "write paths get NotFound on a dead slug" invariant it protects is a good one, stated clearly in the service comment. But note what it costs: a boolean parameter in the `Protocol`, a two-branch conditional in `detail_context`, and a second repo read method (`read_including_deleted`) whose only difference from `read_by_event_and_slug` is the manager. Three layers of optionality, one caller passing `True`.

At minimum, drop the `= False` default from the *protocol* signature. A default in a `Protocol` method means implementors are contracted to two call shapes, and every reader has to work out which callers rely on the implicit one. Making it explicit at the three call sites costs nine characters and removes the question.

Worth checking whether the branch disappears entirely: if `detail_context` always read through `read_including_deleted` and the edit GET/POST rejected `facilitator.deleted_at` explicitly, you'd trade one boolean + one conditional + one repo method for two guards in the views that also read more honestly ("you can't edit a deleted facilitator" is a rule worth stating out loud, rather than one that emerges from which manager the read happened to use). I don't think that's obviously better — it's two checks instead of one flag — but it's close enough that it deserves a deliberate answer rather than the default.

mise.toml
```diff
@@ -263,13 +263,24 @@ env.PYTHONPATH = "src"
 run = [{ task = "format" }, { task = "lint" }]
 
 [tasks.devcheck]
-run = [{ task = "lint:djlint" }, { task = "check:py" }]
+run = [
```
fancysnake, line 266: **Unrelated to soft delete, and it leaves `devcheck` and `fullcheck` as near-duplicates.**

`devcheck` now runs format + djlint + lint:py + full test suite with coverage + `coverage xml` + diff-cover + tingle. `fullcheck` runs format + lint + diff-cover + tingle, and no longer delegates to `devcheck` — so the two tasks now list overlapping step sequences that have to be kept in sync by hand, which is how they drifted apart in the first place. The fast inner-loop check is also no longer fast; a full `-xvv` suite plus coverage plus diff-cover isn't something you run between edits, so in practice this deletes the quick check rather than improving it.

If `devcheck` is meant to be the thorough one now, make `fullcheck` call it and add only what it adds (`poetry install`, the e2e env), instead of re-listing the shared steps. And this belongs in its own PR — a reviewer looking at facilitator soft delete has no reason to be auditing the check pipeline.

fancysnake (commented):

The core move — replacing the `flagged_for_deletion` triage boolean with the codebase's existing `SoftDeleteModel` — is the right one. It deletes a concept instead of adding one, the repo `soft_delete`/`restore` pair matches the canonical `SessionRepository`/`DiscountRepository` shape exactly, the uniques-count-dead-rows decision is deliberate and documented, and the import restore path is a genuinely good catch. Test coverage is real, not decorative. No file crosses 1k lines because of this PR.

What I'm pushing back on, in priority order:

1. **`FacilitatorPanelService.delete` is the only write in the class with no transaction** — and it's the one doing check-then-act.
2. **The "has sessions" guard has a hole that defeats its own purpose**: soft-deleted sessions don't count, so restoring a session can resurrect an orphaned byline — exactly the outcome the refusal exists to prevent.
3. **The bulk view re-invents refusal copy** that `_ORGANIZER_REFUSALS` already models, and asserts one specific reason from a blanket `except`.
4. **The bulk action bar doesn't follow the list it's attached to** — in the bin, three of four buttons are dead ends.
5. **`list_by_slugs` didn't get the `session_count` fix `list_by_event` got** — same class, same annotation, now two different answers.
6. **Delete/restore write no change-log entry** while every neighbouring write does; the History tab now has a hole where the biggest state change should be.

Details inline.

src/ludamus/links/db/django/repositories/submissions.py
fancysnake: Fixed in 6617ca4 — took the first option. `has_sessions` drops the `deleted_at` filter: a facilitator named on any session, alive or in the bin, is not deletable. No cross-feature coupling, and the cost is the honest one — the session bin gets emptied first.

Tests at the seam: `test_restoring_a_session_keeps_its_byline` (soft-delete session → delete facilitator → restore session → assert the byline is there), plus `test_delete_is_refused_while_a_deleted_session_names_them`. `test_delete_goes_through_once_the_session_is_deleted` was asserting the bug, so it now covers the session letting go of the facilitator instead. The refusal copy says "deleted ones included" so the organizer isn't left guessing why a facilitator with a 0 in the Sessions column won't delete.

src/ludamus/mills/panel_facilitators.py
fancysnake: Both wrapped in 6617ca4, along with the change-log write from your other comment.

One correction so it isn't recorded as more than it is: `atomic()` alone does not close the TOCTOU you describe. Under Django's default READ COMMITTED, organizer B's through-table insert still commits between A's `has_sessions` check and A's write — closing that needs a locking read or a conditional update. What the transaction does buy is that the check, the soft delete and the log entry land together, which is the reason it can't stay bare now.

src/ludamus/gates/web/django/chronology/panel/views/facilitators.py
fancysnake: Fixed in 6617ca4. `refused` is a `Counter[OrganizerActionRefusal]`, and `_report` emits one message per bucket.

Took the ngettext variant you offered rather than `n × msg`: one reason-agnostic wrapper ("%(count)d facilitator was skipped: %(reason)s") around `_ORGANIZER_REFUSALS[refusal]`. Keeps the count and reads properly in Polish, and the bespoke "runs sessions and was kept" string with its four plural forms is gone from the catalog. The bulk path can no longer name a reason it didn't check.

src/ludamus/templates/panel/facilitators.html
fancysnake: Fixed in 6617ca4 — the bar is gated on `filter_deleted`: Restore alone in the bin, Delete / Mark as guest / Merge in the live list.

Extended it to the row actions while I was there: assign-organizer, step down and mark-as-guest were rendering on deleted rows too, and each would have come back "Facilitator not found" about a row on screen. A deleted row now offers restore only. Asserted in the e2e spec, since the claim is about markup.

src/ludamus/templates/panel/facilitators.html
fancysnake: Fixed in 6617ca4 — took the cheapest and most honest option. The delete form renders unconditionally and the service refuses; the disabled button, its `title`, and the duplicated refusal string in the template are all gone, so the rule lives in `has_sessions` and `FacilitatorPanelService.delete` and nowhere else. As you note, it is also correct at click time rather than render time.

The e2e test that asserted the button was disabled now clicks it and expects the refusal message.

src/ludamus/links/db/django/repositories/submissions.py
fancysnake: Fixed in 6617ca4. The expression is a module-level `_live_session_count()` used by both `list_by_event` and `list_by_slugs`, with a repository test asserting the list and the merge basket report the same number for a facilitator holding one live and one deleted session.

src/ludamus/mills/panel_facilitators.py
fancysnake: Fixed in 6617ca4. Both writes log a `deleted` `ContentFieldChange` ("" → "yes" on delete, the reverse on restore) inside the same transaction, so who/when/what land atomically. `content_field_label` gained a "Deleted" label so the History tab renders it like any other change; the value is stored raw, the way accreditation types already are. Unit tests cover both directions and assert nothing is logged when the deletion is refused.

src/ludamus/pacts/panel.py
fancysnake: Took the minimum in 6617ca4: the `= False` default is gone from the `Protocol` and from the service, so all three call sites state their intent.

The larger question — always read through `read_including_deleted` and have the edit GET/POST reject `deleted_at` explicitly — is genuinely close, so it gets a deliberate answer rather than a drive-by: zagrajmy/ludamus#802, with both sides written down and "keep the flag, close this" listed as a valid outcome.

mise.toml
fancysnake: Reverted in 6617ca4 — `mise.toml` is back to main. Agreed on both counts: it isn't this PR's business, and re-listing the shared steps is what let the two drift. If `devcheck` becomes the thorough one, `fullcheck` should call it and add only what it adds.

src/ludamus/templates/panel/facilitators.html
```diff
@@ -178,28 +178,30 @@
...
                                             {% csrf_token %}
                                             <input type="hidden" name="next" value="{{ request.get_full_path }}">
                                             <button type="submit"
-                                                    title="{% translate "Flag for deletion" %}"
-                                                    aria-label="{% blocktranslate with name=facilitator.display_name %}Flag {{ name }} for deletion{% endblocktranslate %}"
+                                                    title="{% translate "Delete" %}"
+                                                    aria-label="{% blocktranslate with name=facilitator.display_name %}Delete {{ name }}{% endblocktranslate %}"
                                                     class="text-foreground-muted hover:text-danger inline-flex items-center">
-                                                {% icon "flag" class="w-4 h-4" %}
+                                                {% icon "trash" class="w-4 h-4" %}
```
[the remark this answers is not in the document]
fancysnake: Not in this PR, and the reason is worth writing down: Tessera has no icon-only submit control — `tessera_button` with `icon=` always renders text alongside it — so these two would be the only converted buttons in a cell that holds four more hand-rolled ones. Converting one action at a time just moves the inconsistency around.

Tracked on zagrajmy/ludamus#690, which already covers extracting the surrounding POST-form partial; I updated it with the Tessera control as the other half of the same job. The disabled-button shape that issue described is gone as of 6617ca4 — the row delete always submits now.

(The HTMLHint findings above are that linter parsing Django template tags; `djlint` and `frontend-analysis` are both clean.)

tests/integration/web/panel/test_facilitators_page.py
```diff
@@ -443,28 +445,28 @@ def test_sort_by_name_descending(
             },
         )
 
-    def test_flagged_filter(self, authenticated_client, active_user, sphere, event):
+    def test_deleted_filter(self, authenticated_client, active_user, sphere, event):
```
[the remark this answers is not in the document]
fancysnake: Won't do — `pyproject.toml` waives `PLR0913` and `PLR0917` for `tests/**.py` on purpose, and the suite is ~1070 positional-fixture signatures to ~54 keyword-only. Making these nine the exception would leave the file inconsistent with every test around it.

fancysnake:
Review round addressed in 6617ca4 — all twelve threads replied to and resolved.

**Fixed**

- Deleted sessions now hold a facilitator deletion up too. They are restorable, so the old filter let `delete session → delete facilitator → restore session` produce a live session with a missing byline. Covered by a seam test.
- `delete` / `restore` run in one transaction and write a `deleted` change-log entry, so the History tab records who and when.
- Bulk refusals are counted per reason and reported with the refusal the service actually gave.
- The list offers only the actions that work in the mode it shows — restore in the bin, delete / mark as guest / merge in the live list — for the bulk bar and the row actions alike.
- The row delete always submits and lets the service refuse; the unreachable `disabled` + `title` button and the duplicated rule in the template are gone.
- One shared session-count annotation for the list and the merge basket.
- `include_deleted` loses its protocol default; the deeper question is zagrajmy/ludamus#802.
- The `devcheck` / `fullcheck` rewrite is reverted — separate PR.

**Deferred, with reasons on the threads:** the Tessera icon-button control (zagrajmy/ludamus#690 — no icon-only submit exists yet, so it is an all-row-actions job) and keyword-only test fixtures (waived for `tests/**` in `pyproject.toml`).

**tingle** is red at +121 (`legacy-loc` +7, `old-subdomain-loc` +114) and will stay red. Every fix above lands in `gates/web/django/chronology/panel/views/facilitators.py` and `links/db/django/repositories/submissions.py`, both counted as legacy — there is no version of this change that shrinks those files. tingle located no removable additions.

src/ludamus/links/db/django/migrations/0143_facilitator_flag_to_soft_delete.py
```diff
@@ -0,0 +1,36 @@
...
+    # can delete by hand now. A flagged facilitator still on a session keeps
+    # its byline and only loses the marker, which nothing else read.
+    # Reversible: a dead row flags again.
+    facilitator = apps.get_model("db_main", "Facilitator")
+    running_a_session = facilitator.objects.filter(
+        sessions__isnull=False, sessions__deleted_at__isnull=True
+    ).values("pk")
+    facilitator.objects.filter(
+        flagged_for_deletion=True, deleted_at__isnull=True
+    ).exclude(pk__in=running_a_session).update(deleted_at=timezone.now())
```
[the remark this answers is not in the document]
fancysnake: Fixed in 09779ba. Right — the migration disagreed with `FacilitatorRepository.has_sessions`, which counts deleted sessions on purpose (`test_a_deleted_session_counts_too` pins it), so it converted exactly the rows an organizer may not delete by hand. `sessions__deleted_at__isnull=True` is gone; any session now holds the conversion up.

src/ludamus/links/db/django/migrations/0147_facilitator_flag_to_soft_delete.py
```diff
@@ -0,0 +1,36 @@
...
+    facilitator.objects.filter(
+        flagged_for_deletion=True, deleted_at__isnull=True
+    ).exclude(pk__in=running_a_session).update(deleted_at=timezone.now())
+
+
+def deleted_at_to_flag(apps, schema_editor):
+    facilitator = apps.get_model("db_main", "Facilitator")
+    facilitator.objects.filter(deleted_at__isnull=False).update(
+        flagged_for_deletion=True
+    )
```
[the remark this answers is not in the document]
fancysnake: Not doing this. `flagged_for_deletion` was a triage marker nothing read but the old table template — no service, no query, no export. Preserving each row's pre-migration value across a column drop needs a side table to park it in, which is a lot of machinery for a badge colour that only reappears if someone rolls back to the old UI. Best-effort reverse is the intent; the inline comment says so.

src/ludamus/mills/panel_facilitators.py
```diff
@@ -627,11 +637,61 @@ def set_columns(self, *, event_id: int, columns: list[str]) -> None:
...
+        # A facilitator named on sessions cannot be deleted: their byline would
+        # vanish from the program with nothing on the session saying why. The
+        # sessions go first, or the facilitator stays.
+        with self._transaction.atomic():
+            facilitator = self._repos.facilitators.read_by_event_and_slug(
+                event_id, facilitator_slug
+            )
+            if self._repos.facilitators.has_sessions(facilitator.pk):
+                raise FacilitatorActionError(OrganizerActionRefusal.HAS_SESSIONS)
+            self._repos.facilitators.soft_delete(facilitator.pk)
```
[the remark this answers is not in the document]
fancysnake: Fixed in 09779ba. `atomic()` is not a lock, agreed — under READ COMMITTED an assignment could commit into the gap. Both sides now take the same facilitator row lock: `FacilitatorRepository.lock` runs before `has_sessions` in the delete, and `_lock_facilitators` runs before the m2m write in `SessionRepository.create` / `set_facilitators` / `replace_facilitators_in_sessions` (ordered by pk so two multi-facilitator writes cannot deadlock). That leaves only the two orderings that end well: the check sees the link, or the link waits and then finds a deleted row. Unit test asserts the lock precedes the check.

tests/integration/web/panel/test_facilitators_page.py
```diff
@@ -1009,37 +1105,190 @@ def test_redirects_non_manager_user(self, authenticated_client, event, name):
...
+        session = _session(event)
+        session.facilitators.add(facilitator)
+        session.soft_delete()
+
+        authenticated_client.post(
+            self._url("panel:facilitator-delete", event, facilitator)
+        )
+        session.restore()
+
+        assert list(session.facilitators.all()) == [facilitator]
```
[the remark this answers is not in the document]
fancysnake: False positive — the test does catch it. Verified by mutation: flipped `has_sessions` to ignore soft-deleted sessions, ran the test, it fails. The premise that the m2m row keeps the facilitator visible is wrong: `Facilitator._default_manager` is `AliveManager`, so `session.facilitators.all()` hides soft-deleted rows and the final assertion goes red the moment the delete succeeds.

src/ludamus/links/db/django/repositories/sessions.py
```diff
@@ -74,6 +75,21 @@
...
+    # same lock before writing the link leaves only the two orderings that end
+    # well — the check sees this link, or this link waits and the delete wins.
+    # `order_by("pk")` keeps two concurrent multi-facilitator writes from
+    # deadlocking on each other.
+    list(
+        Facilitator.all_objects.select_for_update()
+        .filter(pk__in=list(facilitator_ids))
+        .order_by("pk")
+        .values_list("pk", flat=True)
+    )
```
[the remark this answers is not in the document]
fancysnake: Fixed in 7458fcb — you are right, and the code comment above the helper was overclaiming.

`_lock_facilitators` locked through `all_objects` and discarded the result, so only one of the two orderings the lock exists to create actually ended well. When the delete won the lock, the waiting assignment woke up and wrote the link onto the row that had just been soft-deleted — exactly the live-session-with-a-missing-byline state `HAS_SESSIONS` guards. The comment claimed "this link waits and the delete wins" as if that were a good outcome; nothing was checking.

Now the lock goes through `Facilitator.objects` and every requested pk must come back, `NotFoundError` otherwise. Postgres re-applies the filter once the lock is granted, so a row deleted in the gap drops out of the result and the missing pk is the signal. The comment says that instead of asserting an invariant it did not implement.

For the record on reachability: no non-race path could hit this. All six callers already resolve alive-only ids — the two panel views intersect the submitted ids with `list_by_event`, the importer's `_resolve_facilitator` restores a matched dead row before linking, `legacy._find_or_create_facilitator` reads alive-only, and merge's `list_by_slugs` is alive-only. So this is defence at the seam, not a hole in a form flow.

Seam test: `TestSessionFacilitatorLinkRejectsDeleted` soft-deletes a facilitator, calls `set_facilitators`, and asserts `NotFoundError` with no link written. The concurrency half rests on `select_for_update` re-evaluating the filter, which the suite already relies on elsewhere; I did not add a threaded Postgres test for it.

src/ludamus/links/db/django/repositories/sessions.py
```diff
@@ -74,6 +75,29 @@
 _FIELD_SORT_PREFIX = "field_"
 
 
+def _lock_facilitators(facilitator_ids: Iterable[int]) -> None:
```
fancysnake: ## Thermo-nuclear code quality review

**Two facilitator row-lock implementations now exist, in two modules, with contradictory contracts.**

`_lock_facilitators` here locks through the alive manager and raises `NotFoundError` naming the missing pks. `FacilitatorRepository.lock` (`repositories/submissions.py`) locks the same rows through `all_objects`, calls `.first()`, throws the result away, and never raises. Both were written to make one invariant hold — "a facilitator cannot be soft-deleted and assigned at the same time" — and a reader now has to hold two spellings of it, in two files, to see that it holds. `SessionRepository.lock` is a third `lock` in the same package whose contract matches neither.

That `.first()` result being discarded is also the exact shape that was already a bug in this helper once (7458fcb). Leaving one of the pair in that shape invites the same fix a second time.

Code judo: one canonical entry on the facilitator repo — `lock(pks)`, alive-only, raising `NotFoundError`, matching `SessionRepository.lock` — and this module calls it rather than hand-rolling a `Facilitator` queryset inside the *session* repository. It is behaviour-compatible for the delete path too: a row deleted concurrently would come back `NotFound` from `lock` instead of travelling on to `soft_delete` to raise the same error a step later. The session repository stops owning facilitator liveness, the invariant is stated once, and the two halves can no longer drift.

src/ludamus/links/db/django/repositories/submissions.py
```diff
@@ -1023,13 +1046,55 @@ def release(pk: int, *, organizer_id: int | None) -> bool:
...
+
     @staticmethod
     def delete(pk: int) -> None:
-        Facilitator.objects.filter(pk=pk).delete()
+        # Destroys the row. The merge is the only caller: a source that merged
+        # away was never a separate person, so it leaves no restorable trace.
+        Facilitator.all_objects.filter(pk=pk).delete()
+
+    @staticmethod
+    def soft_delete(pk: int) -> None:
```
fancysnake: ## Thermo-nuclear code quality review

**Third verbatim copy of the soft-delete repository boilerplate — comment included.**

`SessionRepository.soft_delete`, `DiscountRepository.soft_delete` and now `FacilitatorRepository.soft_delete` are the same five lines under the same copy-pasted comment ("Reach through `all_objects` so an already-dead row raises NotFound instead of silently re-stamping `deleted_at`"). `restore` is on its way to the same fate: sessions have one, facilitators now have one, and they differ only in the extra scoping predicate.

Two copies is a coincidence; three is a missing helper. `SoftDeleteModel` already owns the state transition — the repo-side invariant around it (already-dead / missing → `NotFoundError`, never a silent no-op) should be stated once too:

```python
def soft_delete_row(manager, **lookups) -> None: ...
def restore_row(manager, **lookups) -> None: ...
```

in a shared repositories module, with each repo reduced to a one-liner naming its own scoping (`category__event_id=event_pk` for sessions, nothing for facilitators). One place to read the rule, one place to change it, and the fourth soft-deletable model costs a line instead of a paste.

src/ludamus/mills/submissions/engine.py
```diff
@@ -588,16 +589,41 @@ def _facilitator_ident(
...
     ) -> int:
+        matched = self._match_facilitator(
+            event_id=event_id, ident=ident, display_name=display_name
+        )
+        if matched is not None:
+            if matched.deleted_at is not None:
+                # The source row still names them, so a deleted facilitator
+                # comes back rather than colliding with the slug and ident
+                # their dead row keeps reserved.
+                self._repos.facilitators.restore(matched.pk)
```
fancysnake, line 601: ## Thermo-nuclear code quality review

**The import restores facilitators straight through the repository, so it reopens the audit hole this PR just closed — on the path most likely to hit it.**

`FacilitatorPanelService.delete` / `restore` now write a `deleted` `ContentFieldChange` inside the transaction, on the stated grounds that `deleted_at` records only *when* and a restore erases even that. This call bypasses the service entirely: no log entry, no `user_id`, nothing in the History tab. `test_run_restores_a_matched_facilitator` pins that shape (`facilitators.restore.assert_called_once_with(55)` and no log assertion at all), so it is the tested contract, not an oversight in passing.

The consequence compounds: an organizer deletes a facilitator, the next scheduled import re-run brings them back because the source sheet still names them, and the only trace anywhere is the delete entry — the History tab reads "deleted", the panel shows them alive, and nothing reconciles the two. Every subsequent re-run repeats it.

"A restore leaves a trace" must not depend on which caller reached the repo. Give the restore one owner — a `restore_facilitator(...)` helper (or the service method) used by both the panel and the engine, logging with `user_id=None` for the import — and the invariant becomes structural instead of a property of two call sites happening to agree.

Separately, worth stating out loud rather than in a code comment: silently undoing a deliberate organizer deletion is a *policy*. If it is the right one, the import log is the place it should be visible.

src/ludamus/gates/web/django/chronology/panel/views/facilitators.py
```diff
@@ -775,14 +797,22 @@ def _apply(self, action: str, event_id: int, facilitator_slug: str) -> None:
                 accreditation_type=AccreditationType.GUEST.value,
                 user_id=self.request.context.current_user_id,
             )
+        elif action == "delete":
+            panel.delete(
+                event_id=event_id,
+                facilitator_slug=facilitator_slug,
+                user_id=self.request.context.current_user_id,
+            )
         else:
```
fancysnake: ## Thermo-nuclear code quality review

**The bulk action that runs by falling off the end of an if-chain.**

`_apply` dispatches on a string through `mark-guest` → `delete` → `else`, and `else` silently means **restore**. Nothing in the code says so; it holds only because `merge` returns early upstream and `_BULK_FACILITATOR_ACTIONS` happens to list exactly four entries. Add a fifth bulk action and forget to extend `_apply`, and the selection gets restored rather than erroring — the one failure mode a dispatcher must never have, since it is invisible in review and reports success to the organizer.

The three branches are also the same call with the same three kwargs and a different method name, so the branching buys nothing.

Judo: one table, with the allow-list derived from it instead of maintained beside it.

```python
_BULK_FACILITATOR_ACTIONS = {
    "delete": lambda panel, kw: panel.delete(**kw),
    "restore": lambda panel, kw: panel.restore(**kw),
    "mark-guest": lambda panel, kw: panel.set_accreditation(
        **kw, accreditation_type=AccreditationType.GUEST.value
    ),
}
```

`merge` stays the redirect it already is, checked on its own before the loop. That deletes the duplicated tuple, deletes the chain, and makes "unknown action" the only way out — which is what the membership check upstream already promises.

If you would rather keep the chain, the minimum is that the last branch names its action (`elif action == "restore":`) with an explicit failure after it. A catch-all that mutates data is not a default worth keeping.

src/ludamus/templates/panel/facilitators.html
```diff
@@ -124,22 +124,28 @@
...
-                                            value="mark-guest"
-                                            class="btn btn-secondary text-sm bulk-action-btn">{% translate "Mark as guest" %}</button>
-                                    <button type="submit"
-                                            name="action"
-                                            value="merge"
-                                            class="btn btn-secondary text-sm bulk-action-btn">{% translate "Merge selected" %}</button>
+                                    {# Deleted and live never mix, so neither do their actions: every
+                               action below reads through the alive-only list and would report
+                               the rows on screen as missing in the other mode. #}
+                                    {% if filter_deleted %}
```
fancysnake: ## Thermo-nuclear code quality review

**The bin as a mode flag, re-derived in seven places, when the sibling page already models it without one.**

`deleted` is not a filter. Every other key in `FacilitatorListFilters` narrows the list; this one *swaps which list you are on* — `deleted_at__isnull=not filters.get("deleted")` inverts a predicate rather than adding one — and it swaps which actions are legal. The template then has to reconstruct that mode over and over: `{% if filter_deleted %}` around the bulk bar here, `{% if not facilitator.deleted_at %}` at 199, `{% if facilitator.deleted_at %}` at 243, `{% if not facilitator.deleted_at %}` again at 273, the badge at 183, the row tint at 169 — plus `filters_active` counting it alongside real filters, and a "Deleted only" checkbox sitting in the filter panel as if it were one.

Every one of those predicates answers the same question the request already answered. That is the definition of a missing boundary.

`proposals.html` solved this without a mode: `list_deleted(event_id)` feeds `deleted_proposals`, rendered as its own "Recently deleted" section carrying the single action that applies there, and nothing else on the page knows deletion exists. No flag in the filter DTO, no inverted queryset, no conditionals in the live rows.

The judo move is the same shape here — `list_deleted` plus its own section or route — and it deletes the `deleted` filter key, the predicate inversion, the bulk-bar gate and every `deleted_at` conditional in the action cell at once.

The honest counterweight: the facilitator bin can be long, and search / sort / column config plausibly belong in it, which the proposals section does not need. If that is the reason for one list with a mode, say so in the PR — but then make the mode explicit at the boundary (two views over one template plus a shared row partial that takes the legal actions as an argument), not seven template predicates each re-deriving which list the reader is looking at.

src/ludamus/links/db/django/repositories/submissions.py
```diff
@@ -1023,13 +1046,55 @@ def release(pk: int, *, organizer_id: int | None) -> bool:
...
+        # Held across the `has_sessions` check and the soft delete. Every path
+        # that puts a facilitator on a session takes the same row lock first,
+        # so the two serialize: either the check sees the assignment, or the
+        # assignment waits and then finds a deleted row. Without it `atomic()`
+        # alone lets an assignment commit into the gap and leave a deleted
+        # facilitator on a live session.
+        Facilitator.all_objects.select_for_update().filter(pk=pk).first()
+
+    @staticmethod
+    def has_sessions(pk: int) -> bool:
```
fancysnake: ## Thermo-nuclear code quality review

**One noun, two definitions, in one class: `has_sessions` means "any session, ever" while `session_count` means "live sessions only".**

`_live_session_count()` (line 828) filters to `deleted_at__isnull=True`; this method deliberately does not. Both describe a facilitator's relationship to sessions, both feed the same screen, and neither name carries the distinction — a maintainer reading the list column and the delete guard has no way to know they disagree until an organizer reports that a facilitator showing **0 sessions** refuses to delete.

The refusal copy ("deleted ones included") patches the symptom, and it is the only place in the product where that second definition surfaces. Nowhere in the facilitator UI can the organizer *see* the rows holding the delete up: the list column excludes them and the detail page's session list is alive-only. So the guard is honest but unactionable — the instruction is "remove them from those sessions first" and the sessions are invisible.

Make the model carry both numbers from one place instead of two methods quietly disagreeing. Annotating live and deleted counts from the same expression lets the guard read `deleted_session_count or session_count`, and lets the row or the refusal say the actionable thing — "1 deleted session still names them" with a link to the proposal bin. One definition, stated once, and the organizer gets told what to do rather than what is forbidden.

At minimum the names should stop lying: `has_any_session` next to `live_session_count` costs nothing and removes the trap.

src/ludamus/links/db/django/repositories/submissions.py
```diff
@@ -825,6 +825,13 @@ def _to_dto(field: SessionField) -> OrganizerFieldDTO:
         return _session_field_dto(field)
 
 
+def _live_session_count() -> Count:
```
fancysnake: ## Thermo-nuclear code quality review

**`repositories/submissions.py` is at 1227 lines and this PR adds 65 more to the part that has outgrown the file.**

The file already holds six repositories. `FacilitatorRepository` is now ~265 of those lines and, after this PR, a genuinely separate concern from the proposal / field repos around it: identity reservation across dead rows (`read_including_deleted`, `find_by_ident`, `set_ident`, `slug_exists` all deliberately reaching `all_objects`), a lifecycle (`soft_delete`, `restore`, `delete`), and a concurrency guard (`lock`, `has_sessions`). Four of the module-level helpers at the top (`_readable_facilitators`, `_order_facilitators`, `_live_session_count`) serve only that class.

That is a seam, and the alive-vs-all manager rule is exactly the kind of invariant that wants a file boundary to state it once at the top rather than a comment per method scattered through a file about something else.

Split it in half along that seam: `repositories/facilitators.py` for the facilitator repo and its three helpers, the rest staying put. Same code, no behaviour change, and the next facilitator feature stops paying the cost of landing in a 1.2k-line file that is mostly about proposals.

hasparus:
## Stale-PR sweep

Flagged: no commits in 2+ days, `checks` and `tingle` failing on `7a67833`.

**`checks` — fixed.** `messages-check` was failing on a stale PL catalog (drifted `#, python-brace-format` flags + `POT-Creation-Date`). Regenerated via `mise run messages`.

**Merge conflict — resolved.** Merged current `main` in. Three conflicts:
- `PAPERCUTS.md` — pure log-append collision, took the union of both sides' entries.
- `django.po` — header timestamp only; regenerated after resolving.
- `templates/panel/facilitator-detail.html` — `main` independently re-indented this template (wrapped `{% block content %}` in a `<div class="p-3 sm:p-4">`) since this PR's base. Took `main`'s structure and re-applied this PR's three actual changes: gating the header "Edit" button and the empty-state edit link on `not facilitator.deleted_at`, and the deleted-facilitator restore banner.

**Migration collision — fixed.** `main` had independently landed `0142_guild_guildmembership` off the same `0141` parent this PR's `0142_facilitator_deleted_at` → `0143_facilitator_flag_to_soft_delete` chain used, so Django's test-db setup failed with "Conflicting migrations detected: multiple leaf nodes". Renumbered this branch's pair to `0143_facilitator_deleted_at` / `0144_facilitator_flag_to_soft_delete`, chained after `main`'s `0142` — no operations changed, dependency-only fix.

**`tingle` — left red, already justified by the PR author.** The 2026-08-07 comment on this PR explains the net debt (currently +133: `legacy-loc`/`old-subdomain-loc`, "no located additions") as inherent to where the touched files sit in the noun/verb migration, and that it "will stay red." Not re-litigating an already-accepted tradeoff.

**Review comments — none outstanding.** CodeRabbit's only comment was a Design System pre-merge check (raw `<button>` vs `tessera_button`); the 2026-08-07 comment on this PR says all twelve review threads were addressed and resolved in `6617ca4`, which is in this branch's history. No new CodeRabbit or Rams threads since.

**Verification:** full `mise run test:py` in an isolated worktree (to avoid cross-contamination with parallel work on other PRs) — **4023 passed, 9 skipped**, clean.

Pushed as `689471f` (catalog) → `86b7490` (merge + migration renumber). Not merged — left ready for review.

---

src/ludamus/links/db/django/repositories/sessions.py
fancysnake: Fixed. There is one facilitator row lock now: `FacilitatorRepository.lock(pks)`, alive-only, raising `NotFoundError` naming the missing pks — the contract `_lock_facilitators` had. `_lock_facilitators` is gone; `SessionRepository.create` / `set_facilitators` / `replace_facilitators_in_sessions` call the repo method, so the session repository no longer owns facilitator liveness. The panel delete passes `[facilitator.pk]`, and a row soft-deleted between its read and the lock now comes back `NotFound` from `lock` instead of from `soft_delete` a step later.

Repository tests cover an alive row, a dead one, a missing pk, and a missing pk mixed in with an alive one.

src/ludamus/links/db/django/repositories/submissions.py
fancysnake: Agreed, and filed rather than done here: the extraction touches `SessionRepository` and `DiscountRepository`, neither of which this PR is about, so it wants its own diff.

https://github.com/zagrajmy/ludamus/issues/873

src/ludamus/mills/submissions/engine.py
fancysnake: Already closed on this branch, so no change. The import restore is immediately followed by `log_facilitator_deletion(..., user_id=None, deleted=False)` — the same helper the panel service calls — so the History entry does not depend on which caller reached the repo. `tests/unit/test_mills.py::test_run_logs_the_restore_it_made` pins it.

The policy half is a fair point, but it is the import log’s question, not this thread’s: noted separately rather than answered here.

src/ludamus/gates/web/django/chronology/panel/views/facilitators.py
fancysnake: The hazard is already gone, so no change. `_apply` has an explicit `elif action == "restore":` and the `else` raises `ValueError` naming the unhandled action — the minimum this thread asked for. A fifth action added to `BULK_FACILITATOR_ACTIONS` without a branch now fails loudly instead of restoring the selection.

Leaving the chain as a chain: the lambda table is a style preference across three call sites and buys nothing beyond what the explicit failure already guarantees.

src/ludamus/templates/panel/facilitators.html
fancysnake: Fixed — the mode is gone, and the bin got a route rather than a section.

* `deleted` is out of `FacilitatorListFilters` and out of `FacilitatorListQuery`; `list_by_event` filters `Facilitator.objects` (alive only) with no inverted predicate.
* New `list_deleted_by_event` / `FacilitatorPanelService.list_deleted`, feeding a `Bin` tab (`panel:facilitator-bin`, `panel/facilitator-bin.html`) with its own pagination, its own empty state, and the one action that is legal there.
* `facilitators.html` lost the "Deleted only" checkbox, the bulk-bar gate, the row tint, the badge, and all three `deleted_at` conditionals in the action cell. Nothing on the live list knows deletion exists any more.

I went with a route rather than the inline "Recently deleted" section because of the counterweight you raised: the bin can be long, and the page-level bulk toolbar is a singleton (`#bulk-status-form`), so an inline second table would have cost bulk restore. Its own page keeps bulk restore and pagination, and leaves room for search/sort there later without a mode flag reappearing.

src/ludamus/links/db/django/repositories/submissions.py
fancysnake: Fixed both halves.

One definition: `has_any_session` is gone, replaced by `count_sessions(pk) -> FacilitatorSessionCountsDTO(live, deleted)`, annotated off one row and reusing `_live_session_count()` for the live half — so the guard and the Sessions column can no longer disagree. The service refuses on `counts.live or counts.deleted` and attaches the counts to `FacilitatorActionError`.

Actionable refusal: the single-facilitator delete now says how many, how many of them are deleted, and links where they are — "This facilitator is named on 2 sessions, 1 of them deleted. Remove them from all of these first — the deleted ones are in the *proposal bin*", the link pointing at `panel:proposals#deleted-proposals` (the heading got the anchor). With no deleted ones it drops to the plain count.

The bulk path keeps the static wording: it aggregates by reason across rows, so per-row counts have nowhere to go there.

src/ludamus/links/db/django/repositories/submissions.py
fancysnake: Split. `repositories/facilitators.py` now holds `FacilitatorRepository` plus the three helpers that only served it (`_readable_facilitators`, `_order_facilitators`, `_live_session_count`) and `_FACILITATOR_SORT_FIELDS`; `submissions.py` drops to 904 lines and is about proposals and fields again. No behaviour change — the only edits inside the moved code are the ones the other threads asked for.

The alive-vs-all manager rule is stated once at the top of the new module, as you suggested, instead of per method.

`repositories/__init__.py` re-exports from the new module, so `request.di.uow` / `inits` wiring is untouched; `SessionRepository` imports `FacilitatorRepository` directly for the shared lock.

PR #789, opened 2026-08-05
Stop clamping the organizer's participants limit to the category's bounds

hasparus:
An organizer couldn't raise a proposal's participants limit past 255 — a bare browser tooltip (*"Wartość nie może być większa niż 255."*) with no way past it. The 255 wasn't in code or schema: the panel form was copying the proposal category's `max_participants_limit` (a CFP setting someone had set to 255) onto the organizer's own input.

## Change

- The panel proposal form no longer applies the category's min/max to `participants_limit`. Those bounds are a call-for-proposals guardrail and now bind only the submission wizard, where their four-branch construction collapses to two. `create_proposal_form` takes `durations` instead of the whole category — bounds were the only other thing it read.
- The category minimum was the worse half: it forced the panel to render a stored `0` as blank so existing proposals stayed editable. Gone; the panel shows the stored value.
- One bound remains, on **every** form writing the column, and it's storage, not policy: Postgres `integer` tops out at 2³¹−1, dev sqlite doesn't, so an overflow only appeared in production. One shared `MaxValueValidator` (`STORAGE_LIMIT_VALIDATOR`) covers the panel form and the wizard — neither puts a `max` attribute back on an input.
- Help text `Empty or 0 = no limit` / `Puste pole lub 0 = bez limitu` on both surfaces; "blank means unlimited" was unguessable.

## Tests

- Panel edit page (the reported surface): saves 500 over a category max of 9; saves 1 under a category min of 4; rejects 2³¹ with a field error and no write. Each fails on `main`.
- Wizard unit test: 2³¹ rejected, no `max` attribute rendered.
- e2e: the panel input carries no `max` attribute and 500 passes `checkValidity()`.

## Known gap (own issue)

`min_age` has no form ceiling despite a 0–80 DB check constraint, and the facilitator self-edit write path handles no constraint errors at all — the pre-existing "forms don't mirror storage constraints" class, not expanded here.

## History

The branch explored and reverted deleting `max_participants_limit` outright (the min/max pair earns its keep for submitters), and merged main after #739. The `ios-regressions` deflaking that this branch stumbled into is split out to #801; this PR no longer touches those specs. Squash on merge.

## Summary by CodeRabbit

- **New Features**
  - Participant limits now support “empty or 0 = no limit” guidance.
  - Organizer-created forms can accept limits beyond proposal category bounds, within the supported storage limit.
  - Duration choices are updated consistently when selecting a proposal category.
  - Session edit forms now display participant-limit help text.

- **Documentation**
  - Updated participant-limit explanations and Polish translations to clarify proposal and organizer form behavior.


tests/integration/web/panel/test_proposal_create_page.py
```diff
@@ -385,6 +386,49 @@ def test_post_creates_session_with_unique_slug_on_collision(
         )
         assert new_session.slug != "my-new-session"
 
+    def test_post_accepts_a_limit_above_the_category_maximum(
+        self, authenticated_client, active_user, sphere, event
+    ):
```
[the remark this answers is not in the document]
hasparus: Skipping this one. The CLAUDE.md rule is about call sites a human writes — a call with three bare positional arguments is unreadable, so the declaration forces keywords. Pytest fixtures have no such call site: the framework resolves them by parameter name and passes them as keyword arguments already.

Your own AST scan bears that out — every other test method in this file declares its fixtures positionally. Making one new test the exception would leave the file inconsistent without making any call clearer.

The test this pointed at is gone anyway: the regression coverage moved to `tests/integration/web/panel/test_proposal_edit_page.py`, since the bug was reported against the edit page and both pages build the form through the same `_ProposalFormBase._build_form`.

---

src/ludamus/gates/web/django/chronology/forms.py
```diff
@@ -50,20 +50,22 @@ def build_session_details_form(
...
+        "min_value": min_limit,
+    }
+    if max_limit:
         participants_kwargs["max_value"] = max_limit
+    elif not min_limit:
+        participants_kwargs |= {
+            "required": False,
+            "initial": 0,
+            "help_text": _("Empty or 0 = no limit"),
+        }
```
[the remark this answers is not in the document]
hasparus: This is wrong because we'll always have limitless no sign up sessions

scripts/ios-regressions/scrubber-longpress.test.ts
```diff
@@ -21,55 +25,211 @@ const {
...
+  new Set([...html.matchAll(pattern)].flatMap((m) => (m[1] ? [decodeEntities(m[1])] : [])));
+
+// Take the accessible names the page actually renders rather than guessing at
+// them. Two earlier rounds guessed: `viewport.width - 4` assumed where the rail
+// sits, and matching a label ending in a time assumed only the rail carries one
+// — the device then pressed the schedule body's own "11:00" headings in the
+// left gutter. The served markup is ground truth for both the rail and the
+// session links, needs no geometry, and is already translated, so this holds
+// whether or not the message catalogue is ever compiled.
+const RAIL_HOUR_NAMES = /class="schedule-rail-hour[^"]*"[^>]*aria-label="([^"]+)"/g;
```
hasparus: **Blocking — these names are read in a different language than the device renders.**

`LocaleMiddleware` is enabled with `LANGUAGE_CODE = "pl"` and `LANGUAGES = [pl, en]` (`edges/settings.py:134,272,274`). Bun's `fetch` sends **no** `Accept-Language` — I dumped the headers it actually emits: `accept`, `accept-encoding`, `connection`, `host`, `user-agent`, nothing else — so `fetchReadyPage` gets the **pl** rendering.

The rail's accessible name is `Jump to {{ day }} {{ time }}` with `day = day.first_start|date:'l'`, and that weekday comes from **Django's own** catalogue, which ships compiled. Verified in a shell here:

```
pl 'Sobota'    'Jump to %(day)s %(time)s'
en 'Saturday'  'Jump to %(day)s %(time)s'
```

So `railNames = {"Jump to Sobota 11:00", …}` while an English simulator reports `"Jump to Saturday 11:00"`. `pressable`'s `names.has(node.label)` never holds and `railHourTargets` throws on both attempts.

The escape hatch closes the other way: if the simulator *is* Polish, then `calloutSignals` (`"Hide preview"`, `"Open in New Tab"`, …) never match and the whole spec passes vacuously. Nothing in `mobile.yml` or the harness pins either side.

Why `SESSION_LINK_NAMES` transfers and this doesn't: `test:e2e:prep` never runs `messages-compile`, so `"Open details for %(title)s"` falls back to its English msgid under both languages — which is exactly the run that produced the "110 session links" evidence cited at line 77. The comment's "already translated, so this holds whether or not the message catalogue is ever compiled" is true of the *project* catalogue and false of Django's.

Remedy: pin the language on both sides. Cheapest is `fetch(url, { headers: { "Accept-Language": "en-US,en;q=0.9" } })` inside `fetchReadyPage`, matching what the simulator sends, with the constant living next to `calloutSignals` so the two English assumptions are stated once. Sturdier still: set the simulator's language explicitly in `mobile.yml` and derive both from that.

---

src/ludamus/gates/web/django/chronology/forms.py
```diff
@@ -50,20 +50,22 @@ def build_session_details_form(
...
+    # The only place a category's participant bounds bind — the panel form and a
+    # facilitator's self-edit are deliberately unbounded — so a category of large
+    # rooms can refuse a two-person session. A floor also makes the number
+    # mandatory; with neither bound the field is optional and 0 means no limit.
+    participants_kwargs: dict[str, Any] = {
+        "label": _("Max participants"),
+        "min_value": min_limit,
+    }
+    if max_limit:
         participants_kwargs["max_value"] = max_limit
```
hasparus, line 69: The four-branch collapse is behaviour-preserving — I walked all four inputs: `(0,0)` → `min_value=0, required=False, initial=0, help_text`; `(min>0, 0)` → `min_value=min`; `(0, max>0)` → `min_value=0, max_value=max`; `(min>0, max>0)` → both. Good.

But the ceiling this PR adds is missing here. When the category sets no maximum — the default — `participants_limit` is `IntegerField(required=False, min_value=0, initial=0)` with **no** `max_value`, and it writes the same `PositiveIntegerField` (`integer` on Postgres) that `SessionEditForm.clean_participants_limit` now guards. `clean_participants_limit`'s own rationale is that the ceiling is storage, not policy — so it belongs on every form that writes the column, and this is the *public* one, reachable by anyone submitting through the call for proposals. That is the surface where an unhandled `DataError` is most expensive.

Remedy: import `MAX_STORED_PARTICIPANTS_LIMIT` and add the same check to the details form (a `clean_participants_limit` on the generated class, or `max_value=MAX_STORED_PARTICIPANTS_LIMIT` when `max_limit` is 0 — the latter re-adds a `max` attribute, which is harmless on the wizard since the wizard is the surface where bounds are supposed to bind). Better still per GLIMPSE: put the storage invariant in the service that writes the column, so the Google Docs import path (`google_docs_import.py:55` carries `participants_limit` too) gets it for free.

---

scripts/ios-regressions/modal.test.ts
```diff
@@ -196,7 +221,7 @@ beforeAll(async () => {
   await clickNodeCenter(closeButton);
   await client.command.wait({ ...deviceOptions, durationMs: 1000 });
 
-  if (await hasVisibleText("Close")) {
+  if (await isShowing("Close")) {
```
hasparus: This one swapped an over-strict check for an under-strict one. `hasVisibleText("Close")` asked "is the label anywhere in the tree"; `isShowing` now also requires the rect centre inside the band. For the *content* probe that is the right trade — for "did the modal close" it is backwards: a modal that failed to close but whose X sits outside `80 … height-120` now reads as closed, and closure is the thing this test exists to prove. Practically the X lands near mid-screen (the dialog is `max-h-[90dvh]` and centred), so it will usually be caught — but the assertion no longer *guarantees* it.

Remedy: keep `snapshotLabels()` for `closeIssue` and reserve `isShowing` for `contentInitiallyVisible`.

Separately, this file and `scrubber-longpress.test.ts` now carry two different answers to the same question. `isNodeInViewport` (line 82) uses `80` top / `120` bottom off `viewportOf(snapshot).height` and ignores `viewport.y`; `centreOnScreen` uses `CHROME_INSET = 120` on both edges and *does* add `viewport.y`. They cannot both be right about how much top chrome Safari has. The `hittable=…/rect=…` node describer is now written out three times too (here as `describeNode`, twice in the scrubber). Hoist one `centreOnScreen(snapshot, node)` and one `describeNode` into `harness.ts` with a single inset constant and a single derivation of the viewport, and let both specs consume them — otherwise the next device finding gets applied to one file and not the other, which is how the `hittable` round went.

---

tests/integration/web/panel/test_proposal_edit_page.py
```diff
@@ -514,6 +521,146 @@ def test_post_ignores_category_from_other_event(self, panel_client, sphere, even
         session.refresh_from_db()
         assert session.category_id == original_category_id
 
+    def test_post_accepts_a_limit_above_the_category_maximum(
+        self, authenticated_client, active_user, sphere, event
```
hasparus: Merge resolution: worth converting. `panel_client` (`tests/integration/web/panel/conftest.py:27`) is exactly `authenticated_client` + `sphere.managers.add(active_user)` — the same three lines all three new tests hand-roll. This file already uses `panel_client` 120 times after #739; main left two old-style holdouts, and this PR adds three more, moving the file back the way main just moved it.

None of the three needs `sphere` or `active_user` for anything but the manager add, so the conversion is mechanical for each:

```python
    def test_post_accepts_a_limit_above_the_category_maximum(self, panel_client, event):
        bounded = ProposalCategory.objects.create(...)
```

drop `sphere.managers.add(active_user)`, and swap `authenticated_client` → `panel_client` at the call site.

---

scripts/ios-regressions/scrubber-longpress.test.ts
```diff
@@ -21,55 +25,211 @@ const {
...
+): Promise<CaptureSnapshotResult> => {
+  const deadline = Date.now() + timeoutMs;
+  let snapshot: CaptureSnapshotResult | null = null;
+  do {
+    // A snapshot taken while Safari is still launching throws; that is a state
+    // to wait out, not to end the run on.
+    try {
+      snapshot = await takeSnapshot();
+    } catch (error) {
+      console.warn("Snapshot failed while the page was settling; retrying.", error);
```
hasparus: The retry is right, but the error it eventually throws loses the reason. If every snapshot in the 90s window throws, `snapshot` stays `null` and the failure reads `None of the N rendered session names appeared in the snapshot (0 nodes).` — which says the tree was empty when in fact it was never read. On a branch that has learned four separate things from CI error text, that is the one message you cannot afford to be misleading.

Remedy: keep the last error and fold it into the throw, the way `fetchReadyPage` already does with `lastError`:

```ts
let lastSnapshotError: unknown = null;
// …
} catch (error) { lastSnapshotError = error; console.warn(…); }
```

then in the final `throw`, when `snapshot === null`, report `no snapshot ever succeeded: ${String(lastSnapshotError)}` instead of the `(0 nodes)` branch.

---

hasparus (commented):
**CHANGES REQUESTED** — the Django payload is clean and regression-free, but two mechanisms in the iOS/forms work were verified in an environment they do not transfer from: the rail's accessible names are read over a connection that negotiates a *different language* than the device does, and the storage ceiling this PR introduces lands on only one of the two forms that write the column.

## Blocking

**1. `RAIL_HOUR_NAMES` cannot match on the device — the two clients negotiate different languages.**
`LocaleMiddleware` is on, `LANGUAGE_CODE = "pl"`, `LANGUAGES = [pl, en]`. Bun's `fetch` sends no `Accept-Language` (verified by dumping the headers it emits), so `fetchReadyPage` reads the **pl** page. The rail label is `Jump to {{ day }} {{ time }}` and `day` is `|date:'l'`, which resolves through **Django's own** always-compiled catalogue: `'Sobota'` (pl) vs `'Saturday'` (en). An English simulator therefore reports a label that is not in `railNames`, and `railHourTargets` throws on every attempt.

The alternative is not better: if the simulator is Polish, `calloutSignals` (`"Hide preview"`, `"Open in New Tab"`, …) never match and the spec passes vacuously. Nothing pins either side.

`SESSION_LINK_NAMES` transfers only by accident — `test:e2e:prep` never runs `messages-compile`, so `"Open details for %(title)s"` falls back to its English msgid in both languages. That is the run the "110 session links" evidence came from. The comment's "already translated, so this holds whether or not the message catalogue is ever compiled" is true of the project catalogue and false of Django's. Remedy inline.

**2. The storage ceiling covers the panel form but not the public wizard.**
`SessionEditForm.clean_participants_limit` rejects values wider than `integer`, with the right rationale — the ceiling is storage, not policy. But `build_session_details_form`, whose four-branch block this PR rewrote, still builds an `IntegerField` with no `max_value` whenever the category sets no maximum, and it writes the same column from the *public* call-for-proposals surface. Remedy inline; the sturdier fix is to put the invariant in the service that writes the column, which also covers the Google Docs import.

## Worth fixing before merge

3. **Two visibility conventions, one question.** `modal.test.ts:isNodeInViewport` uses 80/120 off `height` and ignores `viewport.y`; `scrubber:centreOnScreen` uses 120/120 and adds `viewport.y`. They cannot both be right about Safari's top chrome. The `hittable=…/rect=…` describer now exists in three copies. Hoist one of each into `harness.ts`.
4. **`isShowing("Close")` loosened the assertion that matters.** Rect-band visibility is right for the content probe and wrong for "did the modal close" — that check should stay on the full label set.
5. **Merge resolution.** The three new edit-page tests hand-roll `authenticated_client + sphere.managers.add(active_user)` in a file that uses `panel_client` 120 times after #739. Mechanical conversion, no behaviour change.
6. **`waitForSchedule` swallows the snapshot error** and can report `(0 nodes)` for a window in which no snapshot ever succeeded.

## Verified, no action needed

- **`decodeEntities` ordering is correct and complete.** Django's `escape()` emits exactly `&amp; &lt; &gt; &quot; &#x27;`; all five are handled and `&amp;` last is the right fix — a served `&amp;lt;` contains no `&lt;` substring, so the early passes cannot touch it, and the final pass yields the literal `&lt;`. `&#39;` is redundant (Python's `html.escape` emits `&#x27;`) but not wrong, and it is three tokens in a six-line decoder.
- **`SESSION_LINK_NAMES` matches the real templates.** Checked all three renderers of `a.session-link` (`_compact_session_row.html`, `_room_lane_tile.html`, `_session_card.html`). `[^>]*` matches newlines in JS, so the attribute block spanning four source lines is fine, and `\s*` covers the indent before `<span class="sr-only">`. `_session_card.html` also carries an `aria-label` — same text, so no divergence. Nothing in the current markup breaks it.
- **`CHROME_INSET = 120`: the comment's claim holds.** Walked the short-rail case: the rail is `sticky top-20` and `self-start`, so a 5-hour rail occupies roughly screen y 148–230 against a 120–754 band. Being conservative does only shift which markers get pressed. The comment stops one step short: at the degenerate extreme *no* marker survives, and `railHourTargets` throws rather than passing on an empty target list. That is the right behaviour and worth one clause in the comment.
- **No Django regression after the merge.** Full suite: 3711 passed, 9 skipped. `messages-check`: extraction current, all 1749 translated. `tingle check`: no new debt, 58 paid off. `ts-check:ios` clean. The `max_participants_limit` delete/revert cancels out — no model or migration in the net diff.
- `EVENT_PATH` moving to `/event/…` is a real improvement: `/chronology/event/…` is a permanent redirect (`urls.py:66`).
- The T2 waitlist comment is now accurate — `session` is not refreshed before the comparison, so `session.participants_limit` is genuinely the old value.
- `"Nabór zgłoszeń"` is the catalogue's term and the new CfP copy uses it consistently.

## Nits

- Accessible names are whitespace-collapsed by the platform; `namesFrom` keeps raw markup whitespace. A `.replace(/\s+/g, " ").trim()` is cheap insurance against a seeded title that wraps.
- `panel/proposal-form.html:82` still says the swapped region exists because "the category rebuilds them" — after this PR only `duration` is category-dependent; `participants_limit` no longer is.
- `session-edit-form.html` hand-rolls the help `<p>` instead of going through tessera, so it renders help text *and* the error together, where `render_help_text` suppresses help on error. It also has no `id`/`aria-describedby` tie to the input.
- `_("Enter a smaller number.")` names no bound. The string it replaced ("That is more people than we can store.") said less about what to do but more about why.
- `harness.ts:28-31` — "which it falls back off when the image has no such device" doesn't parse. The substance is right: `mobile.yml` prefers `iPhone 17 Pro` and falls back to the first available iPhone.
- The scrubber has no positive control: nothing proves the callout *would* be seen if `-webkit-touch-callout:none` were dropped. Absence-as-pass plus an English-only signal list is a lot of trust in one direction.

---

src/ludamus/gates/web/django/chronology/forms.py
```diff
@@ -50,20 +55,31 @@ def build_session_details_form(
...
+    participants_kwargs: dict[str, Any] = {
+        "label": _("Max participants"),
+        "min_value": min_limit,
+        "validators": [
+            MaxValueValidator(
+                MAX_STORED_PARTICIPANTS_LIMIT, message=_("Enter a smaller number.")
+            )
+        ],
+    }
+    if max_limit:
```
hasparus, line 68: **Blocker: this commit turns `tingle check` red.** At `1e02e54` it reported "🎉 no new debt" (+1). At `f3d6df2`:

```
old-subdomain-loc (line_count): +35 / -20 = +15
check failed: metrics grew by a net +14 (policy: sum)
```

The dedicated `tingle` CI job (the one checked out with `fetch-depth: 0`) is failing on this head. The `checks` job going green is not counter-evidence — it checks out without history, so tingle can't find the merge-base there.

All of it is this file: `chronology/forms.py` +30/−14 against `proposal_edit.py` +5/−6. The two easiest pieces to move out of `chronology/` are the 8-line explanatory comment and this inline `MaxValueValidator` construction. Hoisting the validator to a module-level constant in `gates/web/django/forms.py` next to `MAX_STORED_PARTICIPANTS_LIMIT` and importing that instead pays most of the +15 back and fixes the duplication flagged below in one move.

The rest of this rewrite verifies clean: the validator sits outside the branch so it binds in all four (min, max) combinations, and the collapsed two-branch form is case-for-case equivalent to the old four-branch one, including `required` and `initial`.

---

src/ludamus/gates/web/django/forms.py
```diff
@@ -632,31 +635,33 @@ class SessionEditForm(forms.Form):
...
+        required=False,
+        min_value=0,
+        label=_("Participants Limit"),
+        help_text=_("Empty or 0 = no limit"),
     )
     min_age = forms.IntegerField(required=False, min_value=0, label=_("Minimum Age"))
     duration = forms.CharField(required=False, label=_("Duration"))
     cover_image = cover_image_field()
 
+    def clean_participants_limit(self) -> int | None:
```
hasparus: One rule, two implementations, same user-facing string. The comment's premise — "no `max_value`, because a max attribute is what this form set out to remove" — argues against `max_value`, not against a validator, and the wizard already proves the validator route keeps the attribute off. Django 6.0.7's `IntegerField.widget_attrs` reads only `self.min_value` / `max_value` / `step_size`; `validators` never reach the widget.

So this whole method is:

```python
validators=[MaxValueValidator(MAX_STORED_PARTICIPANTS_LIMIT, message=_("Enter a smaller number."))],
```

on the field declaration above. Same error, same key, no attribute, one implementation instead of two that can drift apart. Share the validator instance with `build_session_details_form` and the LOC growth flagged in `chronology/forms.py` mostly goes away too.

Minor, separate: when a category *does* set a max, a value over both bounds now collects two errors — Django's "Ensure this value is less than or equal to 6." plus "Enter a smaller number." Cosmetic, but worth knowing it's reachable.

---

scripts/ios-regressions/modal.test.ts
```diff
@@ -22,12 +29,33 @@ const {
...
+// heading not hittable while the very next step tapped it; the field reads
+// false inside Safari's web content. A rect centred in the viewport is the
+// check that both survives the device and still fails for content parked
+// below the fold of a modal that no longer sizes itself.
+const showingLabels = async (): Promise<string[]> => {
+  const snapshot = await takeSnapshot();
+  return snapshot.nodes.flatMap((node) =>
+    node.label && node.rect && isNodeInViewport(snapshot, node) ? [node.label] : [],
+  );
+};
```
hasparus: Flagging the environment question this branch keeps getting bitten by, since `ios-regressions` is still in progress on `f3d6df2` and this is not yet device-verified.

The prior round framed this as "80 top → 120 top." For the content probe it's larger than that. The old check was `hasVisibleText`, which read `snapshotLabels()` — the **whole tree, no band at all**. The new one filters through `centreOnScreen`, so the assertion that passed on device now has to clear a band that is narrower at the top (80 → 120) *and* offset by `viewport.y`, which the modal's old band never applied. `waitUntilShowing` requires both `"About this session"` and the description inside that band simultaneously.

Failure direction is a false *failure*, not a false pass, which is the right way round — and the 400 ms polling absorbs the animation. So this isn't wrong, it's unconfirmed: if the green iOS run lands on this head, it's settled and I'd drop the point.

Two things that did check out while reasoning about the band:

- `scrollUntilTriggerInViewport` is safe under the narrower band. It spans 634 px on an 874 px screen against a 450 px scroll step, so it can't be jumped over and can't oscillate; the direction heuristic still resolves correctly for a node above the band.
- `CHROME_INSET` is exported from `harness.ts` but imported nowhere. Either something was meant to use it or it should stay module-private.

---

hasparus (commented):
## CHANGES REQUESTED

Verification round on `f3d6df2`. **Both prior blockers are genuinely fixed.** One new blocker came in with the fix commit: it turns a green CI gate red.

### New blocker

**`tingle check` fails on this head.** At `1e02e54` it reported "🎉 no new debt" (+1). At `f3d6df2`: `old-subdomain-loc +35 / −20 = +15`, net **+14**, `check failed`. The dedicated `tingle` job (checked out with `fetch-depth: 0`) is red. The `checks` job going green isn't counter-evidence — it checks out without history, so tingle can't find the merge-base there. Entirely from `chronology/forms.py` (+30/−14) against `proposal_edit.py` (+5/−6). Line comment has the concrete way to pay it back, which is the same edit that fixes the duplication below.

### Prior findings — verified

**Blocker 1, language negotiation — resolved.** Header is on the fetch call itself. `LANGUAGES = [("pl", …), ("en", …)]`, so `get_supported_language_variant("en")` matches exactly rather than falling through the generic path; `LocaleMiddleware` is enabled and a bare Bun fetch carries no session or `django_language` cookie, so Accept-Language decides. The comment is accurate.

I walked both mismatch directions and found no vacuous pass left. Rail `aria-label` is `Jump to {day} {time}` with `day` from `date:'l'`, and the session-link `sr-only` is `Open details for {title}` — both carry a translated prefix, so neither set survives a language flip. A Polish device makes `waitForSchedule` throw after 90 s and `railHourTargets` throw. That also closes the English-only `calloutSignals` hole transitively: a non-English device makes Safari send a non-English Accept-Language, so the page never matches the names and the spec dies before the callout check is ever reached.

**Blocker 2, storage ceiling — resolved.** Validator sits outside the branch, so it binds in all four (min, max) combinations, and the two-branch rewrite is case-for-case equivalent to the old four-branch one including `required` and `initial`. No `max` attribute reaches the input: Django 6.0.7's `IntegerField.widget_attrs` reads only `min_value`/`max_value`/`step_size`, and `IntegerField.__init__` appends its own validators *after* `Field.__init__` seeds `self.validators` from the kwarg, so the passed list survives intact. `propose/parts/details.html` renders `max` from `field.max_value`, which is now `None`. Import is gates→gates, no layering problem. `test_rejects_a_limit_wider_than_the_column` is sound (`2**31` > `2_147_483_647`, and it pins `max_value is None`). Caveat in the line comments: the rule is now implemented twice.

**Item 3, one visibility convention — done, one unconfirmed tightening.** Details in the line comment; the trigger-scroll loop is provably safe under the narrower band.

**Item 4, modal close checks the full tree — resolved.** Also tightened from substring to exact match, which now agrees with `findNodeByLabel`/`waitForLabel`.

**Item 5, panel tests on `panel_client` — resolved and passing.** `test_proposal_edit_page.py` + `test_proposal_create_page.py` + `tests/unit`: 3712 passed, 9 skipped. `lint:mypy`, `lint:ruff`, `ts-check:ios` all clean.

**Item 6, `waitForSchedule` no-snapshot error — resolved**, quotes `lastSnapshotError`.

**Nits — all done.** `collapse()` is applied at all three lookup sites (`pressable`, the rejected-node report, and `waitForSchedule`'s `matched`), not just the first. `decodeEntities` puts `&amp;` last and the reasoning in its comment is correct. Fallback-viewport comment, `proposal-form.html` comment, and the dead `isShowing` all handled.

### Fresh eyes, non-blocking

- The new e2e test is not vacuous, which was worth confirming given the pattern on this branch. `_resolve_category` falls back to `categories[0]` on a fresh create, and frostfire-con's only category is `max_participants_limit=6` (`bootstrap_data.py:404`), so on `main` the input really does render `max="6"` and `checkValidity()` on 500 is false. The comment's "6 in this fixture" checks out.
- `session-edit-form.html` renders the new help text as a bare `<p>` with no `aria-describedby` pointing at `#id_participants_limit` — the hand-rolled input misses what `tessera_field` does for the panel. Small gap on a line this PR touches.
- `harness.ts`: "the runner always sets a rect" is asserted rather than shown, and every call site still guards `node.rect &&`. Same class of claim the branch has been paying for; either the guards are unnecessary or the comment overstates.
- The T2 waitlist comment is accurate — `session` is the pre-`apply` DTO, so `session.participants_limit` really is the old value at that point.

Fix tingle and this is done. The two blockers were closed properly rather than papered over.

---

src/ludamus/gates/web/django/chronology/forms.py
```diff
@@ -44,26 +49,28 @@ def build_personal_data_form(
...
+    # facilitator's self-edit are deliberately unbounded — so a category of large
+    # rooms can refuse a two-person session. A floor also makes the number
+    # mandatory; with neither bound the field is optional and 0 means no limit.
+    participants_kwargs: dict[str, Any] = {
+        "label": _("Max participants"),
+        "min_value": min_limit,
+        "validators": [STORAGE_LIMIT_VALIDATOR],
+    }
+    if max_limit:
         participants_kwargs["max_value"] = max_limit
```
hasparus, line 69: Non-blocking, pre-existing (the old four-branch version had it too, and `test_only_max_set_accepts_zero` pins it): when a category sets only a max, a submitter can still type `0` — which means *unlimited* — and sail under the ceiling. So `max_participants_limit=10` bounds every value except the one that means infinity. Now that this block is the single place the bounds bind and carries the comment explaining the policy, that loophole reads like it was chosen rather than inherited. If it is intended (0 as "organizer will decide"), one clause in the comment would settle it; if not, it's a one-line `min_value=1`-when-max-is-set fix — either way, fine as a follow-up, not this PR.

---

hasparus (commented):
Thermo-nuclear pass on the post-split diff (`origin/main...2417bfa`, 13 files, +312/−104). Verdict: **approve** (submitted as a comment — the review account is the PR author, so GitHub refuses a formal approval).

This is what the bar looks like when it's met — the PR fixes the bug by deleting the mechanism that caused it, and the structure comes out simpler than it went in.

**What holds up under pressure:**

- `_participants_limit_field` is gone entirely, not relocated. `create_proposal_form` narrowed its contract from `category: ProposalCategoryDTO | None` to `durations: Sequence[str] = ()` — the optionality and the `category.durations if category else []` fallback inside the builder died with it, and the caller in `proposal_edit.py:282` now states the one thing it actually needs. That's the boundary getting *more* explicit, not less.
- `build_session_details_form(requirements, *, category=...)` deletes three kwargs at both call sites (`chronology/views.py:388,762`) and the four-branch kwargs table collapses to one `if`/`elif` whose two branches map one-to-one onto the two real policies (ceiling set vs. fully unbounded). The comment above it earns its lines — it names *why* the bounds bind here and nowhere else.
- `STORAGE_LIMIT_VALIDATOR` is one shared constant used by both surfaces, with the rationale (Postgres `integer` vs. dev sqlite, panel savepoint vs. self-edit 500) written where the constant lives. Validator-not-`max_value` is the right mechanism and the unit test pins the distinction (`max_value is None`, no rendered `max`).
- The `session.participants_limit or None` pre-fill hack in `proposal_edit.py` is deleted rather than commented harder — the workaround died with the constraint that necessitated it. That's the shape of fix this skill exists to demand.
- Tests land in the right layers per the project's strategy: form-builder truth table in unit, panel round-trips (over-max saves, under-min saves, 2³¹ rejected with no write) in integration with `assert_response`, the rendered `max` attribute in e2e — and the e2e comment correctly notes the attribute, not the Python field, is what blocked the user. The fixture's `max_participants_limit=6` matches the "6 in this fixture" claim.

**Checked and clean:** full suite passes (3712 passed, 9 skipped); `tingle stat --diff` is net negative (typing −1, refactoring −5); no file crosses 1k lines (`chronology/views.py` was over before and shrinks; the big test file was already over); pl translations updated including the reworded cfp-edit copy, which now tells organizers what the bounds actually govern.

**One non-blocking line note** on the surviving `if max_limit:` branch — a pre-existing `0`-means-unlimited loophole under a category max that now reads as deliberate policy; worth one clause of comment or a follow-up, not a change here.

The `min_age`/self-edit constraint gap is already named in the description as its own issue; agreed that it's the same class, not this PR's job.

---

scripts/ios-regressions/harness.ts
```diff
@@ -25,28 +25,51 @@ export const hookTimeoutMs = Number(env.IOS_HOOK_TIMEOUT_MS ?? "300000");
...
+// pass. The runner's `hittable` is NOT this check — it reads false inside
+// Safari's web content (device-falsified on this branch).
+const CHROME_INSET = 120;
+
+export const centreOnScreen = (rect: Rect, viewport: Rect): boolean => {
+  const centreY = rect.y + rect.height / 2;
+  return (
+    centreY >= viewport.y + CHROME_INSET && centreY <= viewport.y + viewport.height - CHROME_INSET
+  );
+};
```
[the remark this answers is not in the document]
hasparus: Valid point, but the code left this PR — the iOS spec work now lives in #801, where `scripts/ios-regressions` carries the current `centreOnScreen`. Applied there in 35055d7: the horizontal centre must now sit inside the viewport as well.

---

tests/integration/web/chronology/test_forms.py
```diff
@@ -1,11 +1,14 @@
...
 from ludamus.gates.web.django.chronology.forms import build_session_details_form
+from tests.unit.factories import category
 
 
 class TestBuildSessionDetailsFormParticipantLimits:
     def test_both_limits_zero_makes_field_optional(self):
-        form_class = build_session_details_form([], min_limit=0, max_limit=0)
+        form_class = build_session_details_form(
+            [], category=category(min_participants_limit=0, max_participants_limit=0)
+        )
```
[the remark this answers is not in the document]
hasparus: Done in 08fd994. The strategy doc settles it: a test that builds a form is not pure, so the whole `build_session_details_form` file (which predates this PR) moves to `tests/integration/web/chronology/test_forms.py` unchanged, not just the lines this PR touched.

---

tests/integration/web/chronology/test_forms.py
```diff
@@ -128,8 +145,27 @@ def test_both_limits_set_enforces_range(self):
...
+                "title": "Test",
+                "description": "A test session",
+                "display_name": "Presenter",
+                "participants_limit": str(2**31),
+            }
+        )
+
+        assert not form.is_valid()
+        assert form.errors["participants_limit"] == ["Enter a smaller number."]
+        assert form.fields["participants_limit"].max_value is None
```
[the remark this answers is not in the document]
hasparus: Done in 08fd994 — `test_accepts_the_widest_limit_the_column_can_store` submits `2**31 - 1` and asserts it validates and cleans to the same value.

---

src/ludamus/gates/web/django/forms.py
hasparus, line 1: Non-blocking follow-up: the validator's comment frames the storage bound as belonging on "every field writing the column", and this PR covers both `participants_limit` surfaces — but `ProposalCategoryForm.min_participants_limit` / `max_participants_limit` (lines 346–347, untouched here) write same-width `PositiveIntegerField` columns with no ceiling, so an organizer typing 2³¹ into the CFP settings still hits the raw DataError path. Pre-existing gap, out of this PR's scope, but now that `STORAGE_LIMIT_VALIDATOR` exists as the canonical fix it's a two-line follow-up — worth an issue so it doesn't get lost.

---

tests/integration/web/chronology/test_forms.py
```diff
@@ -1,11 +1,12 @@
-"""Unit tests for build_session_details_form with participant limit parameters."""
-
 from ludamus.gates.web.django.chronology.forms import build_session_details_form
+from tests.unit.factories import category
```
hasparus, line 2: Note, not a blocker: this is now the only integration test importing from `tests.unit.factories` (verified by grep). The import is right — the `category` factory builds a DTO, exactly what a form-builder test needs, and duplicating it would be worse — but its home under `tests/unit/` is now slightly misnamed since both trees consume it. If a second integration test ever reaches for it, consider promoting the DTO factories to `tests/factories.py`. Nothing to do in this PR.

---

hasparus (commented):
## Thermo-nuclear review, round 2 — verdict: APPROVE

(GitHub forces this to COMMENT because the account authors the PR — read it as an approval.)

The prior thermo-nuclear round approved this PR at `2417bfa`. This pass audited the two commits added since, then re-reviewed the full diff against `main` at head `08fd994`. The earlier approval stands.

### Delta since the approved head

**`15a2822` (merge of main).** Verified with `git show --remerge-diff`: the only conflict was `django.po`, resolved by keeping main's POT-Creation-Date and both sides' obsolete `#~` entries. No production code was touched by hand in the resolution; nothing surprising.

**`08fd994` (review-bot follow-ups).** Three small changes, all verified:
- The `build_session_details_form` test file moved to `tests/integration/web/chronology/test_forms.py` unchanged except for the dropped docstring. This is what `docs/TESTING_STRATEGY.md` requires — "a test that builds a form is not pure" — so the move is a correction, not churn. The old `tests/unit/gates/.../chronology/` directory still holds genuinely pure-helper tests, so no orphaned tree.
- `test_accepts_the_widest_limit_the_column_can_store` pins the inclusive boundary (2³¹−1 validates), the missing happy-path twin of the 2³¹ rejection. Both sides of `STORAGE_LIMIT_VALIDATOR` are now tested, not just the failing one.
- One comment clause documenting that 0 stays a valid unlimited marker under a category ceiling — written at the single place the bounds bind, matching the maintainer-confirmed policy. The code path it describes is real: with `max_limit` set, `min_value=0` still admits 0.

### Fresh full-diff pass

The structural story is unchanged and still good: the four-branch `participants_kwargs` construction collapsed to two, `_participants_limit_field` deleted outright, `create_proposal_form` narrowed from the whole category DTO to the `durations` it actually reads, and `build_session_details_form` widened to the category DTO that all its callers already held — both signatures moved toward their real data dependency. All callers migrated (grep-verified: no stragglers on the old signatures). The one bound that survives is storage, not policy, expressed once as a shared validator that renders no `max` attribute — which is exactly the invariant the e2e test asserts.

Two non-blocking notes left inline:
1. `ProposalCategoryForm.min_participants_limit` / `max_participants_limit` write the same-width columns without the storage validator — pre-existing gap, natural two-line follow-up now that the canonical validator exists.
2. The moved test file is the sole `tests/integration` → `tests.unit.factories` import; the sharing is correct, the factory's home is now slightly misnamed. Future cleanup, not this PR.

### Verification run
- `tests/integration/web/chronology/test_forms.py`: 18 passed at the new location.
- Panel proposal create + edit pages: 114 passed, 1 skipped.
- Full unit suite: 974 passed (factory signature change absorbed cleanly).
- `tingle stat --diff`: net negative (−26 refactoring, −1 typing); no debt counter grows.

No file approaches 1k lines, no new branching, no wrappers, no boundary leaks. The delta since approval is exactly what it claims to be.

---

PR #792, opened 2026-08-05
Give the panel's phone layout its width and height back

hasparus:
Started as "the panel dashboard on mobile needs some love" and became the pattern behind it: the panel spent **130px of chrome above content on every page** (177px with tabs), then nested padding under it — `p-4` outside a card whose body added 24px put content 40px from each edge of a 440px screen, 60px where a card sat inside a tab shell.

**The dashboard itself is not in this PR.** Two attempts at it were worse than what was there and it is reverted to `main`, byte for byte. Everything below is independent of it.

Measured at 440px before touching anything:

| page | chrome | content width | lost to padding | padding chain |
| --- | --- | --- | --- | --- |
| dashboard | 130px | 168px | 272px | `grid:16 > card:16` |
| cfp-edit | 130px | 338px | 102px | `p-4:16 > card-body:24` |
| facilitators | 130px | 296px | 144px | `p-4:16 > flex-1:12 > card:32` |
| proposals | 176px | 328px | 112px | `p-4:16 > flex-1:12 > bg:16` |

## Sections run full-bleed and stack as bands

| before | after |
| --- | --- |
|  |  |

`.panel-bleed` is a real class in `index.css`, next to `.card`. Below `sm` it drops the wrapper's padding entirely; the cards give up the side border and corners they can no longer sit inside, and the gaps between them close. Each card also drops its own top border and leans on the rule above — the header's `border-b` for the first, the previous card's for the rest — so a closed gap draws **one** 1px divider instead of stacking two. That avoids a negative margin, which the panel lints against.

It lives in `@layer utilities` rather than `components` because bled cards carry `mt-6` and only a same-layer rule with two classes of specificity outranks it. The wrapper's own padding sits behind `:where()` for the opposite reason: it is a default a page may want to change, and at full specificity it would silently beat a `p-*` utility written on the same element.

Only `card-body`'s own 16px now separates content from the edge, instead of 12 + 16 + a border.

**Only standalone pages bleed.** `facilitators`, `proposals` and `facilitator-merge` put their wrapper inside a tab shell, which already frames and pads its content — bleeding there stranded a card with only top and bottom rules floating inside another border.

| | before | after |
| --- | --- | --- |
| chrome above content | 130px | **122px** |
| header height | 73px | **63px** |
| page title | 20px / 28 | **17px / 21.25** |
| header → first section | 12px | **0px** |
| section → section | 24px | **0px** |
| cfp-edit content width | 338px | **362px** |

## Header CTAs go icon-only

| before | after |
| --- | --- |
|  |  |

Below `sm` the glyph carries the action and the button squares off to **44 × 44** — a deliberate tap-target floor, not whatever the padding produced. "Call for Proposals" and its subtitle now fit whole where "Call for Propos…" did not.

The label is `sr-only`, **not** `hidden`, so the buttons keep their accessible names — verified `["New Category"]` and `["Cancel","Save"]`. Hiding it means selecting it, and a bare text node cannot be selected — so **all 31 hand-rolled header CTAs moved to `{% tessera_button %}`**, which emits the `.btn-label` span itself. Two unit tests pin that contract. Their `mr-2` went too, since `.btn` already sets `gap: 0.5rem` and was double-spacing all along.

Routing them through the tag also settled a size split the panel was carrying: a hand-rolled `btn btn-primary` picks up `.btn`'s own 0.625rem/1rem padding at the inherited font size, while `tessera_button` renders `px-4 py-2 text-sm`. Headers mixing both — proposal-detail's Edit sat in a row of tag-rendered status buttons — now agree.

**Everything about how a header button behaves on mobile is decided in `panel/base.html`, once:**

- the label goes `sr-only`, and the button squares to 44 × 44 — both gated on `:has(> svg)`. A button with no icon has nothing to fall back on, so hiding its label would leave a blank pill, which is what proposal-detail's Accept / Hold / Reject / Delete bar would have become.
- `min-h-11` covers *every* button, icon or not. An iconless one keeps `py-2` and rendered **81 × 38** next to 44 × 44 squares — under the tap-target floor and 6px shy of its own row. Now 81 × 44.
- `w-auto` cancels the `max-md:w-full` that `tessera_button` adds to form-submit buttons by default. That default is right in a body form and never right in this strip, so 22 header templates had each been passing `full_width_mobile=False` to say so; the rule now lives with the other three.

Buttons with no icon keep their label. Text "Cancel" beside an icon-only "Save" is deliberate — there is no glyph that reads as "cancel" without one.

## Tab strip

An even 6px inset below `sm` — top and sides, never the bottom. The selected tab is pulled a pixel over the bar's border so it reads as attached to the panel below, and padding under it floats the whole strip off its rail.

The body edge moves with the sides. Shrinking only the bar relocates the misalignment rather than removing it, since the first tab is meant to sit flush with the content under it:

| viewport | bar padding | body px | `tab.left − content.left` |
| --- | --- | --- | --- |
| 440 | `6 / 6 / 0 / 6` | 6px | **0.0px** |
| 1280 | `6 / 12 / 0 / 12` | 24px | −12px (pre-existing, untouched) |

## Padding, everywhere else

- `card-body` drops to 16px below `sm`. 24px is right once a card is one column of several; full-bleed on a phone it is 48px of 440.
- **11 templates restated `card-body`'s padding as `p-4 sm:p-6` literals** on the card itself, so a change to `card-body` would never have reached them — mostly the sphere create/edit/delete forms plus the space and discount ones. They now carry the real class.
- **13 more divs restated `.card` itself** — `bg-bg-secondary rounded-2xl shadow-… border border-border`. Three of them with a static `shadow-sm`, which reads too weak against the dark-mode-aware `--shadow-card`. All collapsed onto `.card`.
- The nine panel page wrappers and the tab shell body go to 12px on mobile.

| before | after |
| --- | --- |
|  |  |

Desktop is unchanged throughout — 77px header, 20px title, 16 and 24px gaps, labelled buttons:

## Defects the screenshots turned up

**Date inputs overflowing their card.** iOS Safari renders `datetime-local` as a native picker whose intrinsic width exceeds a phone viewport, and `width: 100%` will not shrink it past that. Normalised once in the base layer for `date`/`time`/`datetime-local`/`month`/`week` rather than per widget — several are built by Django `DateTimeInput`s that pass no class at all (`EnrollmentWindowForm` among them), so a per-template fix would have left the rest broken. Plus `min-w-0` on the CFP grid items so the single-column track shrinks with them.

**The content log's empty text sat flush against the viewport edge** while the heading above it was inset 12px. A bled page has no card-body to inset its non-card content, so each such element asks for `max-sm:px-3` itself; that one was missed.

**The discounts export button was wider than a phone viewport** and overflowed its right edge. Its label is long enough that no toolbar alignment saves it, so it stretches on mobile instead.

**Column labels.** "Statistics" labelled cells reading `0 / 1` (accepted / all proposals) → **Accepted**. Its neighbour "Submissions" held the submission window's state and dates → **Submission window** / "Okno zgłoszeń". `makemessages` fuzzy-matched that to "Otwarcie zgłoszeń" ("Submissions open"); corrected by hand.

## Tooling

- **`.venv/bin` onto PATH for agent shells.** `mise.sandbox.toml` disables the `python` tool, so mise generates no `python` shim and a bare `python` fell through to the image's 3.11. That reads as a repo bug the first time it meets 3.14-only syntax — PEP 758's unparenthesised `except A, B:` in `scripts/impeccable_lint.py` raises `SyntaxError` there, and "fixing" it fights Black, which normalises back to it under 3.14. `black`/`pytest`/`mypy` have the same trap in reverse.
- **Vendored the impeccable design skill**, and taught the linters to skip the vendored trees. `lint:impeccable` scans tracked HTML/CSS/JS and impeccable's detector carries the CSS patterns it hunts for as string literals, so it reported its own rulebook as four gradient-text findings; `lint:codespell` had `./.claude` in its skip list, which only matches when codespell is handed the repo root. `hk.pkl` excludes them too — oxfmt restyled ~100 of impeccable's scripts on the first commit attempt, and the whitespace fixer edited Markdown playbooks where trailing spaces are hard line breaks.
- Papercuts logged: `mise run lint:impeccable` silently drops trailing path arguments, and the interpreter trap above.

## Checks

`djlint`, `black`, `ruff`, `mypy`, `codespell`, `ast-grep`, `import-linter`, `oxlint`, `hk` and the impeccable detector all clean. `tingle`: no new debt, **87 paid off** across 33 metrics. `messages-check` fresh, 1749 translated, 0 fuzzy. **3719 Python tests pass**, 9 skipped.

## One thing I could not verify

**The iOS overflow does not reproduce here.** Playwright's WebKit on Linux sizes these inputs like Chromium — measured `spill = -25px`, identical with and without the fix. The change is the standard remedy for a well-understood UA behaviour and is inert in every engine I could test, but it wants a real iPhone before you trust it. (This branch's `ios-regressions` job passes; I have not checked whether it asserts anything about these inputs.)

## Left, deliberately

- The CFP open/closed badge and event dates are still `hidden sm:flex`. Surfacing them costs ~24px on *every* panel page's sticky header, which is the opposite of what this PR is for.
- `card p-8 text-center` is the empty-state pattern, used consistently across the panel. Its 32px is a deliberate choice for an empty state, not a restatement of `card-body` — collapsing it would shrink every one of them. Worth extracting into tessera as its own change.
- **18 standalone create/edit form pages** — sphere announcements and connections, integrations, facilitator/track/field create and edit — have the same shape as `cfp-edit` but are not bled. Their card padding is fixed above; whether they should also run full-bleed is a design call, not a mechanical one.
- Two dashboard numbers still disagree with the pages they link to: *Facilitators* counts `Count("presenter_id", distinct=True)` over `Session` while the page it links to lists `Facilitator` rows, and `total_sessions` is `pending + scheduled` rather than a total. Wrong joins, not layout.


hasparus:
There's too much room between "Call for Proposals" or "Venues" and the text below them. On desktop a little bit, but definitely on mobile. We should do a follow-up @claude.

PR #794, opened 2026-08-05
feat(enrollment): keep the way out open after enrollment closes

hasparus:
The session modal dropped its whole footer when no enrollment config was active, so a participant who could no longer attend had no control to give their seat back — only a green "You are enrolled in this session" banner. Reported from a live event, two minutes before the session started.

The backend already allowed it: `SessionEnrollPageView._validate_request` short-circuits an all-cancels request so it works with every config deleted, and the anonymous service validates with `require_active_enrollment=False`. Only the UI was missing.

## What changed

`build_enroll_actions` (`gates/web/django/event/enroll_presentation.py`) makes one decision — what, if anything, the viewer may do about their seat — and returns an `EnrollActions` carrying the action to post plus the copy that goes with it. Both render paths call it: the modal's GET, and the HTMX fragment `SessionEnrollPageView.post` swaps in.

A shut window now hides the ways *in* and keeps the way *out*:

- **Enrolled or waiting** → Cancel / Leave stay reachable, behind an `hx-confirm` that says the step is one-way.
- **Full and closed** → the warning states both facts: the seat goes to the next person waiting, *and* it can't be taken back. (Previously the closed message won and dropped the handover — the very reason the exit exists.)
- **Session already over** → nothing to hand over, so no controls; the info tab states the seat in words instead.
- **The group page stays linked**, relabelled "Manage the seats you booked for others…", because seats booked for companions and party members can only be released there.

The modal's standalone status banner now renders only when the footer has no badge of its own, so the two never repeat the same sentence.

## Notes for review

- `is_ended` deliberately gates only the closed-window branch. An organizer who leaves a window open past the end time keeps the controls that go with it — that stays their call, and it preserves prior behaviour exactly.
- `enroll_actions` is computed in `SessionModalComponentView`, not on `SessionData`, because the event page patches `is_ended` onto its cards *after* construction — a property would have been silently wrong on a card.
- The one-click control acts on the viewer alone, while the modal counts a companion's seat as theirs. Pre-existing divergence, now documented at the call site rather than papered over.
- **`tingle` is red at +13.** Deliberate — see [the comment below](https://github.com/zagrajmy/ludamus/pull/794#issuecomment-5217795451) for the accounting and the options.

## Tests

- Unit: the full state matrix for `build_enroll_actions` — posted action, badge, confirm copy, group label, and the closed/ended guards.
- Integration: both render paths, including the HTMX path with the window shut and with the session ended — the combination the two definitions used to drift in. Expected `EnrollActions` are spelled out literally where re-deriving them through the function under test would be circular.
- E2E (`resign-after-close`, `enroll-states`): all four seat states from the modal footer, the one-way confirm text, and the past-session banners. Destructive specs take a fresh seeded session per retry so a flake can't cascade.

https://claude.ai/code/session_01Kobi4Bk3XTtHruJjJx45FK


hasparus:
`tingle` is red and I am not fixing it by pushing more code — it needs a call from you.

It reports a net **+11**: `old-subdomain-loc +12`, `legacy-loc −1`. The bulk of the new code (~92 lines) already moved out of `chronology/` into `gates/web/django/event/`, which took it from +104 to +11. What is left is the seam that connects the two — two import lines and a ten-line `SessionData.enroll_actions` property, in the package the metric exists to drain.

Every way I found to shrink it further trades design for the counter:

- a `SeatFacts` protocol parameter instead of five keyword arguments gets to ~+6
- demoting the property to a context variable the modal view supplies gets to ~+2
- migrating `SessionData` out of `chronology/` would clear it outright, but it has 17 importers and belongs in its own change

So this is a justified addition rather than a manufactured offset, per the note in CLAUDE.md. Happy to take any of the three if you would rather have the number than the shape — the last one is the only one I would argue for on its own merits, as a separate PR.

Every other check is green, including `test`, `checks`, `impeccable`, `test-postgres`, and `ios-regressions`.

---

hasparus:
## `tingle` is red at net **+13**, deliberately — needs a call

Everything else on `ad558b6` is green. This one I have not "fixed", because every remaining way to clear it trades design for the counter, which `CLAUDE.md` explicitly rules out ("Don't game a counter to keep a number flat").

```
old-subdomain-loc (line_count): +14
legacy-loc (line_count):         -1
check failed: metrics grew by a net +13 (policy: sum)
```

### What was already paid down

The presenter (`SeatState`/`EnrollActions`/`build_enroll_actions`, ~92 lines) started in `gates/web/django/chronology/` and now lives in `gates/web/django/event/enroll_presentation.py`, outside the counted range. That took the branch from **+104 → +13**. `legacy-loc` is net negative: `adapters/web/django/views.py` shrank (1773 → 1772) because the decision moved out of the legacy view.

### What is left, and why

The residue is the seam that hands the presenter's output to the modal — currently 15 lines in `SessionModalComponentView.get`. It cannot leave the counted range, because **every candidate home is inside `chronology/`**:

- `chronology/event_presentation.py` (as a `SessionData` property) — where it was, at +12. Moved out because it was a latent bug, not for the counter: `is_ended` is a plain field that the event page patches onto its cards *after* construction (`adapters/web/django/views.py:714`), so the property was only correct on modal-built instances, with nothing enforcing that. A card template reading it would have offered a live Cancel on a finished session.
- `chronology/views.py` (where it is now) — same counted package, +14.

Worth flagging explicitly: moving it to the view was the right call on design merit, but it is **two lines worse** on the metric, not better.

The only change that genuinely clears this is migrating `SessionData` itself out of `chronology/` — ~150 lines with 17 importers. That is the migration `old-subdomain-loc` exists to drive, and it deserves its own PR rather than riding along here.

### Options

1. Accept as a justified addition (the escape hatch `CLAUDE.md` describes: *"A justified addition is fine; say so in the PR"*).
2. I open a follow-up moving `SessionData` to `gates/web/django/event/`, and this rebases onto it.
3. Something else you'd prefer.

I did not want to pick option 1 on your behalf, since it means merging with a red check.

---

PR #795, opened 2026-08-05
Claude/remove timetable confirm actions

fancysnake:
## Summary by CodeRabbit

* **New Features**
  * Timetable sessions now show visual states for conflicts, preferred-slot violations, and normal scheduling.
  * Conflict indicators identify both sessions involved, including shared-room bookings across tracks.
  * Session cards display scheduling, confirmation, presenter, and duration details with improved accessibility.
* **Bug Fixes**
  * Invalid track filters now return an error instead of showing unfiltered results.
  * ISO-8601 session durations display correctly.
* **Changes**
  * Removed “Confirm all” and “Confirm block” timetable actions; individual confirmation remains available through the confirmations tab.
  * Updated Polish terminology for held proposals.


src/ludamus/templates/panel/parts/timetable-session-detail.html
```diff
@@ -67,26 +67,15 @@ <h3 class="font-semibold text-foreground text-sm">{% translate "Session details"
...
-                    {% tessera_button t_unconfirm variant="secondary" button_type="submit" size="sm" %}
-                {% else %}
-                    <div class="text-xs text-foreground-muted mb-2">{% translate "Program item not confirmed" %}</div>
-                    <input type="hidden" name="confirmed" value="true">
-                    {% translate "Confirm program item" as t_confirm %}
-                    {% tessera_button t_confirm variant="primary" button_type="submit" size="sm" %}
-                {% endif %}
-            </form>
+            {% comment %}
+            Read-only here on purpose: confirmation is set on the confirmations
```
fancysnake: The comment says confirmation "is set on the confirmations tab" — but that tab cannot reach every item.

`ConfirmationsConfirmActionView` 422s unless both `facilitator_pk` and `track_pk` are digits, and `EventConfirmationsService.set_confirmed` resolves the facilitator before touching anything. `list_with_scheduled_session_in_track` only surfaces facilitators reachable through a track. So after this PR:

- a scheduled session with **no facilitator** can never be confirmed;
- a scheduled session in **no track** can never be confirmed.

The service already knows about the first group — `count_without_facilitator` feeds `without_facilitator_count`, which the page renders as a number the organizer is being asked to care about and given no control to act on. This toggle was the only way to clear it.

Either give the confirmations tab a path for facilitator-less items first, or keep a per-item control here until it has one. Shipping a counter with no corresponding action is worse than the duplication you're removing.

src/ludamus/mills/timetable.py
```diff
@@ -65,7 +65,9 @@ def _slot_windows_by_grid_date(
 
 
 def _position_sessions(
-    items: list[AgendaItemDTO], event_start: datetime
+    items: list[AgendaItemDTO],
+    event_start: datetime,
+    own_item_pks: set[int] | None = None,
```
fancysnake: This is a nullable mode bolted onto an existing flow, and it doesn't need to be one.

`None` and "every item is mine" are the same thing here. Compute the set unconditionally in `build_grid` and the optionality disappears from three signatures at once:

```python
all_items = self._uow.agenda_items.list_by_event(event_pk)
own_item_pks = (
    {item.pk for item in all_items}
    if track_pk is None
    else {item.pk for item in self._uow.agenda_items.list_by_track(track_pk)}
)
```

Then: `own_item_pks: set[int]` here, no default, no `| None` on `_build_day_grid` either, and line 103 collapses to `is_foreign=item.pk not in own_item_pks`. One concept instead of "a set, or the absence of a set, which means something the reader has to reconstruct."

Two more things while this signature is open:

- Once this takes a third parameter it is keyword-only per the project rule, and the call at line 265 went the other way — `_position_sessions(items_for_space, grid_start, own_item_pks)` dropped the `event_start=` keyword it used to have. Three bare positionals at a call site is exactly what that rule exists to prevent.
- The filtered path now costs two queries (`list_by_event` + `list_by_track`) where it used to cost one. Acceptable, but if `AgendaItemDTO` carried its track pks you'd need neither the second query nor the set.

src/ludamus/mills/timetable.py
```diff
@@ -161,10 +164,15 @@ def build_grid(
         dates_to_render = (
             available_dates if date_selection == "all" else [date_selection]
         )
-        all_items = (
-            self._uow.agenda_items.list_by_track(track_pk)
-            if track_pk is not None
-            else self._uow.agenda_items.list_by_event(event_pk)
+        # The grid always shows everything scheduled in the spaces on screen.
```
fancysnake: This comment contradicts one 350 lines below it, in this same file:

> `list_all_for_track`: *"Overlaps are checked against every scheduled item in the event so a track page still surfaces cross-track clashes."*

Cross-track room clashes are already detected and already listed in the conflicts panel on a track-filtered page. "Two tracks could be put in the same room at the same time without either organizer seeing the clash" is not what was happening.

What this change actually buys is worth having, and it is narrower: **occupancy is visible before you create the clash, not only after.** Conflict detection is post-hoc — you learn about the collision once you've made it. Please say that instead, otherwise the next reader has two comments in one module telling incompatible stories about the same invariant and no way to tell which one is stale.

fancysnake (commented):

**The deletion half is excellent.** `SessionConfirmationService`, three views, three URLs, two repo methods and ~590 lines of test gone, `mills/chronology.py` back under 1000 lines (1008 → 973). The deleted `confirm_all_by_event` / `confirm_all_by_track` were near-duplicates of the canonical `set_confirmed_for_facilitator`, which already does scoped one-statement updates. Good riddance.

**The addition half is where the problems are.** The `is_foreign` feature is a second, unrelated change bundled into a removal PR, and it is carrying most of the risk in this diff with none of the surrounding scrutiny.

### Presumptive blockers

1. **Functional hole from the removal.** Every remaining write to `session_confirmed` goes through `ConfirmationsConfirmActionView`, which hard-requires `facilitator_pk` *and* `track_pk`. An item whose session has no facilitator (or no track) now has **no path to confirmation at all** — and `EventConfirmationsService` counts exactly those items as `without_facilitator_count` and shows the number as a to-do it cannot action. The per-item toggle you deleted was the only way to clear them. This needs a path on the confirmations tab before the toggle goes.

2. **The `is_foreign` plumbing is a nullable mode threaded through three signatures.** `set[int] | None = None` plus `own_item_pks is not None and item.pk not in own_item_pks` encodes "no filter" twice for no gain. There is a code-judo move that deletes the optionality entirely (see inline).

3. **The template card is copy-pasted wholesale under a new top-level `{% if %}`.** The per-session conflict/slot-violation conditional chain was already repeated 4×; the diff now duplicates the whole card so a reader holds two near-identical 25-line blocks in their head. The state belongs on `SessionPositionDTO`, not in four repeated template lookups.

4. **The feature partly undercuts its own goal.** The foreign card deliberately drops the conflict and slot-violation styling, so the clash it exists to surface renders as the dimmest, lowest-contrast element on the grid. And the stated rationale contradicts a comment 350 lines below it in the same file — `ConflictDetectionService.list_all_for_track` already checks overlaps against every item in the event *"so a track page still surfaces cross-track clashes"*.

### Also

- Unrelated i18n churn rides along (a Polish retranslation of "On hold" and mass loss of `python-brace-format` flags).
- `mise run lint` passes clean; I did not run the Python or e2e suites.

Inline comments have the specifics and the suggested restructurings.

src/ludamus/templates/panel/parts/timetable-day-grid.html
```diff
@@ -12,41 +12,63 @@
...
-                                ⚠
-                            {% elif pos.agenda_item.session_id in slot_violation_session_pks %}
-                                ⏰
-                            {% endif %}
-                            {% if pos.agenda_item.session_confirmed %}
-                                {% translate "Confirmed" as t_confirmed %}
-                                <span aria-label="{{ t_confirmed }}" title="{{ t_confirmed }}">{% icon "lock-closed" variant="mini" class="w-3 h-3 inline" %}</span>
-                            {% endif %}
-                            {{ pos.agenda_item.session_title|truncatechars:256 }}
+                {% if pos.is_foreign %}
```
fancysnake: This duplicates the entire session card — wrapper div, the four CSS custom properties, body, title, presenter — to change classes and drop three attributes. Two near-identical 25-line blocks, and any future change to card layout has to be made twice or silently diverges.

Worse, the branch you copied already carried the same conditional chain 4×:

```
{% if pos.agenda_item.session_id in conflict_session_pks %}…{% elif … slot_violation_session_pks %}…{% else %}…{% endif %}
```

on the wrapper, the title, the presenter and the duration. `is_foreign` is a *fourth* independent styling axis stacked on that, and it got a whole duplicate card rather than joining the model.

The judo move: the per-session visual state is one value, not four lookups against two context sets. Compute it in `TimetableService` — `SessionPositionDTO` already exists and already knows the item:

```python
state: Literal["foreign", "conflict", "slot_violation", "normal"]
```

(the service already loads conflicts for the page; `ConflictDetectionService` and `TimetableService` live in this same module). Then the template is one card with `{% if pos.state == "foreign" %}` guarding only the three interactive attributes, and the colour picked once from `state`. That deletes the duplicated card, the four repeated chains, and `conflict_session_pks` / `slot_violation_session_pks` from the template context — and makes "foreign" a natural fourth value instead of a new axis.

This file is only 75 lines, so it isn't a size problem yet. It is a "the fifth state will be copy-pasted too" problem.

src/ludamus/templates/panel/parts/timetable-day-grid.html
```diff
@@ -12,41 +12,63 @@
...
+                    {# Another track already booked this room and time. Read-only: it is
+                       here to prevent a clash, not to be edited from someone else's track. #}
+                    {% translate "Scheduled by another track" as t_foreign %}
+                    <div class="timetable-session timetable-session-foreign absolute rounded-md overflow-hidden border border-dashed border-border bg-bg-tertiary text-xs opacity-60"
```
fancysnake: Reusing `.timetable-session` on a non-interactive element makes it interactive to the JS — in exactly the wrong direction.

`timetable.ts` suppresses placement clicks on that class:

```ts
if (armed && !target.closest(".timetable-session")) { /* place the armed session */ }
```

So in assign mode a click landing on a foreign block does **nothing at all**: no placement, no detail pane (there's no `hx-get`), no feedback. Before this PR that region was empty and clickable. Meanwhile `dragover`/`drop` resolve via `.timetable-column`, so dragging into the identical pixels still places the session fine.

Same pixels, two opposite answers depending on how you got there. That's the kind of incidental control flow that produces a bug report nobody can reproduce.

Pick one and make it explicit: either give the ghost its own class so assign-mode clicks pass straight through to the column (occupancy is advisory, you can still schedule against it), or block the drop as well. My preference is the former — the card is informational, it shouldn't be a dead zone.

src/ludamus/templates/panel/parts/timetable-day-grid.html
```diff
@@ -12,41 +12,63 @@
...
+                         aria-label="{{ t_foreign }}: {{ pos.agenda_item.session_title }}"
+                         title="{{ t_foreign }}: {{ pos.agenda_item.session_title }}"
+                         style="--offset: {{ pos.start_minutes }};
+                                --duration: {{ pos.duration_minutes }};
+                                --lane-start: {{ pos.lane_start_pct }}%;
+                                --lane-width: {{ pos.lane_width_pct }}%">
+                        <div class="timetable-session-body px-1 py-0.5 h-full overflow-hidden">
+                            <div class="timetable-session-title font-medium wrap-anywhere text-foreground-muted">
```
fancysnake: The foreign branch ignores `conflict_session_pks` and `slot_violation_session_pks` entirely, so a foreign booking that *is* the clash renders as `opacity-60`, dashed, muted — the least attention-grabbing element on the grid.

Worth checking against the actual highlight wiring while you're here: the view sets `conflict_session_pks = {c.session_pk for c in conflicts}`, and `session_pk` is the **other** side of the pair (`subject_session_pk` is the item that was examined). With a track filter, subjects come from `list_by_track`, so the pks in that set are frequently the foreign items — which until this PR were never rendered, meaning the ⚠ marker on a track-filtered grid was landing on nothing. Now they are rendered, in the one branch that doesn't check the set.

So the card that should be shouting is the one you've made quietest. Once the state lives on the DTO (see the comment on line 15) this resolves itself: `conflict` should win over `foreign` for colour, with the dashed border and the missing interactive attributes carrying "not yours".

src/ludamus/pacts/chronology.py
```diff
@@ -384,6 +376,9 @@ class SessionPositionDTO(BaseModel):
     duration_minutes: int
     lane_start_pct: float = 0.0
     lane_width_pct: float = 100.0
+    # Outside the track being filtered on: shown so the space's occupancy is
+    # honest, but not editable from here.
+    is_foreign: bool = False
```
fancysnake: Two concerns with this field.

**It's the second notion of "foreign track" in this feature.** `ConflictDetectionService._foreign_track_attribution` already answers "is this session someone else's, and whose?" — and answers it better, with the track's name and its managers, precisely so *"organizers know whom to talk to"*. The grid's version is a boolean built from a separate `list_by_track` query, and it renders as an anonymous "Scheduled by another track". An organizer who spots a collision here still has to go find who owns it.

If the ghost card carried the track name ("Board Games Track") the feature would be complete rather than a hint that sends you elsewhere — and it would reuse the canonical helper instead of standing up a parallel one.

**The name describes the request, not the data.** Whether an item is "foreign" depends entirely on which track the *viewer* filtered by; the same row is foreign on one request and not on the next. That's fine as a rendering flag, but `is_foreign` on a shared DTO reads like a property of the item. `is_outside_filtered_track` (or folding it into a `state` field per the day-grid comment) says what it actually means and won't get misread by whoever wires up the next consumer.

src/ludamus/locale/pl/LC_MESSAGES/django.po
```diff
@@ -1626,7 +1617,7 @@ msgstr "Zaakceptowane"
 #: src/ludamus/templates/panel/parts/confirmation-status-badge.html
 #: src/ludamus/templates/panel/timetable-overview.html
 msgid "On hold"
-msgstr "Wstrzymane"
+msgstr "Rezerwa"
```
fancysnake, line 1629: Unrelated to this PR: "On hold" → `Rezerwa`, "Hold" → `Do rezerwy`, "Proposal put on hold." → `…przeniesione do rezerwy`. That's a product decision about proposal-status vocabulary riding into a timetable PR where nobody is looking for it. Split it out.

Separately, this regeneration stripped `#, python-brace-format` from ~10 entries (`"Enrolled: {}"`, `"Session proposal '{}' submitted successfully!"`, …). That flag is what makes gettext validate brace placeholders between msgid and msgstr; dropping it silently removes a safety check on the strings most likely to break at runtime. Looks like a local xgettext version difference rather than intent — please regenerate per the documented `mise run _lint -- env … django-admin` workflow so the only catalog delta in this PR is the four obsoleted confirm strings and the one new `"Scheduled by another track"`.

src/ludamus/templates/panel/parts/timetable-session-detail.html
fancysnake: Right, and the counter is worse than you describe: `confirmation-orphan-warning.html` renders `without_facilitator_count` and links to **"Open the schedule"** — the page whose only control this PR deleted. The dead end was one click long.

Confirmed both holes: `set_confirmed` is facilitator-keyed for *every* scope including `all` (`_read_facilitator_in_event` → `set_confirmed_for_facilitator`), and `list_with_scheduled_session_in_track` filters `sessions__tracks=track_pk, sessions__agenda_item__isnull=False`, so a trackless item is not even counted.

Per-item toggle is back in f1dd7be2 — view, URL, `SessionConfirmationService.set_session_confirmed`, DI, and `test_timetable_confirm.py`. The e2e spec goes back to exercising confirm/undo. "Confirm all" and "Confirm track" stay deleted; those were the ones being hit with the wrong scope in mind. The service now carries a comment saying why it is keyed on the agenda item rather than a facilitator, so the next person to tidy it finds the reason.

src/ludamus/mills/timetable.py
fancysnake: Done in f1dd7be2. `own_item_pks` is computed unconditionally (`{item.pk for item in all_items}` when nothing is filtered), so the `| None` is gone from all three signatures and the flag is a plain `item.pk not in own_item_pks`. `_position_sessions` is keyword-only, and the call site got its keywords back — you were right that it had *lost* `event_start=`.

The signature did not stay open long: ruff caps `build_grid` at five arguments and it was already at five, so the track filter and the two highlight sets now travel as one `GridMarksDTO(track_pk, conflict_session_pks, slot_violation_session_pks)`. Same idea as your suggestion, one rung further up.

Second query stands for now — `AgendaItemDTO` carrying its track pks is a wider change than this branch. There is now a *third* read in the filtered path (`list_track_names_by_session`, for the owner name — see the `is_foreign` thread), skipped entirely when nothing is outside the filter.

src/ludamus/mills/timetable.py
fancysnake: Fixed in f1dd7be2 — the comment now says occupancy is visible *before* the clash is created, and credits `list_all_for_track` with the after-the-fact detection instead of contradicting it. The commit message carried the same overclaim; that is corrected too.

src/ludamus/templates/panel/parts/timetable-day-grid.html
fancysnake: Taken, with one deliberate difference. In f1dd7be2 `SessionPositionDTO` carries `state: Literal["normal", "conflict", "slot_violation"]`, resolved once in `TimetableService`, and the duplicate card is gone — one card, one colour chain, and `conflict_session_pks` / `slot_violation_session_pks` are out of the template context.

Where I diverged: "foreign" is not a fourth value of that enum. It is not mutually exclusive with the other three — a booking can be someone else's *and* the clash, which is exactly the bug in your next comment. Folding it in would have made that case unrepresentable. So ownership stays a separate boolean, `is_outside_filtered_track`, and it drives only the dashed border and the missing interactive attributes; `state` drives colour. Two orthogonal things instead of four lookups.

One wart: djlint collapses the opening tag onto a single long line, because a `{% if %}` sitting *between* attributes defeats its attribute formatter. Its output, not a choice — but worth knowing before the next edit.

src/ludamus/templates/panel/parts/timetable-day-grid.html
fancysnake: Confirmed and fixed the way you preferred. `timetable.ts:285` reads `if (armed && !target.closest(".timetable-session"))`, so the ghost carrying that class made an armed click a no-op while `dragover`/`drop` resolving via `.timetable-column` still placed onto the identical pixels.

In f1dd7be2 the ghost gets `.timetable-session-foreign` and nothing else, so assign-mode clicks pass straight through to the column — occupancy is advisory, you can still schedule against it and conflict detection will say so. The geometry rule in `timetable.css` now matches both selectors, so it still positions itself.

Two things that fell out while checking: `timetable-session-foreign` had no CSS rule and no JS reference at all — it was a dead hook — and `.timetable-session:hover` sets `opacity: 1; z-index: 10`, so hovering un-dimmed the ghost and lifted it over real cards. Both fixed by the class no longer being shared. The e2e spec now asserts the ghost never carries `.timetable-session`.

src/ludamus/templates/panel/parts/timetable-day-grid.html
fancysnake: Confirmed, including the part about the highlight wiring: `conflict_session_pks = {c.session_pk for c in conflicts}` and `session_pk` is the counterpart, not `subject_session_pk`. With a track filter the subjects come from `list_by_track`, so those pks are frequently the foreign items — which means before this PR the ⚠ on a track-filtered grid really was landing on rows that never rendered.

Fixed in f1dd7be2 via the `state` field: conflict wins for colour, the dashed border and the absent interactive attributes carry "not yours", and the dimming is dropped for anything that is not `normal` so the loud card is actually loud. The unit test now asserts a foreign booking that is also the clash comes back `("Theirs", True, "Board Games Track", "conflict")`.

Separately, the grid *partial* passed `conflict_session_pks: set()`, so the marker vanished on every HTMX swap while the conflicts panel next to it listed the clash. Both now run the same detection — at the cost of one more `list_all_for_track` per refresh, alongside the one the conflicts panel already does. Say the word if you would rather have the cheaper inconsistency back.

src/ludamus/pacts/chronology.py
fancysnake: Both taken in f1dd7be2.

Renamed to `is_outside_filtered_track`, with the comment saying it is named for the request rather than the row.

On the anonymous ghost: it now carries the owning block's name. `SessionPositionDTO.outside_track_name` is filled from `list_track_names_by_session` (the query the confirmations service already uses), and the card renders it in place of the presenter — from a foreign block the useful answer is which block to talk to, not who is presenting. Title and `aria-label` say "Scheduled by track: Board Games Track", falling back to the old anonymous string only when a session has no track at all.

I did not route it through `_foreign_track_attribution`: that one answers per *conflict pair* and also resolves managers, which is a heavier read than the grid needs for every ghost on screen. Same fact, cheaper path — happy to converge them if you would rather have managers on the card too.

src/ludamus/locale/pl/LC_MESSAGES/django.po
fancysnake: Split into its two halves.

**The `python-brace-format` stripping** — already fixed by 6265ae76 before you got here; the diff against `main` no longer touches those 11 entries. Diagnosis was right: the local xgettext drops the flag, `main` and CI keep it. It bites on every regeneration, so `mise run messages` now has to be followed by reverting the catalog and hand-applying only the real msgid deltas. Logged in PAPERCUTS.md so the next person does not rediscover it. `mise run messages-check` still fails locally for exactly this reason and exactly these 11 lines — CI is the authority here.

**The `Rezerwa` vocabulary change** — staying. This branch is the batch of small fixes raised in one meeting, so the proposal-status wording belongs with the rest of it rather than in a PR of its own.

The catalog delta is otherwise what you asked for: the new `Scheduled by track: %(track)s`, the restored `Undo confirmation` / `Confirm program item` (the toggle is back), and `Confirm block` / `Confirm all` left obsoleted.

src/ludamus/mills/timetable.py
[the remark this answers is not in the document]
fancysnake: Valid — fixed in f1dd7be2. `_position_sessions` is keyword-only and the call site passes every argument by name; it had also silently lost the `event_start=` keyword it used to have.

src/ludamus/mills/timetable.py
```diff
@@ -161,10 +164,15 @@ def build_grid(
...
+        # Filtering the items by track hid the other tracks' bookings, so two
+        # tracks could be put in the same room at the same time without either
+        # organizer seeing the clash.
+        all_items = self._uow.agenda_items.list_by_event(event_pk)
+        own_item_pks = (
+            None
+            if track_pk is None
+            else {item.pk for item in self._uow.agenda_items.list_by_track(track_pk)}
```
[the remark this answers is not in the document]
fancysnake: Valid, with one correction: this is not new here. `list_space_pks(track_pk)` and `ConflictDetectionService.list_all_for_track(track_pk)` already took the same unvalidated pk before this PR, and the impact is a blank grid rather than a leak — a foreign track's spaces do not intersect this event's leaves.

The gap is real though, and it is upstream of `build_grid`: `get_track_filter_context` in `panel/views/base.py` only checked `track_param.isdigit()`. Fixed there in f1dd7be2 — the pk must be in this event's own tracks, which the helper already loads, so it costs no extra query and covers all six pages that use the track switcher rather than just this one. An out-of-event pk falls back to the unfiltered view instead of 404ing: it is a filter parameter, and a stale bookmark degrading is better than an error page.

src/ludamus/templates/panel/parts/timetable-day-grid.html
```diff
@@ -12,41 +12,63 @@
...
-                            {{ pos.agenda_item.presenter_name }}
-                        </div>
-                        <div class="timetable-session-meta timetable-session-duration {% if pos.agenda_item.session_id in conflict_session_pks %}text-danger{% elif pos.agenda_item.session_id in slot_violation_session_pks %}text-warning{% else %}text-info{% endif %} {% if pos.duration_minutes <= 52 %}hidden{% endif %}">
-                            {{ pos.agenda_item.session_duration_minutes }} min
+                    </div>
+                {% else %}
+                    <div class="timetable-session absolute rounded-md overflow-hidden cursor-pointer border text-xs opacity-85 focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-primary {% if pos.agenda_item.session_id in conflict_session_pks %} bg-danger-bg border-danger hover:opacity-100 {% elif pos.agenda_item.session_id in slot_violation_session_pks %} bg-warning-bg border-warning hover:opacity-100 {% else %} bg-info-bg border-info hover:opacity-100 {% endif %}"
```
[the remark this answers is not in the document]
fancysnake: Valid — fixed in f1dd7be2, though not by reordering the branches. The two branches are now one card: per-block state lives on `SessionPositionDTO`, conflict and slot-violation win for colour, and the dashed border plus the absent interactive attributes carry "not yours". Dimming is dropped for anything that is not `normal`, so the card that should shout is no longer the quietest one.

src/ludamus/templates/panel/parts/timetable-session-detail.html
```diff
@@ -67,26 +67,15 @@ <h3 class="font-semibold text-foreground text-sm">{% translate "Session details"
...
+            tab, where an organizer works one facilitator at a time.
+            {% endcomment %}
+            {% if agenda_item.session_confirmed %}
+                <div class="mt-2 text-xs text-success">{% icon "check-circle" class="w-3.5 h-3.5 inline" %} {% translate "Program item confirmed" %}</div>
+            {% else %}
+                <div class="mt-2 text-xs text-foreground-muted">{% translate "Program item not confirmed" %}</div>
+            {% endif %}
```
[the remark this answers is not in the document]
fancysnake: First suggestion taken; second one does not apply.

The per-item confirmation path is restored in f1dd7be2 — `TimetableConfirmView`, its URL, `SessionConfirmationService.set_session_confirmed`, DI wiring and `test_timetable_confirm.py` — and the e2e spec goes back to exercising confirm and undo instead of asserting the controls are absent. The reachability gap was real and narrower than "manual confirmation": the confirmations tab is keyed on a facilitator inside a track, so items with neither could never be confirmed there.

"Remove manual-confirmation mode and migrate existing unconfirmed items" is not on the table — `Event.auto_confirm_sessions` is a per-event opt-in setting, not a mode.

PR #798, opened 2026-08-06
Fix/timetable split sessions at midnight

fancysnake:
## Summary by CodeRabbit

* **Bug Fixes**
  * Overnight sessions now appear on every calendar day they span.
  * Session blocks are clipped at local midnight, preventing them from extending into the wrong day.
  * Timetable positioning preserves full session duration while accurately displaying only the visible portion.
  * Presenter and duration details are hidden appropriately for very short visible session blocks.
* **Tests**
  * Added coverage for cross-midnight sessions, timezone handling, day segmentation, and clipped timetable rendering.


fancysnake (commented):

The behaviour goal (night program visible on both sides of midnight) is right. The implementation gets there by clipping at three separate call sites, adding a fourth duration concept to a DTO, and re-implementing a midnight splitter that already exists in `mills/timeslots.py` — while quietly dropping the mitigation for the exact failure mode the deleted code documented.

**The one that needs an answer before merge:** the panel grid's time axis is grid-wide (`total_minutes` → `--grid-extent` in `timetable-grid.html`, one value for every day column). Splitting slots at midnight guarantees a `00:00` window on the later date and a `24:00` end on the earlier one, so `total_minutes` is now `1440` for the whole grid as soon as *any* slot crosses midnight. The branch's own test asserts this (`13 * 60` → `24 * 60`) with `day_one.columns[0].sessions == []` — a full-height, entirely empty day column. The comment deleted from `_slot_windows_by_grid_date` named this precisely: *"mint a 00:00-anchored phantom day and stretch the grid's shared time axis to a full 24 hours"*. The mitigation is gone; the replacement isn't there.

**The code-judo move**, in one line: `mills/timeslots.slot_windows` already computes *"which local-date windows does this interval cover, clipped to each date"*. Generalise it to `local_day_windows(start, end, tz)` (keep `slot_windows(slot, tz)` as the one-line adapter) and every piece of new machinery in this PR is a consumer of it — `_next_local_midnight`, `_local_day_segments`, `day_limit`, and both hand-rolled `max(start, lo) / min(end, hi)` clamps all disappear.

Details inline. Nothing here is a behaviour objection — it's all "same behaviour, much less code".

src/ludamus/mills/timetable.py
```diff
@@ -152,7 +139,7 @@ def build_grid(
         spaces = leaf_spaces[start : start + TIMETABLE_ROOM_PAGE_SIZE]
 
         all_slots = self._uow.time_slots.list_by_event(event_pk)
-        windows_by_date = _slot_windows_by_grid_date(all_slots, tz)
+        windows_by_date = slot_windows_by_local_date(all_slots, tz)
```
fancysnake, line 192: **Structural regression — the grid's shared time axis silently becomes 24h.**

`_grid_minute_bounds` folds *all* rendered dates into one `(start, end)` pair, and `timetable-grid.html` renders that as a single `--grid-extent` shared by every day column. With clamped windows, an overnight slot contributes `end = 1440` on day N and `start = 0` on day N+1, so `total_minutes == 1440` grid-wide. A three-day con with one late-night slot now draws every day as a 24-hour column, most of it empty. `test_overnight_slot_adds_the_day_it_reaches_into` locks this in (`24 * 60`, `day_one.columns[0].sessions == []`).

The deleted `_slot_windows_by_grid_date` comment called this out by name. Removing the workaround without replacing it is the blocker here.

The honest fix is that the shared axis and midnight-splitting are incompatible assumptions: move the axis onto `TimetableDayGridDTO` (each day owns its own start + extent, the time gutter renders per day) instead of one grid-wide `total_minutes`. If that's too big for this PR, say so explicitly and derive the bounds from the *unsplit* slots so the split stays a placement concern, not an axis concern — but don't leave it asserted-as-correct in a test.

**Also — removed the cause, kept the workaround.** `days_past_midnight` and its comment (`# An overnight window ends past 24:00 on its own day's clock`) existed *only* because `_slot_windows_by_grid_date` merged two windows into one running past 24:00. `slot_windows_by_local_date` clamps to `[00:00, 24:00]`, so that term is now always 0, or exactly 1440-with-zero-time. The comment is false as written. Collapse it to a plain `window_end - day_midnight` in minutes and delete the branch.

src/ludamus/gates/web/django/chronology/schedule.py
```diff
@@ -48,22 +49,42 @@ class RoomLaneDay:
     tiles: list[RoomLaneTile]
 
 
+def _next_local_midnight(moment: datetime) -> datetime:
+    return timezone.localtime(moment + timedelta(days=1)).replace(
+        hour=0, minute=0, second=0, microsecond=0
+    )
+
+
+def _local_day_segments(start: datetime, end: datetime) -> list[datetime]:
```
fancysnake: **This is a near-duplicate of the canonical helper, in the wrong layer.**

The PR deletes a midnight splitter from `mills/` and adds a second, differently-shaped one to a `gates/` presentation module. `mills/timeslots.slot_windows` already answers exactly this question — "which local-date windows does this interval cover" — and answers it better: explicit `datetime.combine(date + offset, time.min, tzinfo=tz)` date arithmetic, rather than leaning on wall-clock `+ timedelta(days=1)` followed by `.replace(hour=0, …)`. That form happens to be correct (aware-datetime arithmetic is wall-clock), but a reader has to prove that to themselves; the existing one requires no proof.

Proposed judo:

```python
# mills/timeslots.py
def local_day_windows(start: datetime, end: datetime, tz: tzinfo) -> list[SlotWindow]:
    ...  # today's slot_windows body, on raw datetimes

def slot_windows(slot: TimeSlotDTO, tz: tzinfo) -> list[SlotWindow]:
    return local_day_windows(slot.start_time, slot.end_time, tz)
```

Then `build_schedule_days` is `for window_start, _ in local_day_windows(item.start_time, item.end_time, tz):` and both `_next_local_midnight` and `_local_day_segments` are deleted. One definition of "a day", used by mills and gates alike.

src/ludamus/gates/web/django/chronology/schedule.py
```diff
@@ -144,8 +171,10 @@ def build_room_lanes(schedule_days: list[ScheduleDay]) -> list[RoomLaneDay]:
                     data.loc["parent_slug"],
                     data.loc["parent_name"],
                 )
-                start_hour = int((item.start_time - day_start).total_seconds() // 3600)
-                end_offset = (item.end_time - day_start).total_seconds() / 3600
+                visible_start = max(item.start_time, day_start)
```
fancysnake: **Third copy of the same clamp — this signals a missing model, not a missing helper.**

`max(item.start_time, lo)` / `min(item.end_time, hi)` now appears here, again two lines up in `day_end = min(day_limit, max(...))`, and a third time in `mills/timetable._position_sessions`. Three independent hand-rolls of "clip this session to this day", each of which has to get the boundary condition right on its own.

If `local_day_windows` (see the comment on `_local_day_segments`) returns the clipped `(start, end)` per local date — which the existing `slot_windows` already does — then both call sites *read* the window instead of recomputing the clamp, `day_limit` stops existing in `build_room_lanes`, and `_position_sessions` no longer needs `grid_end` threaded in at all. Clipping becomes a property of the day segment rather than something every consumer re-derives.

src/ludamus/pacts/chronology.py
```diff
@@ -382,6 +382,9 @@ class SessionPositionDTO(BaseModel):
     agenda_item: AgendaItemDTO
     start_minutes: int
     duration_minutes: int
+    # What this day's column shows: the same as duration_minutes, except for a
+    # session clipped by midnight, which also renders on the next day.
+    visible_minutes: int
```
fancysnake: **Adding a third duration field to this DTO isn't necessary — there's a deletion available instead.**

After this change, a template author picking a duration has to choose between `pos.duration_minutes`, `pos.visible_minutes`, and `pos.agenda_item.session_duration_minutes` — and the comment here ("the same as `duration_minutes`, except…") is an admission that two of the three are the same value almost always. That's exactly the kind of near-duplicate field that gets picked wrong six months from now, silently, in a way no test catches.

But `AgendaItemDTO.session_duration_minutes` *is* the real length (`links/db/django/agenda_item.py` computes it as `end - start`), and `timetable-day-grid.html` already renders it on the very next line. So:

- make `duration_minutes` mean the one thing the position DTO is for — how tall is this tile — i.e. the clipped value;
- point `data-duration` (the drag payload, `timetable.ts`) at `pos.agenda_item.session_duration_minutes`, which is the real length by construction.

Net effect: the PR *removes* a duration concept instead of adding one, the template diff collapses to a single line, and there is no longer a pair of fields a reader has to disambiguate.

src/ludamus/gates/web/django/chronology/schedule.py
```diff
@@ -48,22 +49,42 @@ class RoomLaneDay:
...
-            hours.append(ScheduleHour(start=hour_start, sessions=[]))
-        hours[-1].sessions.append(data)
+        item = data.agenda_item
+        for segment_start in _local_day_segments(item.start_time, item.end_time):
+            hour_start = segment_start.replace(minute=0, second=0, microsecond=0)
+            by_hour[hour_start].append((item.start_time, data))
+
+    days: list[ScheduleDay] = []
+    for hour_start in sorted(by_hour):
+        sessions = [data for _, data in sorted(by_hour[hour_start], key=itemgetter(0))]
```
fancysnake: **The sort key smuggled through a tuple is working around `Optional`, and it costs a re-sort per hour.**

`dict[datetime, list[tuple[datetime, SessionData]]]` + `itemgetter(0)` + a sort inside the output loop exists only because `data.agenda_item` is `SessionData | None` and the narrowing doesn't survive into a key lambda. Sort once, up front, over `(item, data)` pairs and the whole apparatus goes away:

```python
scheduled = sorted(
    ((d.agenda_item, d) for d in sessions_data.values() if d.agenda_item is not None),
    key=lambda pair: pair[0].start_time,
)
by_hour: dict[datetime, list[SessionData]] = defaultdict(list)
for item, data in scheduled:
    for window_start, _ in local_day_windows(item.start_time, item.end_time, tz):
        by_hour[window_start.replace(minute=0, second=0, microsecond=0)].append(data)
```

Appends land in start order, so the per-hour sort, the tuple, and the `operator` import all disappear.

Two smaller things while you're in here:

- `build_room_lanes` still re-checks `data.agenda_item is not None` twice, even though `build_schedule_days` only ever emits non-`None` items. Those branches are unreachable and unprovable to a reader. If the pairs above were carried through (or `ScheduleHour` held items rather than `SessionData`), the type would state the invariant and both checks would go.
- `ScheduleDay.first_start` changed meaning here — it used to be the first session's UTC `start_time`, it's now a local truncated hour. Same name, different contract. `day_start` would say what it is.

src/ludamus/gates/web/django/chronology/schedule.py
```diff
@@ -116,11 +137,17 @@ def build_room_lanes(schedule_days: list[ScheduleDay]) -> list[RoomLaneDay]:
...
 
         day_start = day.hours[0].start
-        day_end = max(
-            data.agenda_item.end_time
-            for hour in day.hours
-            for data in hour.sessions
-            if data.agenda_item is not None
+        # The day's lanes stop at midnight; the rest of a night session is
+        # drawn again on the day it runs into.
+        day_limit = _next_local_midnight(day_start)
```
fancysnake: **Same phantom-day effect as the panel grid, on the public schedule — please confirm this is intended.**

Once a night session spills into day N+1, that day's first `ScheduleHour` is `00:00`, so `day_start` is midnight and `hour_marks` is generated from `range(hour_count)` starting there. A day whose own program starts at 09:00 now renders nine empty hour rows above the first real session, and the rail in `_compact_schedule.html` gains nine dead scrubber entries.

That's the unavoidable cost of "the tile must appear on both days" given the current model — but it's a visible change to every day that follows a late night, and nothing in the PR description or tests acknowledges it. Either own it explicitly, or let the marks start at the earlier of (the spilled tile's end, the day's first own session) so the gap collapses to just the hours the night tile actually occupies.

src/ludamus/mills/timetable.py
fancysnake: Fixed in f12d30ae2 — by taking the structural option rather than documenting the cost.

`TimetableDayGridDTO` now owns `total_minutes` and `time_labels`; `TimetableGridDTO` lost both. `_grid_minute_bounds` is gone, replaced by `_day_range`, which reads one day’\s windows only. The calendar’\s track list alternates gutter, rooms, gutter, rooms — unrolled in the template because CSS `repeat()` cannot nest — so each day renders its own axis next to its own columns.

On the example that motivated this: a Jan 1 12:00→Jan 2 01:00 slot plus a Jan 2 12:00→22:00 slot used to give every day a 24h axis. Jan 1 is now 12 hours and Jan 2 is 22 hours, each spanning only its own program.

You were also right about `days_past_midnight`: with windows clamped to `[00:00, 24:00]` it was never *past* midnight, only ever exactly at it, so the comment was false. `_day_range` is a plain `window_end - midnight` in minutes and the branch is gone.

One correction on the supporting evidence: `day_one.columns[0].sessions == []` in `test_overnight_slot_adds_the_day_it_reaches_into` was not locking the bug in — that test's agenda item is `2026-01-02 00:00 → 01:00`, entirely on Jan 2, so an empty Jan 1 column was the right answer there. Only the `24 * 60` assertion was wrong, and it now reads `[12 * 60, 22 * 60]`.

src/ludamus/gates/web/django/chronology/schedule.py
fancysnake: Fixed in f12d30ae2, as proposed. `slot_windows` is now a one-line delegate over `local_day_windows(start, end, tz)` in `mills/timeslots.py`; `build_schedule_days` calls it directly; `_next_local_midnight` and `_local_day_segments` are deleted. One definition of a day, used by mills and gates alike.

For the record I checked the parenthetical: `+ timedelta(days=1)` on a tz-aware datetime is wall-clock arithmetic and does advance exactly one calendar day across both Europe/Warsaw transitions, including the 25-hour fall-back day. It was correct — it just needed a reader to prove it, which is the point.

src/ludamus/gates/web/django/chronology/schedule.py
fancysnake: Fixed in f12d30ae2 — the missing model is now `ScheduleTile(data, start, end)`: one session as it appears on one local date, already clipped to it. `build_schedule_days` produces them from `local_day_windows`, and `build_room_lanes` reads `tile.start` / `tile.end` instead of recomputing the clamp. `day_limit` is gone, and `_position_sessions` no longer needs `grid_end` threaded in for clipping (it still takes the day range, but as the day’\s own bounds rather than a clipping cap borrowed from elsewhere).

That leaves one clamp instead of three, and it lives in `local_day_windows`.

src/ludamus/pacts/chronology.py
fancysnake: Fixed in f12d30ae2 — taken as the deletion. `SessionPositionDTO.duration_minutes` is now the clipped value (how tall the tile is here), `visible_minutes` is gone, and `data-duration` points at `pos.agenda_item.session_duration_minutes`, which `links/db/django/agenda_item.py:_to_dto` computes as `end - start`. The template diff is a single line and there is no longer a pair of near-identical duration fields to disambiguate.

src/ludamus/gates/web/django/chronology/schedule.py
fancysnake: Fixed in f12d30ae2, all three.

The tuple and the per-hour sort are gone: `build_schedule_days` sorts `(item, data)` pairs once up front, appends land in start order, and the `operator` import this PR added is removed.

The unreachable `agenda_item is not None` checks went with the `ScheduleTile` change — `build_room_lanes` iterates `day.tiles`, whose type states the invariant, so there is nothing left to re-check.

`ScheduleDay.first_start` is now `day_start`, in the dataclass, in `_schedule_day_header.html`, in the scrubber rail, and in the tests. You were right that it had quietly changed contract.

src/ludamus/gates/web/django/chronology/schedule.py
fancysnake: Checked this one rather than fixed it — I think the effect is smaller than described, and the part that remains is pre-existing behaviour.

**The scrubber claim does not hold.** The rail in `_compact_schedule.html` iterates `day.hours`, and `day.hours` only contains hours that actually have a session in them (`by_hour` is built from tiles, one entry per tile). A night session spilling into day N+1 adds exactly one rail entry — `00`, which is real program — not nine dead ones.

**The hour rows are real but not new in kind.** `hour_marks` in `_room_lanes.html` runs `range(hour_count)` from `day_start` to `day_end`, so a day whose spilled tile ends at 02:00 and whose own program starts at 09:00 does render ~7 empty gridlines. But that is the same thing the grid already does for any mid-day gap: a day with sessions at 09:00 and 20:00 renders 11 hour rows today, on `main`, with no night session involved. The night spill makes the day *start earlier*, which is correct — the 00:00–02:00 tile is genuine content on that date.

Collapsing the gap would mean skipping hours in the row mapping, which changes the existing gap behaviour for every day, not just the ones after a late night. That reads like a separate product decision rather than something this PR should smuggle in.

Leaving this open rather than resolving it — happy to take the skip-empty-hours change if you want it, but I did not want to decide it here.

PR #801, opened 2026-08-07
Deflake the iOS regression specs: anchors, polling, scoped snapshots

hasparus:
Split out of #789 — the `ios-regressions` job&#39;s flakiness had nothing to do with the participants limit, and fixing it turned into its own investigation.

## What was wrong

Three independent problems, each fixed at its root:

1. **Scroll gestures outran the runner.** Reaching mid-schedule cost up to 14 gestures, and one gesture on the 110-session seed page can exceed the runner&#39;s 180 s per-command ceiling. The rail&#39;s own `#slot-YYYYMMDD-HH` anchors land Safari there on load instead; the slot id is read from the served markup.

2. **Fixed waits raced the UI.** A snapshot taken before Safari applies the fragment scroll, or before a callout has had time to appear, reads as success. Every async check now polls to a deadline through one harness `pollUntil`; &#34;nothing surfaced&#34; is only concluded once the window has elapsed, and each window ends on a probe, not a sleep.

3. **The snapshot silently truncates at 300 nodes.** `fastSnapshotLimit` in the runner&#39;s `RunnerTests+Snapshot.swift` caps the walk and reports the rest only as `truncated: true`, which nothing read. The rail nav is the *last* element in the document, so it never appeared in any snapshot — every failed rail hunt across seven device runs, including the &#34;aria-labels never surface&#34; conclusion, was this cap, not label semantics. Snapshots are now **scoped** to the rail nav&#39;s own `aria-label` — the scope resolves via a live element query with no node cap, and only the rail subtree is walked. The callout check gets the same treatment (scoped to &#34;Open in&#34;), so the sheet can never be truncated away into a vacuous pass either.

The modal spec shares the harness changes: accessible names come from the served, language-pinned markup (Bun&#39;s `fetch` sends no `Accept-Language`, so Django would otherwise serve Polish while the simulator renders English), and visibility checks share one conservative screen-band convention instead of trusting `hittable`, which reads false inside Safari web content.

## Validation

Four consecutive on-device green runs (327ee7a, 32faeb7, 3195cbe, 35055d7). Mutation-tested in #803: with `[-webkit-touch-callout:none]`, the `contextmenu` preventDefault, and the modal description stripped, the scrubber failed with `surfacedCalloutSignals = ["Open in New Tab"]` and the modal failed its content check — on both retry attempts each — while the untouched behaviors kept passing. The specs fail on exactly the regressions they guard, and only on those.

https://claude.ai/code/session_01AT39sjmMjfTKLVwLXmG6pB

### DIFF
--- a/scripts/ios-regressions/harness.ts
+++ b/scripts/ios-regressions/harness.ts
@@ -33,22 +33,23 @@ const FALLBACK_VIEWPORT: Rect = { x: 0, y: 0, width: 402, height: 874 };
 
 // The one poll-to-deadline loop: probe until it yields a value or the window
 // closes, and only conclude "nothing" (null) once the window has actually
-// elapsed — which is what specs asserting absence rely on. Paced by a local
-// sleep rather than the runner's wait command, because the same loop must
-// serve before a device session exists (fetchReadyPage) as during one; the
-// probes themselves are the only traffic the runner needs to see. A throwing
-// probe aborts the poll.
+// elapsed — which is what specs asserting absence rely on. The window ends on
+// a probe, not a sleep, so its final interval is observed rather than slept
+// away. Paced by a local sleep rather than the runner's wait command, because
+// the same loop must serve before a device session exists (fetchReadyPage) as
+// during one; the probes themselves are the only traffic the runner needs to
+// see. A throwing probe aborts the poll.
 export const pollUntil = async <T>(
   probe: () => Promise<T | null>,
   { timeoutMs, intervalMs = 500 }: { timeoutMs: number; intervalMs?: number },
 ): Promise<T | null> => {
   const deadline = Date.now() + timeoutMs;
-  do {
+  for (;;) {
     const result = await probe();
     if (result !== null) return result;
+    if (Date.now() >= deadline) return null;
     await new Promise((resolve) => setTimeout(resolve, intervalMs));
-  } while (Date.now() < deadline);
-  return null;
+  }
 };
 
 const deviceName = env.IOS_DEVICE_NAME ?? "iPhone 17 Pro";
--- a/scripts/ios-regressions/harness.ts
+++ b/scripts/ios-regressions/harness.ts
@@ -25,10 +25,31 @@ export const hookTimeoutMs = Number(env.IOS_HOOK_TIMEOUT_MS ?? "300000");
 
 export type Rect = { x: number; y: number; width: number; height: number };
 
-// Only reached when a snapshot comes back with no nodes to read a rect from. Sized for the iPhone 17 Pro the workflow asks for first;
-// when the image lacks that device the workflow picks another, so treat this as
-// a shape that keeps arithmetic finite, not as this run's screen.
+// Only reached when a snapshot comes back with no nodes to read a rect from.
+// Sized for the iPhone 17 Pro the workflow asks for first; when the image
+// lacks that device the workflow picks another, so treat this as a shape that
+// keeps arithmetic finite, not as this run's screen.
 const FALLBACK_VIEWPORT: Rect = { x: 0, y: 0, width: 402, height: 874 };
+
+// The one poll-to-deadline loop: probe until it yields a value or the window
+// closes, and only conclude "nothing" (null) once the window has actually
+// elapsed — which is what specs asserting absence rely on. Paced by a local
+// sleep rather than the runner's wait command, because the same loop must
+// serve before a device session exists (fetchReadyPage) as during one; the
+// probes themselves are the only traffic the runner needs to see. A throwing
+// probe aborts the poll.
+export const pollUntil = async <T>(
+  probe: () => Promise<T | null>,
+  { timeoutMs, intervalMs = 500 }: { timeoutMs: number; intervalMs?: number },
+): Promise<T | null> => {
+  const deadline = Date.now() + timeoutMs;
+  do {
+    const result = await probe();
+    if (result !== null) return result;
+    await new Promise((resolve) => setTimeout(resolve, intervalMs));
+  } while (Date.now() < deadline);
+  return null;
+};
 
 const deviceName = env.IOS_DEVICE_NAME ?? "iPhone 17 Pro";
 const runtime = env.IOS_RUNTIME;
@@ -215,7 +236,6 @@ export const createIosHarness = (session: string): IosHarness => {
   // with gestures, which are the slowest thing either spec can do.
   const fetchReadyPage = async (url: URL, contains: string): Promise<string> => {
     console.log(`Checking local event page at ${url.toString()}...`);
-    const deadline = Date.now() + 60000;
     let lastError = "no response";
 
     // Pin the language both halves of a spec see. Bun's fetch sends no
@@ -227,24 +247,34 @@ export const createIosHarness = (session: string): IosHarness => {
     // the spec fails loudly rather than passing vacuously.
     const headers = { "Accept-Language": "en" };
 
-    while (Date.now() < deadline) {
-      try {
-        const response = await fetch(url, { headers });
-        const text = await response.text();
-        if (response.ok && text.includes(contains)) return text;
+    const unusable = (): Error =>
+      new Error(
+        `Local event page is not usable at ${url.toString()} (${lastError}). ` +
+          "Make sure the e2e server is running and serving the seeded event.",
+      );
 
-        lastError = `HTTP ${response.status}; body starts with ${JSON.stringify(text.slice(0, 160))}`;
-        if (response.status >= 400 && response.status < 500) break;
-      } catch (error) {
-        lastError = error instanceof Error ? error.message : String(error);
-      }
-      await new Promise((resolve) => setTimeout(resolve, 1000));
-    }
+    const page = await pollUntil<string>(
+      async () => {
+        let status: number | null = null;
+        try {
+          const response = await fetch(url, { headers });
+          const text = await response.text();
+          if (response.ok && text.includes(contains)) return text;
 
-    throw new Error(
-      `Local event page is not usable at ${url.toString()} (${lastError}). ` +
-        "Make sure the e2e server is running and serving the seeded event.",
+          status = response.status;
+          lastError = `HTTP ${status}; body starts with ${JSON.stringify(text.slice(0, 160))}`;
+        } catch (error) {
+          lastError = error instanceof Error ? error.message : String(error);
+        }
+        // A 4xx is a wrong URL or a missing seed, not a server still starting;
+        // no amount of polling fixes it.
+        if (status !== null && status >= 400 && status < 500) throw unusable();
+        return null;
+      },
+      { timeoutMs: 60000, intervalMs: 1000 },
     );
+    if (page === null) throw unusable();
+    return page;
   };
 
   return {
--- a/scripts/ios-regressions/harness.ts
+++ b/scripts/ios-regressions/harness.ts
@@ -25,28 +25,51 @@ export const hookTimeoutMs = Number(env.IOS_HOOK_TIMEOUT_MS ?? "300000");
 
 export type Rect = { x: number; y: number; width: number; height: number };
 
-// Only reached when a snapshot comes back without a root rect. Sized for the
-// iPhone 17 Pro the macOS runner image actually provides -- CI ignores
-// `deviceName` below, because the workflow picks a UDID and passes it in.
+// Only reached when a snapshot comes back with no nodes to read a rect from. Sized for the iPhone 17 Pro the workflow asks for first;
+// when the image lacks that device the workflow picks another, so treat this as
+// a shape that keeps arithmetic finite, not as this run's screen.
 const FALLBACK_VIEWPORT: Rect = { x: 0, y: 0, width: 402, height: 874 };
 
 const deviceName = env.IOS_DEVICE_NAME ?? "iPhone 17 Pro";
 const runtime = env.IOS_RUNTIME;
 const providedUdid = env.UDID;
 
+// One convention for "is this on screen": rect centre inside the viewport,
+// clear of a band for Safari's top and bottom chrome. The band's exact size is
+// unknowable from this sandbox; being conservative only narrows what counts as
+// visible, which specs must treat as "press/check something else", never as a
+// pass. The runner's `hittable` is NOT this check — it reads false inside
+// Safari's web content (device-falsified on this branch).
+const CHROME_INSET = 120;
+
+export const centreOnScreen = (rect: Rect, viewport: Rect): boolean => {
+  const centreY = rect.y + rect.height / 2;
+  return (
+    centreY >= viewport.y + CHROME_INSET && centreY <= viewport.y + viewport.height - CHROME_INSET
+  );
+};
+
+export const describeNode = (node: SnapshotNode): string => {
+  const rect = node.rect
+    ? ` x=${Math.round(node.rect.x)} y=${Math.round(node.rect.y)} w=${Math.round(node.rect.width)} h=${Math.round(node.rect.height)}`
+    : "";
+  return `${node.type ?? "node"} ref=@${node.ref}${rect} hittable=${String(node.hittable)} label=${JSON.stringify(
+    node.label ?? node.value ?? "",
+  )}`;
+};
+
 export type IosHarness = {
   client: AgentDeviceClient;
   deviceOptions: IosDeviceOptions;
-  takeSnapshot: () => Promise<CaptureSnapshotResult>;
+  takeSnapshot: (scope?: string) => Promise<CaptureSnapshotResult>;
   viewportOf: (snapshot: CaptureSnapshotResult) => Rect;
-  viewportRect: () => Promise<Rect>;
   close: () => Promise<void>;
-  snapshotLabels: () => Promise<string[]>;
+  snapshotLabels: (scope?: string) => Promise<string[]>;
   findNodeByLabel: (label: string) => Promise<SnapshotNode | null>;
   wait: (durationMs: number) => Promise<void>;
   openUrl: (url: string, udid: string) => Promise<void>;
   prepareDevice: () => Promise<string>;
-  assertPageReady: (url: URL, contains: string) => Promise<void>;
+  fetchReadyPage: (url: URL, contains: string) => Promise<string>;
 };
 
 export const createIosHarness = (session: string): IosHarness => {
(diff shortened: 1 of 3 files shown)


scripts/ios-regressions/scrubber-longpress.test.ts
```diff
@@ -21,55 +32,201 @@ const {
...
+      lastSnapshotError = error;
+      console.warn("Snapshot failed while the page was settling; retrying.", error);
+    }
+    if (screen && scoped) {
+      const markers = railMarkersFrom(scoped, markerNames, screen);
+      if (markers.length > 0) return markers;
+    }
+    await wait(500);
+  } while (Date.now() < deadline);
+
```
hasparus, line 154: **Blocking.** This PR makes "poll to a deadline" the house mechanism for every async check — and then hand-rolls the same skeleton at each call site. Counting them: `waitForRailMarkers` here, `surfacedCallout` below, `waitUntilShowing` in modal.test.ts, plus the pre-existing `waitForLabel` and `fetchReadyPage` loops. Five copies of `deadline = Date.now() + timeoutMs; do { probe; wait(interval) } while (...)`, each with its own interval (400/500/500/1000ms) and its own deadline style.

The polling machinery belongs in the harness, next to `wait`, as one canonical helper:

```ts
const pollUntil = async <T>(
  probe: () => Promise<T | null>,
  { timeoutMs, intervalMs = 500 }: { timeoutMs: number; intervalMs?: number },
): Promise<T | null> => {
  const deadline = Date.now() + timeoutMs;
  do {
    const result = await probe();
    if (result !== null) return result;
    await wait(intervalMs);
  } while (Date.now() < deadline);
  return null;
};
```

Then `waitForLabel` is `pollUntil(() => findNodeByLabel(label), { timeoutMs })`, `waitUntilShowing` and `surfacedCallout` become one-liner probes, and this function keeps only what is genuinely its own: the screen-rect caching, the error retention, and the diagnostic throw — with closure state, all of which survive the extraction unchanged:

```ts
const markers = await harness.pollUntil(async () => {
  try {
    screen ??= viewportOf(await takeSnapshot());
    const scoped = (lastScoped = await takeSnapshot(navName));
    const found = railMarkersFrom(scoped, markerNames, screen);
    return found.length > 0 ? found : null;
  } catch (error) {
    lastSnapshotError = error;
    return null;
  }
}, { timeoutMs });
if (!markers) throwRailDiagnostics(...);
```

The loop bodies are where the hard-won device knowledge lives; the loop scaffolding should not be re-derived (with drifting intervals and orderings) every time a new check is added. None of the deadline semantics change, so this is safe to do without a device run.

---

scripts/ios-regressions/modal.test.ts
```diff
@@ -22,12 +29,33 @@ const {
...
+// mid-animation reports content that is about to be there as missing.
+const waitUntilShowing = async (texts: string[], timeoutMs: number): Promise<boolean> => {
+  const deadline = Date.now() + timeoutMs;
+  do {
+    const labels = await showingLabels();
+    if (texts.every((text) => labels.some((label) => label.includes(text)))) return true;
+    await client.command.wait({ ...deviceOptions, durationMs: 400 });
+  } while (Date.now() < deadline);
+  return false;
 };
```
hasparus: Same finding as the `waitForRailMarkers` comment: this is the third hand-rolled deadline loop the PR adds. With a harness `pollUntil` this collapses to a probe:

```ts
const waitUntilShowing = async (texts: string[], timeoutMs: number): Promise<boolean> =>
  (await pollUntil(
    async () => {
      const labels = await showingLabels();
      return texts.every((t) => labels.some((l) => l.includes(t))) || null;
    },
    { timeoutMs, intervalMs: 400 },
  )) === true;
```

Also: this loop reaches for `client.command.wait({ ...deviceOptions, durationMs: 400 })` inline while the harness already exports `wait` for exactly this — the scrubber spec uses it, this file bypasses it. A harness-level `pollUntil` erases that inconsistency for free, since the helper owns the waiting.

---

scripts/ios-regressions/scrubber-longpress.test.ts
```diff
@@ -21,55 +32,201 @@ const {
...
+// is the last element in the document — an unscoped snapshot truncates before
+// ever reaching it (seven device runs of "no rail names appeared" traced back
+// to exactly this). Scoping the snapshot to the nav's accessible name walks
+// only its subtree — a few dozen nodes — so every rendered marker surfaces.
+// Both marker name sets come from the served, already-translated markup: the
+// links' bare hour text ("10") and their aria-labels, whichever the engine
+// reports.
+const RAIL_NAV_NAME = /<nav class="schedule-rail[^"]*"\s+aria-label="([^"]+)"/;
+const RAIL_HOUR_NAMES = /class="schedule-rail-hour[^"]*"[^>]*aria-label="([^"]+)"/g;
+const RAIL_HOUR_TEXTS = /class="schedule-rail-hour[^"]*"[^>]*>([^<]+)<\/a>/g;
```
hasparus, line 91: Non-blocking, recorded so nobody churns on it later: I checked whether Bun's built-in `HTMLRewriter` could replace this regex stack and delete `decodeEntities`. It cannot delete it — `el.getAttribute("aria-label")` returns the raw source value with entities **not** decoded (verified locally: `aria-label="Jump &amp; go"` comes back as `"Jump &amp; go"`), so the decoder stays either way. The rewriter would only buy attribute-order independence.

Which leaves the one real fragility here: all three patterns are coupled to attribute order in `_compact_schedule.html` — `RAIL_NAV_NAME` requires `aria-label` to immediately follow `class` on the `<nav>`, and the hour patterns require `class` first in the `<a>`. A template attribute reorder breaks the spec. Since `railNavName` and the `markerNames.size === 0` guard both throw loudly before any device time is spent, this is acceptable — but consider a short comment on these regexes naming the template file, so whoever reorders attributes there finds this coupling by grep instead of by a red CI run.

---

scripts/ios-regressions/harness.ts
```diff
@@ -25,28 +25,51 @@ export const hookTimeoutMs = Number(env.IOS_HOOK_TIMEOUT_MS ?? "300000");
 
 export type Rect = { x: number; y: number; width: number; height: number };
 
-// Only reached when a snapshot comes back without a root rect. Sized for the
-// iPhone 17 Pro the macOS runner image actually provides -- CI ignores
-// `deviceName` below, because the workflow picks a UDID and passes it in.
+// Only reached when a snapshot comes back with no nodes to read a rect from. Sized for the iPhone 17 Pro the workflow asks for first;
+// when the image lacks that device the workflow picks another, so treat this as
+// a shape that keeps arithmetic finite, not as this run's screen.
 const FALLBACK_VIEWPORT: Rect = { x: 0, y: 0, width: 402, height: 874 };
```
hasparus: Nit, non-blocking: the first line of this comment was not rewrapped after the edit — it runs well past the wrap width of every other comment block in the file, and the sentence break lands mid-line. Rewrap while touching it.

---

hasparus (commented):
Deep code-quality audit of `327ee7a` against `main`. Verdict: **changes requested** — one structural finding; everything else in this PR is genuinely good work. (GitHub refuses a formal REQUEST_CHANGES from the PR author's own account, so this lands as a comment review; treat it as request-changes.)

## What holds up (verified, not assumed)

The three root-cause fixes are the right shape, and I confirmed the claims they rest on rather than taking the comments on faith:

- **Anchor navigation instead of scroll gestures** — `/event/<slug>/` is the real route (`src/ludamus/adapters/web/django/urls.py:26`, mounted at root), so the `EVENT_PATH` default fix from `/chronology/event/…` is correct, and the workflow sets no `EVENT_PATH`, so the default is load-bearing. Reading the slot id and rail names from the served markup instead of hardcoding them is the right dependency direction: the spec follows the seed instead of chasing it.
- **Scoped snapshots** — `fastSnapshotLimit` and the truncation behavior are real (`RunnerTests+Snapshot.swift`), `scope` is a typed option on `capture.snapshot`, and `findScopeElement` is `label CONTAINS[c] OR identifier CONTAINS[c]`, `firstMatch`, nil on miss. I also checked for scope-string collisions on the public event page: nothing there contains "Open in" (the only hit, "Open in Schedule", is panel-only) and the hour links' "Jump to Saturday 10:00" labels do not contain "Jump to time", so `firstMatch` cannot be stolen from the nav or the callout sheet on this page.
- **Anti-vacuous-pass structure** — for a spec whose pass condition is absence, the polling-to-deadline + scoped-callout + loud-diagnostic-throw combination is exactly right, and the failure paths (`railNavName` throw, empty `markerNames` throw, the scoped-snapshot dump with `truncated` and node sample) all fail before burning device time.
- **Consolidation** — `describeNode` and the single `centreOnScreen` visibility convention moving into the harness, and the dead `viewportRect` being deleted, are clean net reductions. The `Accept-Language: en` pin with its fail-loud rationale is well reasoned.

`tsc -p scripts/ios-regressions/tsconfig.json` and `oxlint` both pass on this branch's file contents (run against copies; the working tree here is checked out on a different branch, so I could not run them in place — noted for transparency).

## The blocking finding

The PR's own thesis is "every async check now polls to a deadline" — and then the polling scaffold is copy-pasted at every site instead of becoming a harness primitive. Three new hand-rolled deadline loops (`waitForRailMarkers`, `surfacedCallout`, `waitUntilShowing`) join two pre-existing ones (`waitForLabel`, `fetchReadyPage`), each with its own interval and deadline idiom, and one of them re-spells `client.command.wait` inline past the harness `wait` helper. Details and a concrete `pollUntil` extraction are in the inline comments; the loop *bodies* carry the device-verified knowledge and survive unchanged, so this refactor needs no simulator run to validate. This is the mechanism the next spec will reach for — it should exist once, in the harness, before a sixth copy appears.

## Non-blocking

- The rail regexes are coupled to attribute order in `_compact_schedule.html`; failure is loud, so acceptable, but a pointer comment would help (inline note). I also verified that Bun's `HTMLRewriter` would **not** delete `decodeEntities` — `getAttribute` returns entities undecoded — so do not spend a cycle on that swap.
- One stale comment wrap in `harness.ts` (inline nit).

---

scripts/ios-regressions/scrubber-longpress.test.ts
hasparus: Done in 32faeb7 — all five loops now go through one `pollUntil` in the harness, probe bodies unchanged.

One deviation from the sketch: `pollUntil` paces with a local sleep rather than the harness `wait` (the runner's wait command), because `fetchReadyPage` polls before any device session exists — device-paced polling would have left it as a sixth idiom. The probes themselves are the only traffic the runner needs to see. `wait` stays for genuine settle/pacing waits, and modal.test.ts's inline `client.command.wait` calls now go through it as well.

---

scripts/ios-regressions/harness.ts
```diff
@@ -25,28 +25,72 @@ export const hookTimeoutMs = Number(env.IOS_HOOK_TIMEOUT_MS ?? "300000");
...
+  probe: () => Promise<T | null>,
+  { timeoutMs, intervalMs = 500 }: { timeoutMs: number; intervalMs?: number },
+): Promise<T | null> => {
+  const deadline = Date.now() + timeoutMs;
+  do {
+    const result = await probe();
+    if (result !== null) return result;
+    await new Promise((resolve) => setTimeout(resolve, intervalMs));
+  } while (Date.now() < deadline);
+  return null;
```
hasparus: Non-blocking. The timeout path of this loop ends on a sleep, not a probe: after the final probe it sleeps a full `intervalMs`, then the `while` sees the deadline passed and gives up. So the last interval of every window is dead time nothing observes. Flipping the order keeps the at-least-one-probe guarantee and turns that dead sleep into one more look:

```ts
while (true) {
  const result = await probe();
  if (result !== null) return result;
  if (Date.now() >= deadline) return null;
  await new Promise((resolve) => setTimeout(resolve, intervalMs));
}
```

Where it shows: `surfacedCallout(2500)` times out four times on the scrubber's expected pass path, so every green run spends its final ~500ms x 4 asleep after the last snapshot — and a callout surfacing during that final interval is never seen. The "conclude nothing only once the window elapsed" contract holds either way, since null is still only returned at or after the deadline.

---

scripts/ios-regressions/scrubber-longpress.test.ts
```diff
@@ -21,55 +33,205 @@ const {
...
+const waitForRailMarkers = async (
+  timeoutMs: number,
+  navName: string,
+  markerNames: Set<string>,
+): Promise<RailHour[]> => {
+  let screen: Rect | null = null;
+  let lastSnapshotError: unknown = null;
+  // A slot, not a plain `let`: the assignment happens inside the probe
+  // closure, which TypeScript's flow analysis cannot see from out here.
+  const last: { scoped: CaptureSnapshotResult | null } = { scoped: null };
```
hasparus: I checked this claim before accepting the shape: with a plain `let`, tsc 5.9 strict pins the variable to its `null` initializer (assignments inside the probe closure are invisible to flow analysis), so the later `if (!scoped)` guard narrows to `never` and `scoped.nodes` fails to compile. The slot is a real workaround, not churn.

That said, there is a shape with no slot and no flow-analysis lore to explain: forget the last poll's snapshot and take a fresh scoped one on the failure path —

```ts
if (markers) return markers;

let scoped: CaptureSnapshotResult;
try {
  scoped = await takeSnapshot(navName);
} catch {
  throw new Error(`No snapshot succeeded in ${timeoutMs}ms; last error: ${String(lastSnapshotError)}`);
}
// ...dump as today
```

One extra runner roundtrip, but only on a path that already burned the full 90s, and the dump then describes the tree at failure time rather than one interval earlier. Non-blocking — take it or leave it.

---

hasparus (commented):
Round two, on 32faeb7. **Verdict: approve** (submitted as a comment because GitHub refuses a formal approval from the PR author's account). The round-one blocking finding is genuinely fixed, not cosmetically:

- One `pollUntil` in the harness now backs all five former hand-rolled deadline loops — `waitForRailMarkers`, `surfacedCallout`, `waitUntilShowing`, `waitForLabel`, `fetchReadyPage`. I diffed the probe bodies against the previous head: the device-verified knowledge (scoped snapshots, screen-rect caching from the unscoped root, error retention, the 4xx fail-fast, the loud diagnostic throws) survives unchanged; only the scaffolding moved. `waitForLabel` collapsing to a one-liner is the extraction earning its keep.
- The deviation from my sketch — pacing with a local sleep instead of the runner's `wait` — is the right call, and better than the sketch. `fetchReadyPage` polls before any device session exists, so a runner-paced `pollUntil` could never have served it; it also lets `pollUntil` live at module level with zero client dependency instead of inside the harness factory. The probes are runner traffic anyway, so nothing is lost. Accepted.
- The absence-assertion contract the scrubber depends on (only conclude "nothing" once the window has actually elapsed) is preserved in the primitive and documented on it.
- Both round-one nits applied: the rail regexes now name `templates/chronology/_compact_schedule.html`, and the FALLBACK_VIEWPORT comment is rewrapped.
- Modal's inline `client.command.wait` calls are gone; everything paces through the harness `wait`.
- `tsc -p scripts/ios-regressions` and `oxlint` are clean on 32faeb7.

Two non-blocking line notes: the dead trailing sleep on `pollUntil`'s timeout path (it costs the scrubber's green path ~2s and one unobserved interval per press), and a slotless alternative for the `last.scoped` workaround in `waitForRailMarkers` — whose necessity I verified against tsc before accepting it (a plain `let` really does narrow to `never` there).

One optional cleanup for a later pass, outside this diff: `closeDeviceSessionIfPresent` in harness.ts still hand-rolls the same deadline skeleton (`deadline` + `while` + local sleep, 15s). It predates this PR, but now that the primitive exists a screen above it, it is the last holdout: `pollUntil(async () => (findConflictingSession(await client.sessions.list()) ? null : true), { timeoutMs: 15000 })`. Not this PR's debt.

No structural regressions, no new spaghetti, and the refactor deletes complexity rather than moving it. Ship it.

---

scripts/ios-regressions/harness.ts
hasparus: Taken in 3195cbe, together with the fresh-snapshot failure dump from the other note — the window now closes on a probe, and the `last.scoped` slot plus its flow-analysis lore are gone.

---

hasparus: The per-request bound landed in 35055d7 — `AbortSignal.timeout(5000)` on each readiness fetch, with aborts treated as transient like any other probe failure, so `pollUntil` keeps the 60s window authoritative.

On the pre-merge "avoid class-based HTML regexes" check: the regexes never touch the device — they parse the *served HTML* precisely to learn the accessible names and nav label the spec then uses on-device, which is the semantic path the check asks for. Matching the a11y tree by role alone was tried and disproven across several device runs (bare hour texts like "10" also label unrelated nodes). The coupling to the template is real, loud (both guards throw before any device time), and documented at the regex block.

---

scripts/ios-regressions/scrubber-longpress.test.ts
```diff
@@ -21,55 +33,206 @@ const {
...
+  try {
+    scoped = await takeSnapshot(navName);
+  } catch (error) {
+    throw new Error(
+      `No snapshot succeeded in ${timeoutMs}ms; last error: ${String(lastSnapshotError ?? error)}`,
     );
-    if (sessionOnScreen) return;
-    await client.interactions.scroll({ ...deviceOptions, direction: "down", pixels: 450 });
-    await wait(300);
   }
```
hasparus, line 164: Non-blocking wording nit: this message can now be reached when every probe snapshot *succeeded* (just found no markers) and only this failure-time dump snapshot threw. In that case "No snapshot succeeded in Xms" overstates what happened — `lastSnapshotError` is null and the reader gets the dump error labelled as "last error", which reads as if the whole window failed to snapshot. Something like "Could not take the failure-time snapshot after Xms without markers; last probe error: ..." would keep the two failure modes distinct. Fine to leave as is — the `lastSnapshotError ?? error` fallback already preserves the information.

---

hasparus (commented):
## Thermo-nuclear review, round three — verdict: APPROVE

(Submitted as COMMENT because the platform blocks self-approval; treat this as an approval.)

Round two approved at 32faeb7. This round verifies the two commits that landed since (3195cbe, 35055d7) against the suggestions they implement, plus a fresh pass over the whole diff at 35055d7.

### Delta since 32faeb7 — faithful, and a net structural improvement

**`pollUntil` ends its window on a probe (harness.ts:42-53).** The `do/while` became `for (;;)` with the deadline check sitting between the probe and the sleep, exactly as suggested. The final interval of the window is now observed rather than slept away, which is the guarantee the absence-asserting scrubber spec actually needs, and the comment says so. Semantics otherwise unchanged: at least one probe always runs, a throwing probe still aborts.

**Failure dump takes one fresh scoped snapshot (scrubber-longpress.test.ts:155-174).** This is the best change of the round: the `last: { scoped }` slot — a workaround for TypeScript's closure flow analysis, flagged in round two — is deleted entirely rather than polished. The probe becomes straight-line (`scoped` → `railMarkersFrom` → return), and the failure path pays one extra roundtrip only after the window is already lost. Complexity removed, not rearranged. One wording nit on the fallback error message, inline, non-blocking.

**Bounded readiness fetches (harness.ts:267).** `AbortSignal.timeout(5000)` per request closes the real hole the comment names: `pollUntil` cannot interrupt a pending probe, so one stalled `fetch` could previously hold the hook far past the 60s window. 5s per attempt against a 1s interval and 60s window is sane; the abort surfaces as a caught error into `lastError` like any other failure.

**`centreOnScreen` checks both axes (harness.ts:67-76).** The horizontal check uses the bare viewport edges (no inset — Safari's chrome is top/bottom only), which is right. Both callers only ever *narrow* what counts as visible with it, and both specs treat "not visible" as retry/fail-loud, never as a pass — so the tightening cannot create a vacuous green. Worth noting: the two on-device green runs (327ee7a, 32faeb7) predate this commit, so the delta itself is device-unverified until the workflow runs on this head; the risk is low since the vertical check already proved the rect and viewport share a coordinate space.

### Fresh pass over the full diff

- Checked the three rail regexes against `templates/chronology/_compact_schedule.html` as it exists on this branch: `RAIL_NAV_NAME` tolerates the newline between `class` and `aria-label` (`\s+`), and the hour patterns match the actual attribute order (`class` … `aria-label` … `>text</a>`). The loud `railNavName`/`markerNames` guards mean a template reorder fails at fetch time with a clear message, not as a mystery device timeout.
- The `eventPath` default fix (`/chronology/event/…` → `/event/…`) matches the live URLconf (`adapters/web/django/urls.py:26`) and the workflow's readiness curl.
- No stale references to the removed `viewportRect`/`assertPageReady` anywhere (including `warmup.ts` and `mobile.yml`).
- Structure is healthy: `pollUntil` is now the single polling idiom used five times across harness and both specs; `describeNode` and the visibility convention each have one home; both test files sit under 250 lines with the harness at 300. No spaghetti growth, no wrapper churn — the round-two/three deltas deleted incidental state (`last` slot, `viewportRect`) rather than adding any.
- `tsc -p scripts/ios-regressions/tsconfig.json` and `oxlint --deny-warnings scripts/ios-regressions` both pass at 35055d7.

### Verdict

APPROVE. The delta implements the accepted suggestions faithfully, removes a workaround instead of decorating it, and introduces no structural regression. One non-blocking wording nit inline.

---

PR #814, opened 2026-08-08
Plan/konwencik agenda export

fancysnake:
## Summary by CodeRabbit

* **Documentation**
  * Added an implementation plan for exporting event agendas to Google Sheets.
  * Documented sheet columns, source data mappings, empty-value handling, and configurable icons and photos.
  * Described synchronization behavior, including updates, cleanup, session splitting, and preservation of existing sheet data.
* **Localization**
  * Updated Polish translation catalog metadata and formatting markers.
* **Tests**
  * Improved local end-to-end test configuration for subdomains such as `another.localhost`.

### DIFF
--- a/docs/plans/konwencik-agenda-export.md
+++ b/docs/plans/konwencik-agenda-export.md
@@ -3,9 +3,10 @@
 ## Where we are
 
 Konwencik is an external program app that reads an event's agenda from a
-Google spreadsheet in a fixed layout of its own. Today nothing pushes our
-schedule anywhere: an organizer who wants the program in Konwencik retypes
-it, and every timetable change after that is a manual re-edit.
+Google spreadsheet in a fixed layout of its own. Konwencik owns that sheet
+and shares it with us. Today nothing pushes our schedule into it: an
+organizer who wants the program in Konwencik retypes it, and every timetable
+change after that is a manual re-edit.
 
 Every part needed to do this already exists:
 
@@ -14,10 +15,9 @@ Every part needed to do this already exists:
 - `EventIntegration` binds a connection to an event with a `config_json`
   (plumbing, validated by `check()`) and a `settings_json` (the
   operator-editable recipe) — exactly the split the proposal importer uses.
-- `GoogleSheetsWriter.write_rows` in `links/google_docs.py` does an atomic
-  full-tab replace, padding out to the previous data's extent so a shorter
-  export leaves no stale rows behind. `SHEETS_WRITE_SCOPES` is already
-  defined; a service account only needs editor access on the target sheet.
+- `GoogleSheetsWriter.write_rows` in `links/google_docs.py` writes a whole
+  tab in one atomic request. `SHEETS_WRITE_SCOPES` is already defined; a
+  service account only needs editor access on the target sheet.
 - `DiscountsExportService` plus `DiscountExportPageView` are a working
   export precedent — connection, decrypt, build rows, write, report a count.
 - `agenda_items.list_by_event` returns `AgendaItemDTO`s already carrying
@@ -43,14 +43,14 @@ what feeds three of them is.
 
 | Column | Source |
 | --- | --- |
-| `id` | `Session.pk` |
+| `id` | `Session.pk` — the row key, never rewritten |
 | `day` | agenda item start, `%d.%m.%Y` in `settings.TIME_ZONE` |
 | `start` | agenda item start, `%H:%M` |
 | `end` | agenda item end, `%H:%M` |
 | `title` | `Session.title` |
 | `description` | `Session.description` |
 | `speaker` | `Session.display_name` |
-| `room` | `Space.name` |
+| `room` | `{leaf} ({immediate parent})`, e.g. `RPG 1 (Piętro 1)`; a root space has no parentheses |
 | `room_position` | always empty |
 | `block` | name of the session's highest-priority track |
 | `type` | proposal category name |
@@ -59,6 +59,41 @@ what feeds three of them is.
 | `icon_background_color` | colour of the same track that produced `block` |
 
 <!-- markdownlint-enable MD013 -->
+
+An icon is an opaque string in Konwencik's own prefixed notation
+(`fa:star-filled` and friends, several formats behind a prefix). We do not
+parse, validate or preview it — we carry it.
+
+## The sheet is theirs, so we upsert
+
+The sheet is Konwencik's, not ours: it may hold rows we never wrote and
+columns we do not know about. So the export never replaces the tab.
+
+`id` is the deduplicator. A run reads the whole tab, maps the `id` column
+back to row numbers, and then:
+
+- a session already in the sheet has its cells updated in place, `id`
+  untouched;
+- a session not in the sheet is appended;
+- a row whose `id` we do not recognise — a manual row, or a session that no
+  longer exists — is left exactly as it is;
+- cells in columns we do not manage keep their values.
+
+**Unscheduling never deletes a row.** A session that loses its agenda item
+(or its confirmation, when `confirmed_only` is on) keeps its row and its
+title, description, speaker, category, icon and colour; only `day`, `start`,
+`end`, `room` and `room_position` are cleared. Konwencik then sees a session
+with no time and no place, which is what happened.
+
+Columns are located by the header row rather than assumed by position, so a
+column Konwencik reorders or inserts does not corrupt the export. A sheet
+whose header row has no `id` column is an error with a hint, not a guess —
+writing into a sheet we cannot key is how the sheet gets shredded.
+
+The merge is pure list-of-lists work in the mill, which is where it gets
+unit-tested. `links` grows a `read_rows` on the Google Sheets adapter (the
+private extent read already does the fetch), and one merged matrix goes back
+in a single write, so a failed request leaves the sheet untouched.
 
 ## Shape of the fix
 
@@ -78,17 +113,18 @@ Its config model is just the spreadsheet id.
 
 **A new mill,** `mills/konwencik.py`. `KonwencikExportService` takes the
 agenda-item, track, session, integration and connection repositories, a
-decryptor and a `SheetWriterProtocol` — the `DiscountsExportService`
-constructor shape. `export(*, sphere_id, event_pk, pk)` loads the
-integration scoped to `event_pk` (panel object-scope rule), decrypts the
-connection secret, builds the rows and hands them to the writer, returning
-the row count.
+decryptor and a sheet port that both reads and writes — the
+`DiscountsExportService` constructor shape, one method wider.
+`export(*, sphere_id, event_pk, pk)` loads the integration scoped to
+`event_pk` (panel object-scope rule), decrypts the connection secret, reads
+the sheet, merges by `id` and writes the result back, returning how many
+rows it added and how many it updated.
 
 **Settings, keyed by slug** so a rename never drops a colour:
 
 ```python
 class KonwencikExportSettings(BaseModel):
-    category_icons: dict[str, str] = {}   # category slug -> icon name
+    category_icons: dict[str, str] = {}   # category slug -> "fa:star-filled"
     track_colors: dict[str, str] = {}     # track slug -> "#rrggbb"
     track_order: dict[str, int] = {}      # track slug -> priority
     photo_url_field: str = ""             # session field slug
@@ -102,8 +138,10 @@ class KonwencikExportSettings(BaseModel):
 priority the session belongs to decides both `block` and
 `icon_background_color`.
 
-`AgendaItemDTO` gains a `category_slug` (the repository already selects the
-category), so the icon map is keyed on a slug like everything else.
+`AgendaItemDTO` gains two fields: `category_slug` (the repository already
+selects the category), so the icon map is keyed on a slug like everything
+else, and `space_parent_name`, which the `room` format needs — one more
+`select_related` hop.
 
 ## The configuration page
 
@@ -113,15 +151,16 @@ page at `event/<slug>/export/<pk>/`, reached from a kind-dependent link in
 the Actions column of the integrations table. One page, no tab shell — the
 importer has five tabs because it has five jobs; this has one.
 
-- **Kategorie** — a row per `ProposalCategory`: name plus an icon input,
-  reusing the `panel:icon-preview` HTMX swatch that the session-field form
-  already uses.
+- **Kategorie** — a row per `ProposalCategory`: name plus a plain text icon
+  input. The icon vocabulary is Konwencik's, not ours, so there is no
+  preview and no validation beyond a length cap; the help text carries an
+  example (`fa:star-filled`).
 - **Bloki** — a row per `Track`: name, hex colour, priority. A blank colour
   means an empty cell. A newly created track shows up here on its own with
   blank values instead of being silently missing from a JSON blob.
 - **Nadpisania** — two selects over `session_fields.list_by_event`: which
-  field carries an external photo URL, which carries an icon name. The empty
-  option means no override.
+  field carries an external photo URL, which carries an icon string. The
+  empty option means no override.
 - **Synchronizacja** — the enable toggle and the interval.
 - `Zapisz`, which posts to the existing `save_settings`, and `Eksportuj`,
   which runs the export now.
@@ -149,9 +188,11 @@ integrations whose sync is off, whose event ended more than a day ago — a
 finished event must not push forever — and whose `last_sync_time` is inside
 the interval.
 
-Re-running is safe by construction: `write_rows` replaces the whole tab, so
-a tick over unchanged data is a no-op and a tick over a sheet someone edited
-by hand is a repair. No diffing, no change hash, no dedup.
+Re-running is safe by construction: the merge keys on `id`, so a tick over
+unchanged data writes the same matrix back and a tick after a timetable
+change moves exactly the cells that moved. Nothing accumulates, nothing
+duplicates, and rows Konwencik added themselves survive every run. No
+change hash, no state to keep between ticks.
 
 Each integration runs in its own `try`/`except` inside the step: one
 event's revoked service account must not stop every other event's sync. The
@@ -174,8 +215,10 @@ Each step is reachable through the panel on its own.
 1. Protocol split, `EXPORT` kind, `KonwencikSheetExporter.check()`,
    registry entry. *Demo:* create the integration in the panel, run Check,
    get a green result.
-2. `KonwencikExportService` plus the `Eksportuj` action. *Demo:* the sheet
-   fills with every column except icon, colour and photo, which stay blank.
+2. `read_rows` on the sheet adapter, the merge, `KonwencikExportService` and
+   the `Eksportuj` action. *Demo:* the sheet gains a row per scheduled
+   session with every column except icon, colour and photo; run it twice and
+   nothing duplicates; unschedule one and its row loses only time and room.
 3. The configuration page, and its values wired into the row builder.
    *Demo:* icons and colours land in the sheet.
 4. Sync settings, the last-run columns, the DBOS tick and the management
@@ -186,10 +229,16 @@ Steps 1 to 3 are useful without 4, and 4 adds no export logic.
 
 ## Testing
 
-Unit tests on the mill with fake repositories: column order and formats,
-timezone rendering, multi-track priority, a session field beating the
-category default, the `confirmed_only` filter, and a sweep that keeps going
-after one integration raises `SheetExportError`.
+Unit tests on the mill with fake repositories, the merge carrying most of
+them: an existing `id` updated in place, a new session appended, a foreign
+row left byte-identical, an unknown trailing column preserved, an
+unscheduled session keeping its row with only time and room cleared, a
+second run over an unchanged event producing the same matrix, and a header
+row without `id` refusing to write at all. Plus the row builder itself —
+column order and formats, the `{leaf} ({parent})` room with and without a
+parent, timezone rendering, multi-track priority, a session field beating
+the category default, the `confirmed_only` filter — and a sweep that keeps
+going after one integration raises `SheetExportError`.
 
 Integration tests on the views with `assert_response`: the export action
 redirects with its message, a `SheetExportError` re-renders with the hint, a
@@ -201,6 +250,9 @@ belong to `tests/e2e`.
 
 - A run log in the style of `ImportLogEntry`. The sheet is the output and a
   failed write leaves the previous data intact.
+- Deleting rows. Nothing we do removes a row from Konwencik's sheet; a
+  cancelled session is a row with no time and no room, and pruning it is
+  their call.
 - Alerting on a sync that has been failing for a day. It shows as a red row
   in the panel; notifying the organizer belongs with the notification engine
   (issue #617).
@@ -209,13 +261,11 @@ belong to `tests/e2e`.
 
 ## Open questions
 
-- Does Konwencik want a header row, or raw data from A1?
-- Are the icon names Heroicon names — what `SessionField.icon` already
-  stores and what `panel:icon-preview` renders — or Konwencik's own set? If
-  theirs, the preview endpoint is wrong for this page and the input is plain
-  text.
-- A session crossing midnight: one row with an `end` past 24:00, or split
-  the way the timetable does?
-- `room`: the leaf `Space.name` only, or the full "Building > Room" path?
+- A session crossing midnight goes out as one row for now, with an `end`
+  earlier than its `start`. Needs checking against what Konwencik does with
+  it; if it mis-renders, the fix is a second row, and `id` being the key
+  means that decision can wait.
+- The header row is assumed present, because it is how the `id` column is
+  located. To be confirmed against the sheet Konwencik shared.
 - `agenda_items.list_by_event` does not filter soft-deleted sessions, the
   same as the print pages today. Leave it, or filter here?
--- a/docs/plans/konwencik-agenda-export.md
+++ b/docs/plans/konwencik-agenda-export.md
@@ -43,7 +43,7 @@ what feeds three of them is.
 
 | Column | Source |
 | --- | --- |
-| `id` | `Session.pk` — the row key, never rewritten |
+| `id` | `Session.pk`, plus `+n` for the second and later day of a session that runs past midnight — the row key, never rewritten |
 | `day` | agenda item start, `%d.%m.%Y` in `settings.TIME_ZONE` |
 | `start` | agenda item start, `%H:%M` |
 | `end` | agenda item end, `%H:%M` |
@@ -60,9 +60,40 @@ what feeds three of them is.
 
 <!-- markdownlint-enable MD013 -->
 
-An icon is an opaque string in Konwencik's own prefixed notation
-(`fa:star-filled` and friends, several formats behind a prefix). We do not
-parse, validate or preview it — we carry it.
+An icon is an opaque string in Konwencik's own prefixed notation — their
+sample sheet uses `fa.gamepad`, `fa.trophy`, `fa.comments`, and the prefix
+selects among several icon formats. We do not parse, validate or preview it
+— we carry it.
+
+## The sheet Konwencik hands us
+
+Two header rows, then data:
+
+<!-- markdownlint-disable MD013 -->
+
+| Row | Content |
+| --- | --- |
+| 1 | machine keys: `id`, `day`, `start`, `end`, `title`, `description`, `speaker`, `room`, `room_position`, `block`, `type`, `photo_url`, `icon`, `icon_background_color` |
+| 2 | Polish labels for humans: `Id`, `Dzień`, `Start`, `Koniec`, `Tytuł`, `Opis`, `Prowadzący`, `Sala`, `Sala - stanowisko`, `Blok`, `Rodzaj programu`, `Link do zdjęcia`, `Ikona`, `Tło` |
+| 3+ | data, seeded with three sample rows (`id` 0, 1, 2) whose description says to delete them |
+
+<!-- markdownlint-enable MD013 -->
+
+<!-- markdownlint-disable MD013 -->
+
+```text
+0    08.02.2024    17:00    18:00    Gra konwentowa    …    Konplace        Misz-Masz    Gra        fa.gamepad    #22AAFF
+```
+
+<!-- markdownlint-enable MD013 -->
+
+Row 1 is what we key on: columns are found by machine key, so reordering or
+inserting a column is safe and row 2 is never touched. Data starts at row 3.
+
+The sample rows are ordinary data rows to us. Their ids are small integers,
+so an event whose session pks happen to be 0, 1 or 2 would overwrite them —
+which is the fate the sample rows ask for anyway. Worth a line in the panel
+help text: delete the sample rows before the first export.
 
 ## The sheet is theirs, so we upsert
 
@@ -72,22 +103,40 @@ columns we do not know about. So the export never replaces the tab.
 `id` is the deduplicator. A run reads the whole tab, maps the `id` column
 back to row numbers, and then:
 
-- a session already in the sheet has its cells updated in place, `id`
+- a segment already in the sheet has its cells updated in place, `id`
   untouched;
-- a session not in the sheet is appended;
-- a row whose `id` we do not recognise — a manual row, or a session that no
-  longer exists — is left exactly as it is;
+- a segment not in the sheet is appended;
+- a row whose `id` is not one of ours — a manual row, a sample row we have
+  not reached, a session from another event — is left exactly as it is;
 - cells in columns we do not manage keep their values.
 
-**Unscheduling never deletes a row.** A session that loses its agenda item
-(or its confirmation, when `confirmed_only` is on) keeps its row and its
-title, description, speaker, category, icon and colour; only `day`, `start`,
-`end`, `room` and `room_position` are cleared. Konwencik then sees a session
-with no time and no place, which is what happened.
+**A session that runs past midnight becomes one row per day.** The first row
+keeps the bare session id, the following ones get `+1`, `+2`, and so on, and
+each carries the slice of the session that falls on its day, split at local
+midnight:
 
-Columns are located by the header row rather than assumed by position, so a
-column Konwencik reorders or inserts does not corrupt the export. A sheet
-whose header row has no `id` column is an error with a hint, not a guess —
+```text
+7      Sob    22:00    00:00    Nocne granie
+7+1    Nie    00:00    05:00    Nocne granie
+```
+
+Everything but `day`, `start` and `end` repeats across the segments.
+
+**Pruning is ours, and it is a full sweep.** Every run compares the whole
+sheet against the event's current state, not just the sessions it is
+writing. A row whose id belongs to this event but has no matching segment
+right now — the session was unscheduled, unconfirmed, soft-deleted, or
+simply no longer stretches into a second day — keeps its row and its title,
+description, speaker, category, icon and colour, and has `day`, `start`,
+`end`, `room` and `room_position` cleared. Konwencik then sees a session
+with no time and no place, which is what happened. Nothing is ever deleted.
+
+Deciding "belongs to this event" needs the event's alive session ids, so the
+sessions repository gains a narrow method returning exactly that. The id
+column is parsed as `pk` or `pk+n`; anything else is somebody else's row.
+
+Columns are located by the machine-key header row rather than assumed by
+position. A sheet with no `id` column is an error with a hint, not a guess —
 writing into a sheet we cannot key is how the sheet gets shredded.
 
 The merge is pure list-of-lists work in the mill, which is where it gets
@@ -118,7 +167,12 @@ decryptor and a sheet port that both reads and writes — the
 `export(*, sphere_id, event_pk, pk)` loads the integration scoped to
 `event_pk` (panel object-scope rule), decrypts the connection secret, reads
 the sheet, merges by `id` and writes the result back, returning how many
-rows it added and how many it updated.
+rows it added, updated and pruned.
+
+Soft-deleted sessions never produce segments, so they get pruned like any
+other session that is no longer scheduled. `agenda_items.list_by_event` does
+not filter them today (neither do the print pages), so the export filters
+against the alive session ids it already needs for pruning.
 
 **Settings, keyed by slug** so a rename never drops a colour:
 
@@ -215,10 +269,12 @@ Each step is reachable through the panel on its own.
 1. Protocol split, `EXPORT` kind, `KonwencikSheetExporter.check()`,
    registry entry. *Demo:* create the integration in the panel, run Check,
    get a green result.
-2. `read_rows` on the sheet adapter, the merge, `KonwencikExportService` and
-   the `Eksportuj` action. *Demo:* the sheet gains a row per scheduled
-   session with every column except icon, colour and photo; run it twice and
-   nothing duplicates; unschedule one and its row loses only time and room.
+2. `read_rows` on the sheet adapter, the merge (split, upsert, prune),
+   `KonwencikExportService` and the `Eksportuj` action. *Demo:* the sheet
+   gains a row per scheduled session with every column except icon, colour
+   and photo; run it twice and nothing duplicates; unschedule one and its
+   row loses only time and room; schedule one across midnight and it becomes
+   two rows.
 3. The configuration page, and its values wired into the row builder.
    *Demo:* icons and colours land in the sheet.
 4. Sync settings, the last-run columns, the DBOS tick and the management
@@ -231,14 +287,20 @@ Steps 1 to 3 are useful without 4, and 4 adds no export logic.
 
 Unit tests on the mill with fake repositories, the merge carrying most of
 them: an existing `id` updated in place, a new session appended, a foreign
-row left byte-identical, an unknown trailing column preserved, an
-unscheduled session keeping its row with only time and room cleared, a
-second run over an unchanged event producing the same matrix, and a header
-row without `id` refusing to write at all. Plus the row builder itself —
-column order and formats, the `{leaf} ({parent})` room with and without a
-parent, timezone rendering, multi-track priority, a session field beating
-the category default, the `confirmed_only` filter — and a sweep that keeps
-going after one integration raises `SheetExportError`.
+row left byte-identical, the Polish label row untouched, an unknown trailing
+column preserved, a second run over an unchanged event producing the same
+matrix, and a header row without `id` refusing to write at all.
+
+The prune and split rules get their own: an unscheduled session keeping its
+row with only time and room cleared, the same for an unconfirmed and for a
+soft-deleted one, a midnight-crossing session producing `7` and `7+1` with
+the right slices, and a session that stops crossing midnight leaving `7+1`
+pruned rather than stale or deleted.
+
+Plus the row builder itself — column order and formats, the
+`{leaf} ({parent})` room with and without a parent, timezone rendering,
+multi-track priority, a session field beating the category default — and a
+sweep that keeps going after one integration raises `SheetExportError`.
 
 Integration tests on the views with `assert_response`: the export action
 redirects with its message, a `SheetExportError` re-renders with the hint, a
@@ -261,11 +323,13 @@ belong to `tests/e2e`.
 
 ## Open questions
 
-- A session crossing midnight goes out as one row for now, with an `end`
-  earlier than its `start`. Needs checking against what Konwencik does with
-  it; if it mis-renders, the fix is a second row, and `id` being the key
-  means that decision can wait.
-- The header row is assumed present, because it is how the `id` column is
-  located. To be confirmed against the sheet Konwencik shared.
-- `agenda_items.list_by_event` does not filter soft-deleted sessions, the
-  same as the print pages today. Leave it, or filter here?
+- A pruned `+1` row keeps its title and speaker, so a session that stopped
+  crossing midnight leaves a timeless duplicate of itself in the sheet.
+  Clearing the whole row would be tidier but breaks the rule that we only
+  ever own time and place; leaving it is the consistent choice. Confirm
+  Konwencik renders a timeless row as absent rather than as a broken entry.
+- Times go out as `%H:%M`, so midnight is `00:00`. Konwencik's own sample
+  writes `0:00` — confirm the leading zero is not load-bearing.
+- Whether `id` `0` in the sample rows can collide with a real session pk
+  depends on the sequence; deleting the sample rows first avoids the
+  question entirely.
--- a/docs/plans/konwencik-agenda-export.md
+++ b/docs/plans/konwencik-agenda-export.md
@@ -1,0 +1,221 @@
+# Export the scheduled agenda to Konwencik
+
+## Where we are
+
+Konwencik is an external program app that reads an event's agenda from a
+Google spreadsheet in a fixed layout of its own. Today nothing pushes our
+schedule anywhere: an organizer who wants the program in Konwencik retypes
+it, and every timetable change after that is a manual re-edit.
+
+Every part needed to do this already exists:
+
+- `Connection` (sphere-scoped, encrypted service-account JSON) and
+  `FernetDecryptor` hold and unlock the Google credentials.
+- `EventIntegration` binds a connection to an event with a `config_json`
+  (plumbing, validated by `check()`) and a `settings_json` (the
+  operator-editable recipe) — exactly the split the proposal importer uses.
+- `GoogleSheetsWriter.write_rows` in `links/google_docs.py` does an atomic
+  full-tab replace, padding out to the previous data's extent so a shorter
+  export leaves no stale rows behind. `SHEETS_WRITE_SCOPES` is already
+  defined; a service account only needs editor access on the target sheet.
+- `DiscountsExportService` plus `DiscountExportPageView` are a working
+  export precedent — connection, decrypt, build rows, write, report a count.
+- `agenda_items.list_by_event` returns `AgendaItemDTO`s already carrying
+  space name, session title and description, `presenter_name` (the session's
+  `display_name`), category name, session id, start and end.
+- `agenda_items.list_by_track` gives the session-to-track mapping;
+  `sessions.list_field_values_for_sessions` gives the dynamic answers.
+- `inits/dbos_scheduler.py` runs `@DBOS.scheduled` cron workflows in-process,
+  deduped across gunicorn workers, with management commands as the manual
+  floor.
+
+So this is wiring, not new machinery. The one thing that does not fit is the
+`IntegrationImplementation` protocol: it is import-shaped
+(`fetch_questions`, `fetch_headers`, `fetch_responses`), and an exporter
+would have to carry three dead stubs.
+
+## Field mapping
+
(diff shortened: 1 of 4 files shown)


docs/plans/konwencik-agenda-export.md
```diff
@@ -0,0 +1,335 @@
...
+back to row numbers, and then:
+
+- a segment already in the sheet has its cells updated in place, `id`
+  untouched;
+- a segment not in the sheet is appended;
+- a row whose `id` is not one of ours — a manual row, a sample row we have
+  not reached, a session from another event — is left exactly as it is;
+- cells in columns we do not manage keep their values.
+
+**A session that runs past midnight becomes one row per day.** The first row
```
fancysnake: **Validate this before you build anything else — as written it is the largest complexity generator in the plan, and it rests on an assumption nobody has checked.**

The per-day split buys one thing: rows whose `day`/`start`/`end` are internally consistent. It costs:

- a compound id grammar (`pk` vs `pk+n`) that every read of the sheet must parse, and that the prune sweep must reason about;
- a segment splitter with a local-midnight boundary (so a DST night is its own edge case);
- a prune rule that exists *only* for segments — "a session that stops crossing midnight leaves `7+1` pruned rather than stale or deleted";
- roughly half the test list in the Testing section;
- and, by your own Open Questions, a permanently timeless duplicate row of the session sitting in Konwencik's sheet forever, which you can't justify and are hoping renders as absent.

All of that disappears if Konwencik tolerates one row with `end` < `start` (`22:00` → `05:00`). One row per session, `id` == `Session.pk` with no grammar, no splitter, no segment-prune state, no ghost.

So: **ask Konwencik whether a single row spanning midnight renders**, and only if the answer is no build the split. Bundle it with the `0:00` vs `00:00` question in Open Questions — that one already shows you don't know how their parser reads the time columns, which is exactly the knowledge this whole mechanism depends on.

If the answer *is* no, say so here in one line, and then the `+n` machinery is earned and I have no objection to it.

docs/plans/konwencik-agenda-export.md
```diff
@@ -0,0 +1,335 @@
...
+event's revoked service account must not stop every other event's sync. The
+outcome lands in three new columns on `EventIntegration` —
+`last_sync_time`, `last_sync_status`, `last_sync_detail` (reversible
+migration) — shown on the integrations table and the export page
+("Ostatnia synchronizacja: 12:15 — 84 punkty programu", or the error hint).
+They are deliberately not in `settings_json`: an operator saving settings
+mid-tick would clobber them. There is no run history; the sheet is the
+output.
+
+On the scheduled path `confirmed_only` is forced on, whatever the setting
```
fancysnake: **This is a one-off mode that makes the setting a lie. Delete the setting.**

`confirmed_only` is rendered in the panel, saved to `settings_json`, and then silently overridden on the scheduled path. An organizer who unticks it, saves, and waits will watch nothing happen and have no way to find out why. Two triggers, two meanings for one field, and a branch in the row builder that exists to serve a case the plan itself argues should never happen.

Your own justification — "a mid-planning tick pushing unconfirmed items into a public app is the failure mode worth designing out" — is an argument that unconfirmed items should *never* go to Konwencik, not that they should go on one path and not the other. The manual button has the same failure mode; it just fires when someone is watching.

Drop `confirmed_only` from `KonwencikExportSettings`, always export confirmed items. That removes a settings field, a form widget, a branch, and the sentence at line 261. If someone later wants to preview an unconfirmed programme, that's a separate feature with a separate name — not a boolean with an asterisk.

docs/plans/konwencik-agenda-export.md
```diff
@@ -0,0 +1,335 @@
...
+not filter them today (neither do the print pages), so the export filters
+against the alive session ids it already needs for pruning.
+
+**Settings, keyed by slug** so a rename never drops a colour:
+
+```python
+class KonwencikExportSettings(BaseModel):
+    category_icons: dict[str, str] = {}   # category slug -> "fa:star-filled"
+    track_colors: dict[str, str] = {}     # track slug -> "#rrggbb"
+    track_order: dict[str, int] = {}      # track slug -> priority
```
fancysnake: **Two dicts keyed by the same key is a missing model, and the third one may not need to exist.**

`track_colors` and `track_order` are both `track slug -> …`. They are written together by the same form rows, read together by the same lookup ("the colour of the same track that produced `block`"), and can drift apart — a track present in one and absent from the other is representable and meaningless. That's the shape that turns into `settings.track_colors.get(slug, "")` / `settings.track_order.get(slug, BIG)` scattered through the row builder.

One model:

```python
class TrackStyle(BaseModel):
    color: str = ""
    priority: int = 0

class KonwencikExportSettings(BaseModel):
    tracks: dict[str, TrackStyle] = {}
    category_icons: dict[str, str] = {}
    ...
```

One lookup, one form row per track, one thing that can be missing.

Separately — **does `priority` earn its keep at all?** It is a new configuration concept whose entire job is breaking a tie for multi-track sessions. `Track.Meta.ordering` is already `["name"]`, so "first track by name" is deterministic, needs no storage, no form column, no per-event data entry, and no explanation in the panel. Ship that; add the priority column when an organizer actually complains that the wrong block won. If you keep it, the `TrackStyle` merge above still applies.

docs/plans/konwencik-agenda-export.md
```diff
@@ -0,0 +1,335 @@
...
+writing into a sheet we cannot key is how the sheet gets shredded.
+
+The merge is pure list-of-lists work in the mill, which is where it gets
+unit-tested. `links` grows a `read_rows` on the Google Sheets adapter (the
+private extent read already does the fetch), and one merged matrix goes back
+in a single write, so a failed request leaves the sheet untouched.
+
+## Shape of the fix
+
+**Split the implementation protocol.** `IntegrationImplementation` keeps
```
fancysnake: **`mills/chronology.py` is at 996 lines. This pushes it over 1000 — decompose it first.**

The protocol split plus the `EXPORT` kind plus the narrowing at the import-only methods all land on `EventIntegrationsService`, which lives at the bottom of that file. That's a presumptive blocker.

The seam is already visible: `EventIntegrationsService` and `IntegrationImplementationNotFoundError` are a self-contained unit that shares nothing with the session/content services above it — they take their own repos, their own registry, and are wired independently in `inits/services.py`. Move them to `mills/integrations.py` as step 0 of this plan, before the protocol split touches them. Pure move, no behaviour, reviewable in isolation, and it leaves both halves well under the line.

Same check applies to the file the exporter lands in — see the comment on `links/google_docs.py` below.

docs/plans/konwencik-agenda-export.md
```diff
@@ -0,0 +1,335 @@
...
+what every integration has — `kind`, `config_model`, `check()`. A new
+`ProposalSourceImplementation` extends it with the three `fetch_*` methods.
+The registry is typed to the base protocol; the import-only methods on
+`EventIntegrationsService` narrow before calling, which they are already
```
fancysnake: **"Narrow before calling" needs a concrete mechanism, and the two available ones are both bad.**

`ProposalSourceImplementation` is a `Protocol`. Narrowing from the base to it at runtime means either `@runtime_checkable` + `isinstance` — which only checks that the method *names* exist, so an exporter that happens to grow a `fetch_headers` passes — or `hasattr`, which is worse. Neither is a type boundary; both push the failure to the moment somebody calls `fetch_responses` on an exporter.

Make the boundary explicit at composition time instead, where the real information lives. `inits/services.py` already knows exactly which concrete class it is registering, so build the typed lookup there:

```python
registry: dict[Id, IntegrationImplementation] = {...}          # every impl, for check/create/list
sources: dict[Id, ProposalSourceImplementation] = {...}        # import impls only
```

and hand both to `EventIntegrationsService`. `fetch_questions`/`fetch_headers`/`fetch_responses` read `self._sources`, which is *already typed correctly* — no narrowing, no `isinstance`, no runtime surprise, and the existing "empty result on a registry miss" behaviour you point at falls out unchanged from a miss on `sources`. `check()`, `create()` and `list_implementations()` keep reading `self._registry`.

That's a smaller diff than the narrowing, and mypy actually checks it.

docs/plans/konwencik-agenda-export.md
```diff
@@ -0,0 +1,335 @@
...
+Deciding "belongs to this event" needs the event's alive session ids, so the
+sessions repository gains a narrow method returning exactly that. The id
+column is parsed as `pk` or `pk+n`; anything else is somebody else's row.
+
+Columns are located by the machine-key header row rather than assumed by
+position. A sheet with no `id` column is an error with a hint, not a guess —
+writing into a sheet we cannot key is how the sheet gets shredded.
+
+The merge is pure list-of-lists work in the mill, which is where it gets
+unit-tested. `links` grows a `read_rows` on the Google Sheets adapter (the
```
fancysnake: **Before adding `read_rows`: the sheet port lives in `pacts/discounts.py`.**

`SheetWriterProtocol` is declared inside the discounts pact. If `mills/konwencik.py` imports it from there, Konwencik depends on Discounts for the definition of "a thing that talks to Google Sheets" — a cross-feature boundary leak that gets harder to unpick with every new sheet consumer.

Adding `read_rows` is the moment to move it. Give it its own pact module (`pacts/sheets.py`) holding the port and `SheetExportError`, re-point `pacts/discounts.py` at it, and let both mills import from the same neutral home. It's a mechanical move and it keeps step 2 honest.

While you're there: decide whether the port is one protocol with both methods or two. The exporter needs read+write; `DiscountsExportService` needs write only. One protocol with both means the discounts service depends on a method it never calls — small, but it is the same coupling in miniature.

fancysnake (commented):
Thermo-nuclear quality pass. This is a docs-only PR, so the review is of the *design* — but the design is detailed enough to be reviewed as an implementation, and a few of its decisions bake in complexity that would be expensive to unwind after step 2 ships.

The skeleton is right: separate mill, merge as pure list-of-lists, registry dispatch, DBOS tick next to `printables_reminders_tick`, object-scoped `export(*, sphere_id, event_pk, pk)`. Every claim about existing machinery checks out against the code.

What I'd want resolved before step 1:

- **the midnight split is unvalidated and it is the single largest complexity source in the plan** — it drives the `pk+n` id grammar, the segment splitter, a stale-segment prune rule, half the test list, and a ghost row the plan's own open questions cannot justify. Confirm the assumption before building on it.
- **`confirmed_only` is a setting the panel shows and one of the two triggers ignores.** Drop it.
- **two slug-keyed dicts over the same key** (`track_colors`, `track_order`) is a missing model, and `track_order` may not need to exist at all.
- **this pushes `mills/chronology.py` from 996 lines over 1000.**

Each action item is an inline comment of its own below.

docs/plans/konwencik-agenda-export.md
```diff
@@ -0,0 +1,335 @@
...
+    sync_interval_minutes: int = 60
+```
+
+`track_order` is what makes a multi-track session deterministic: the lowest
+priority the session belongs to decides both `block` and
+`icon_background_color`.
+
+`AgendaItemDTO` gains two fields: `category_slug` (the repository already
+selects the category), so the icon map is keyed on a slug like everything
+else, and `space_parent_name`, which the `room` format needs — one more
```
fancysnake: **`category_slug` is free — keep it. `space_parent_name` is not — drop it.**

`_SELECT_RELATED` in `links/db/django/agenda_item.py` is already `("session", "session__category", "space")`, so `category_slug` costs one attribute read on a row that's already loaded. No objection.

`space_parent_name` is different: it adds a `space__parent` join to `_SELECT_RELATED`, which is shared by `read`, `list_by_event`, `list_by_track`, `read_by_session` and `list_overlapping_in_space` — i.e. every timetable page, every print page, and the conflict check on every scheduling write pay for a column that only Konwencik reads. That's feature logic leaking into a shared path.

You don't need it. `spaces.list_by_event(event_pk)` already returns every space of the event with `parent_id`, `name` and `pk` in one query, and `AgendaItemDTO` already carries `space_id`. Build `{space_id: "leaf (parent)"}` in the mill from that list — one extra repo call in one service, zero effect on anything else, and the `{leaf} ({parent})` formatting rule ends up unit-tested in the mill next to the rest of the row builder rather than baked into a repository DTO.

(`SpaceOptionDTO` proves the shape is already canonical: `name` + `group` = immediate parent's name, empty for a root.)

docs/plans/konwencik-agenda-export.md
```diff
@@ -0,0 +1,335 @@
...
+  operator-editable recipe) — exactly the split the proposal importer uses.
+- `GoogleSheetsWriter.write_rows` in `links/google_docs.py` writes a whole
+  tab in one atomic request. `SHEETS_WRITE_SCOPES` is already defined; a
+  service account only needs editor access on the target sheet.
+- `DiscountsExportService` plus `DiscountExportPageView` are a working
+  export precedent — connection, decrypt, build rows, write, report a count.
+- `agenda_items.list_by_event` returns `AgendaItemDTO`s already carrying
+  space name, session title and description, `presenter_name` (the session's
+  `display_name`), category name, session id, start and end.
+- `agenda_items.list_by_track` gives the session-to-track mapping;
```
fancysnake: **`list_by_track` is the wrong tool for the session→track mapping — it's per-track and it over-fetches.**

Its signature is `list_by_track(track_pk, *, facilitator_pks=None) -> list[AgendaItemDTO]`. To learn which tracks each session belongs to you'd call it once per track — N queries, each running the full `_SELECT_RELATED` join and materialising complete agenda-item DTOs, to extract nothing but membership pairs. Then the mill inverts N lists into the map it actually wanted.

Ask for what you need: a `dict[int, list[int]]` of session pk → track pks for the event, one query over the `Session.tracks` through-table, on the tracks repository (which already owns `list_space_pks`, `list_manager_pks`, `list_manager_names_by_event` — same shape of narrow lookup). The mill then joins it against the `track slug → style` settings directly.

Fix the sentence here too; as written it sends the implementer down the N-query path.

docs/plans/konwencik-agenda-export.md
```diff
@@ -0,0 +1,335 @@
...
+The registry is typed to the base protocol; the import-only methods on
+`EventIntegrationsService` narrow before calling, which they are already
+shaped for (each returns an empty result on a registry miss).
+
+**A new kind and implementation.** `IntegrationKind.EXPORT` and
+`IntegrationImplementationId.KONWENCIK_SHEET_PUSHER`, implemented by
+`KonwencikSheetExporter` in `links/google_docs.py`. Its `check()` reuses the
+existing `_probe` against the spreadsheet metadata endpoint with write
+scopes, so an operator finds out about a missing share before the first run.
+Its config model is just the spreadsheet id.
```
fancysnake: **"Just the spreadsheet id" hides a load-bearing assumption: *which tab*.**

`GoogleSheetsWriter` resolves the target through `_first_tab_title`, i.e. `meta.sheets[0]` — the first tab, whatever it is. For discounts that's fine: we own the sheet. Here we don't. Konwencik's spreadsheet is theirs, they may add tabs, and someone reordering tabs silently redirects our merge into an unrelated one.

The header-key guard catches the friendly version of that (no `id` column → refuse, per line 138), but not the hostile one: a *second* Konwencik-shaped tab, or a copy someone made to experiment with. We'd merge into the wrong one and report success.

Either put the tab title in the config model next to the spreadsheet id (defaulting to the first tab, resolved and reported by `check()` so the operator sees which tab we picked), or state plainly here that first-tab is the contract and have `check()` name the tab it found in its hint. Silent positional resolution against a sheet we don't own isn't a contract.

docs/plans/konwencik-agenda-export.md
```diff
@@ -0,0 +1,335 @@
...
+**Split the implementation protocol.** `IntegrationImplementation` keeps
+what every integration has — `kind`, `config_model`, `check()`. A new
+`ProposalSourceImplementation` extends it with the three `fetch_*` methods.
+The registry is typed to the base protocol; the import-only methods on
+`EventIntegrationsService` narrow before calling, which they are already
+shaped for (each returns an empty result on a registry miss).
+
+**A new kind and implementation.** `IntegrationKind.EXPORT` and
+`IntegrationImplementationId.KONWENCIK_SHEET_PUSHER`, implemented by
+`KonwencikSheetExporter` in `links/google_docs.py`. Its `check()` reuses the
```
fancysnake: **`links/google_docs.py` is 483 lines and already holds three unrelated clients. Split it before adding a fourth thing.**

Today it carries the Forms API schema models and parser (`_FormSchema`, `_Question`, `_source_question`, …), `GoogleDocsProposalImporter`, and `GoogleSheetsWriter` — Forms and Sheets sharing a module because they share an OAuth session builder. This PR adds `read_rows` and `KonwencikSheetExporter` on top, taking it comfortably past 600 and giving it a third reason to change.

The seam is obvious and clean: `links/google_forms.py` (form schema models, `_source_question`, `GoogleDocsProposalImporter`) and `links/google_sheets.py` (`_a1_quote`, `_pad_to_extent`, `GoogleSheetsWriter`, `read_rows`, `KonwencikSheetExporter`), with `_build_session`, `_CredentialsError` and `_probe` in a small shared module. Do it as part of step 1, before the new code lands, so the split diff stays a pure move.

docs/plans/konwencik-agenda-export.md
```diff
@@ -0,0 +1,335 @@
...
+
+`AgendaItemDTO` gains two fields: `category_slug` (the repository already
+selects the category), so the icon map is keyed on a slug like everything
+else, and `space_parent_name`, which the `room` format needs — one more
+`select_related` hop.
+
+## The configuration page
+
+A JSON textarea cannot offer a row per category and per track, and an
+organizer should not be typing slugs. The settings live on their own panel
```
fancysnake, line 262: **Name the file this page lives in.**

The plan gives the URL but not the module, and the nearest precedent — `gates/web/django/chronology/panel/views/google_docs_import.py` — is **1213 lines**. "Next to the other integration page" is the default an implementer will reach for, and it makes the worst file in the panel worse. `integrations.py` (333) is the other magnet.

Say it explicitly: `panel/views/konwencik_export.py`, its own module, page view + form + the two POST actions. It's a distinct page with a distinct service; there is no reason for it to share a file with the importer's five tabs.

Same for the form — the per-row generated form belongs beside the view, not in `panel/forms.py`.

docs/plans/konwencik-agenda-export.md
```diff
@@ -0,0 +1,335 @@
...
+  means an empty cell. A newly created track shows up here on its own with
+  blank values instead of being silently missing from a JSON blob.
+- **Nadpisania** — two selects over `session_fields.list_by_event`: which
+  field carries an external photo URL, which carries an icon string. The
+  empty option means no override.
+- **Synchronizacja** — the enable toggle and the interval.
+- `Zapisz`, which posts to the existing `save_settings`, and `Eksportuj`,
+  which runs the export now.
+
+It is one Django form built from event data: fields are generated per row
```
fancysnake: **"Fields generated per row with slug-suffixed names and cleaned back into the three dicts" is where this page turns into string surgery.**

Building `icon_<slug>`, `color_<slug>`, `priority_<slug>` and then reassembling them by prefix-splitting field names in `clean()` means slugs travel through the form as string fragments — and slugs can contain `_`, so any parse of `color_<slug>` back into `(prefix, slug)` has to split on the *first* underscore and trust the rest. That's the kind of code that works until a track slug is named `blok_glowny`.

Django already has the boring answer: a formset, one form per track and one per category, each carrying its slug in a hidden field and validating it against the event's own slugs in `clean_slug()`. Fields keep their real names (`color`, `priority`), the "drop slugs the event does not own" rule you correctly call out at line 224 becomes one method on one form rather than a filter over parsed keys, and rendering a row per track is what formsets are for.

If you keep the flat-form approach, at minimum specify the delimiter and the slug charset here — but I'd just use the formset.

konwencik.md
```diff
@@ -0,0 +1,23 @@
+# Integration with Konwencik system
```
fancysnake: **This shouldn't be in the repo root, and probably shouldn't be in the repo.**

It's the raw requirements note the plan was written from, and `docs/plans/konwencik-agenda-export.md` now supersedes every line of it — the Field mapping table says the same things, more precisely, plus the sources. Two copies of a column list that must agree is one copy too many; the root one will rot first and it's the one someone stumbling into the repo root will read.

Delete it, and if any of it is load-bearing context the plan doesn't carry — the parenthetical "(configure)" notes, say — fold that into the plan instead. If you want to keep the original ask verbatim as provenance, it belongs in the issue or in `docs/plans/` next to the plan, not at the top level next to `manage.py`.

src/ludamus/locale/pl/LC_MESSAGES/django.po
```diff
@@ -643,37 +643,30 @@ msgid "%(name)s (manages their own enrollment)"
 msgstr "%(name)s (zarządza swoimi zapisami)"
 
 #: src/ludamus/adapters/web/django/views.py
-#, python-brace-format
 msgid "Enrolled: {}"
```
fancysnake, line 647: **Revert this file — it's unrelated regeneration churn, and it removes a real safety check.**

This PR is documentation. The only substantive change here is that ten entries lose their `#, python-brace-format` flag (plus a `POT-Creation-Date` bump and a stray blank line at the end of the obsolete block).

That flag is what makes `msgfmt --check-format` verify the translation's `{}` placeholders against the msgid. Without it, a translator can drop or mistype a placeholder in `"Zapisani: {}"` or `"Nie ma wystarczającej liczby miejsc. {} …"` and the catalogue compiles clean — the failure shows up as an `IndexError` in a user's face. Small regression, but a real one, and gained nothing.

It's also a guaranteed conflict with any branch that touches i18n. `git checkout main -- src/ludamus/locale/pl/LC_MESSAGES/django.po` and leave the catalogue to a PR that actually changes strings. If the flags really did disappear from a legitimate `makemessages` run on main, that's its own bug worth a papercut — but not this PR's diff.

docs/plans/konwencik-agenda-export.md
```diff
@@ -0,0 +1,335 @@
...
+
+Deciding "belongs to this event" needs the event's alive session ids, so the
+sessions repository gains a narrow method returning exactly that. The id
+column is parsed as `pk` or `pk+n`; anything else is somebody else's row.
+
+Columns are located by the machine-key header row rather than assumed by
+position. A sheet with no `id` column is an error with a hint, not a guess —
+writing into a sheet we cannot key is how the sheet gets shredded.
+
+The merge is pure list-of-lists work in the mill, which is where it gets
```
fancysnake: ## Thermo-nuclear code quality review

**`list[list[str]]` is the contract this whole feature is built on, and it is why the merge, the prune and the split all end up as index arithmetic.**

`SheetWriterProtocol.write_rows` in `pacts/discounts.py` takes `rows: list[list[str]]`, and the plan extends that shape to `read_rows` and then does everything on it: locate columns by scanning row 1, map the id column back to row numbers, update in place while leaving "cells in columns we do not manage", clear five named columns on a prune. Every one of those is `row[cols["day"]]`, and the invariant that makes it safe — this row is at least as long as the widest header — lives nowhere.

Give the mill a value object instead: built once from the raw matrix, get/set by machine key, `to_matrix()` back out.

- "a sheet with no `id` column is an error with a hint" becomes one construction-time failure instead of a guard every operation has to remember not to skip;
- a ragged row (someone typed three cells into a manual row) gets padded once, not defensively at each access;
- the merge/prune/split tests assert on keys rather than positions, so a column reordering in Konwencik's sheet doesn't rewrite the test file — which is the whole point of keying on row 1;
- `read_rows`' return type stops being the same anonymous shape as `write_rows`' input, which is currently the only thing keeping the two honest.

This is the largest legibility win available here, and it is far cheaper to decide now than after three sets of tests are written against positional lists.

docs/plans/konwencik-agenda-export.md
```diff
@@ -0,0 +1,335 @@
...
+`export_konwencik` management command as the manual floor — the
+`send_printables_reminders` arrangement. Fifteen minutes is the resolution;
+`sync_interval_minutes` is what an operator actually sets.
+
+The sweep needs a repository method listing EXPORT-kind integrations across
+all events together with their `sphere_id` (through `Event.sphere`), so
+`read_secret` keeps its sphere guard outside a request. It skips
+integrations whose sync is off, whose event ended more than a day ago — a
+finished event must not push forever — and whose `last_sync_time` is inside
+the interval.
```
fancysnake: ## Thermo-nuclear code quality review

**Delete `sync_interval_minutes`. It is a second scheduler layered on the first, and it is the only reason anything has to *read* `last_sync_time`.**

Four lines below, the plan makes the case against itself: "Re-running is safe by construction... Nothing accumulates, nothing duplicates... No change hash, no state to keep between ticks." Then the sweep keeps state between ticks — `last_sync_time` — purely to throttle a run the plan just argued is free.

What the knob costs: a settings field an operator has no basis to tune (why 60 and not 45?), a comparison in the sweep, a second unit of time that has to be reconciled against "fifteen minutes is the resolution", and the whole class of confusion where the two disagree (interval 5, cron 15). What it buys: fewer Sheets calls.

If quota is the real reason, say that and put one number in `specs` rather than exposing it. The sweep is already bounded by "sync on" and "event not finished more than a day ago", so call volume tracks live events, not tenants. Then the sweep filter is two conditions instead of three, `last_sync_time` becomes display-only, and the "Synchronizacja" section of the panel page is a single toggle.

docs/plans/konwencik-agenda-export.md
```diff
@@ -0,0 +1,335 @@
...
+Re-running is safe by construction: the merge keys on `id`, so a tick over
+unchanged data writes the same matrix back and a tick after a timetable
+change moves exactly the cells that moved. Nothing accumulates, nothing
+duplicates, and rows Konwencik added themselves survive every run. No
+change hash, no state to keep between ticks.
+
+Each integration runs in its own `try`/`except` inside the step: one
+event's revoked service account must not stop every other event's sync. The
+outcome lands in three new columns on `EventIntegration` —
+`last_sync_time`, `last_sync_status`, `last_sync_detail` (reversible
```
fancysnake: ## Thermo-nuclear code quality review

**Three new columns on a shared model for one kind, on a model whose own convention is a JSON column.**

`EventIntegration` today is `kind`, `implementation`, `connection`, `display_name`, `config_json`, `settings_json`, `questions_snapshot_json`. Adding `last_sync_time`, `last_sync_status`, `last_sync_detail` gives every proposal-source and ticketing row three permanently-null columns, and it quietly contradicts the plan's own "Not in scope: a run log in the style of `ImportLogEntry`" — because that is exactly what those three columns are, a one-row run log denormalised onto the parent.

The stated reason for keeping them out of `settings_json` is right, and it survives with one column instead of three: a `last_run_json` is still separate from what an operator's save writes, so it still can't be clobbered mid-tick. One migration, one column, matching three existing columns on the same model, and the thing inside it is a Pydantic model the panel renders — rather than two loose strings whose relationship (`status` says error, so what is in `detail`?) is documented only in prose.

Nothing queries these anyway once the interval gate goes; and if the gate stays, that is the one argument for a real `last_sync_time` column — so decide that first and let it pick the storage.

docs/plans/konwencik-agenda-export.md
```diff
@@ -0,0 +1,335 @@
...
+
+Everything but `day`, `start` and `end` repeats across the segments.
+
+**Pruning is ours, and it is a full sweep.** Every run compares the whole
+sheet against the event's current state, not just the sessions it is
+writing. A row whose id belongs to this event but has no matching segment
+right now — the session was unscheduled, unconfirmed, soft-deleted, or
+simply no longer stretches into a second day — keeps its row and its title,
+description, speaker, category, icon and colour, and has `day`, `start`,
+`end`, `room` and `room_position` cleared. Konwencik then sees a session
```
fancysnake: ## Thermo-nuclear code quality review

**"Which columns we own" is stated three times in three vocabularies, and it is really two sets, not one.**

The upsert section says cells in columns we do not manage keep their values. The field-mapping table names the fourteen columns we write. This paragraph names five we clear on a prune — and, by omission, nine we deliberately leave standing on a pruned row. A reader has to diff a table against a sentence to discover that `title` is written-but-never-cleared while `room` is both.

That is a missing model. One ordered tuple of column descriptors — machine key, Polish label, how the value is produced, whether a prune clears it — and the consumers stop restating it: the row builder maps over it, the merge writes only the keys it declares, the prune clears the flagged subset, the header check compares against its keys. Adding `room_position` to the cleared set later is then one flag, not an edit in three places that can drift apart silently.

It also turns the open question at the bottom into a visible decision: "a pruned `+1` row keeps its title and speaker" stops being an emergent property of a hand-written list and becomes a flag someone can look at.

docs/plans/konwencik-agenda-export.md
```diff
@@ -0,0 +1,335 @@
...
+A `@DBOS.scheduled("*/15 * * * *")` workflow sits next to
+`printables_reminders_tick`, building its service through a new
+`build_konwencik_export()` in `inits/builders.py`, with an
+`export_konwencik` management command as the manual floor — the
+`send_printables_reminders` arrangement. Fifteen minutes is the resolution;
+`sync_interval_minutes` is what an operator actually sets.
+
+The sweep needs a repository method listing EXPORT-kind integrations across
+all events together with their `sphere_id` (through `Event.sphere`), so
+`read_secret` keeps its sphere guard outside a request. It skips
```
fancysnake: ## Thermo-nuclear code quality review

**On the scheduled path, `sphere_id` is a guard fed from the row it guards.**

`ConnectionRepository.read_secret(sphere_id, pk)` does `Connection.objects.get(pk=pk, sphere_id=sphere_id)` and raises `NotFoundError` on a miss. That check exists to stop a *request* in sphere A from reaching a connection in sphere B. The sweep has no request and no principal: it reads the integration, reads that integration's own `sphere_id` through `Event.sphere`, and hands both back into the same query. It cannot fail. So the repository method grows a join whose only purpose is to satisfy a signature, and `sphere_id` ends up meaning "check this" in one caller and "I already know" in the other.

Separate them. The panel entry point does the scoping the panel object-scope rule actually requires — resolve the integration under `event_pk` and the sphere, 404 on a miss — and hands a resolved integration to the export, which takes no `sphere_id` at all. The sweep calls that same inner method directly with what it loaded. One authz decision, made in the one place that has a principal, instead of a parameter that has to be manufactured to keep a guard quiet.

docs/plans/konwencik-agenda-export.md
```diff
@@ -0,0 +1,335 @@
...
+## Steps
+
+Each step is reachable through the panel on its own.
+
+1. Protocol split, `EXPORT` kind, `KonwencikSheetExporter.check()`,
+   registry entry. *Demo:* create the integration in the panel, run Check,
+   get a green result.
+2. `read_rows` on the sheet adapter, the merge (split, upsert, prune),
+   `KonwencikExportService` and the `Eksportuj` action. *Demo:* the sheet
+   gains a row per scheduled session with every column except icon, colour
```
fancysnake: ## Thermo-nuclear code quality review

**Step 2 demos the `Eksportuj` action, but the page that carries it lands in step 3.**

The plan puts `Eksportuj` on the export configuration page, and that page is step 3. Step 2's demo requires pressing it. So step 2 either builds part of that page early — in which case say which part, and step 3 stops being "the configuration page" — or it hangs the action off the integrations table and step 3 moves it, which is throwaway work in a plan that otherwise takes "each step is reachable through the panel on its own" seriously.

The cheap resolution is that `Eksportuj` lives in the Actions column from step 2 and *stays* there: it is a row action on an integration, exactly the shape `Check` already has, and it works with nothing configured (icons and colours come out blank, which is what step 2's demo says anyway). Step 3 then adds a page that configures, and never has to own a button.

fancysnake:

**`tests/e2e/playwright.config.ts` does not belong in this branch** (local commit `dfc65cb` — not pushed yet, so there is no line to anchor to).

The change itself is right: `bypass: "localhost,127.0.0.1"` misses `another.localhost`, and adding `.localhost` fixes the sphere-subdomain case. But it lands on a branch whose entire purpose is a plan document, under a commit message — `test: cover the lines this branch changes` — that describes something else entirely: it adds no test and covers no line this branch changes.

Same call as the `django.po` churn: split it out and merge it on its own, where a reviewer looking for the proxy-bypass fix can find it and the plan PR stays a plan PR.

docs/plans/konwencik-agenda-export.md
```diff
@@ -0,0 +1,335 @@
...
+`event_pk` (panel object-scope rule), decrypts the connection secret, reads
+the sheet, merges by `id` and writes the result back, returning how many
+rows it added, updated and pruned.
+
+Soft-deleted sessions never produce segments, so they get pruned like any
+other session that is no longer scheduled. `agenda_items.list_by_event` does
+not filter them today (neither do the print pages), so the export filters
+against the alive session ids it already needs for pruning.
+
+**Settings, keyed by slug** so a rename never drops a colour:
```
fancysnake: ## Thermo-nuclear code quality review

**"keyed by slug so a rename never drops a colour" is backwards — a rename is exactly what drops it.**

`TrackRepository.update` in `links/db/django/repositories/venues.py` re-derives the slug on every name change:

```python
if track.name != data["name"]:
    base_slug = slugify(data["name"])
    track.slug = self.generate_unique_slug(track.event_id, base_slug, exclude_pk=pk)
```

`ProposalCategoryRepository.update` does the same for categories. So renaming *Bloki RPG* to *Sesje RPG* moves the slug, `track_colors["bloki-rpg"]` becomes an orphan, and the track reads back a blank colour. The single failure mode this sentence claims to design out is the one the design causes. The pk is the identifier that actually survives a rename.

Key the three dicts on pk (`dict[int, ...]`). It costs nothing and it collapses two other problems on the way:

- the "fields generated per row with slug-suffixed names" scheme loses its delimiter ambiguity entirely — a pk is `[0-9]+`, so `color_17` parses unambiguously and the "which underscore do I split on" question never arises;
- "slugs in the POST that the event does not own are dropped" becomes set membership against the pks you already loaded to render the rows.

Same question for `photo_url_field` / `icon_field`, which store a session-field slug: confirm that slug is not regenerated on rename before you rely on it, or store the field pk too.

What a slug buys is a readable `settings_json` in the admin. That is not worth silent data loss on a rename.

docs/plans/konwencik-agenda-export.md
```diff
@@ -0,0 +1,335 @@
...
+`DiscountsExportService` constructor shape, one method wider.
+`export(*, sphere_id, event_pk, pk)` loads the integration scoped to
+`event_pk` (panel object-scope rule), decrypts the connection secret, reads
+the sheet, merges by `id` and writes the result back, returning how many
+rows it added, updated and pruned.
+
+Soft-deleted sessions never produce segments, so they get pruned like any
+other session that is no longer scheduled. `agenda_items.list_by_event` does
+not filter them today (neither do the print pages), so the export filters
+against the alive session ids it already needs for pruning.
```
fancysnake: ## Thermo-nuclear code quality review

**One set is being asked two opposite questions, and the soft-delete case in the Testing section cannot pass.**

`Session` extends `SoftDeleteModel`, whose default `objects` manager filters `deleted_at__isnull=True` (`all_objects` is the escape hatch). So "the event's alive session ids" excludes a soft-deleted session — and that one set is then given both jobs:

- ownership, at line 134: *deciding "belongs to this event" needs the event's alive session ids*;
- the export filter, here: *the export filters against the alive session ids it already needs for pruning*.

For a soft-deleted session those want opposite answers. Its pk must be **in** the ownership set, or its sheet row is classified as somebody else's and left untouched — while line 128 promises that row is pruned, and Testing demands a test for exactly it ("the same for an unconfirmed and for a soft-deleted one"). As specified that test fails: the row keeps stale `day`, `start`, `end` and `room` forever, for a session the organizer deleted.

`confirmed_only` has the identical shape — an unconfirmed session must be owned but not exported — so this is not a one-off.

They are two sets, and naming them that way makes the merge trivial to read:

- **owned** — every session pk this event has, soft-deleted included. Answers "is this row ours?"
- **exportable** — alive, confirmed, scheduled. Produces segments.

Prune is then `owned - exportable`. The three prune cases the plan enumerates (unscheduled, unconfirmed, soft-deleted) stop being three special cases in prose and become one set subtraction with nothing to remember.

The narrow repository method the plan asks for is the **owned** one, and it has to go through `Session.all_objects` deliberately. Say so here — the obvious implementation reaches for `objects` and gets it silently wrong, and nothing downstream will notice.

docs/plans/konwencik-agenda-export.md
```diff
@@ -0,0 +1,335 @@
...
+`ProposalSourceImplementation` extends it with the three `fetch_*` methods.
+The registry is typed to the base protocol; the import-only methods on
+`EventIntegrationsService` narrow before calling, which they are already
+shaped for (each returns an empty result on a registry miss).
+
+**A new kind and implementation.** `IntegrationKind.EXPORT` and
+`IntegrationImplementationId.KONWENCIK_SHEET_PUSHER`, implemented by
+`KonwencikSheetExporter` in `links/google_docs.py`. Its `check()` reuses the
+existing `_probe` against the spreadsheet metadata endpoint with write
+scopes, so an operator finds out about a missing share before the first run.
```
fancysnake: ## Thermo-nuclear code quality review

**A GET probe cannot prove write access, so this sentence promises something `check()` will not deliver.**

`_probe` in `links/google_docs.py` is `session.get(url, timeout=10)` and maps the status code. A service account with **viewer** access on Konwencik's spreadsheet gets `200` from the metadata endpoint, whatever scopes the token carries — Google authorizes the operation, not the scope list, and reading is the operation being performed. So `check()` goes green, the operator ticks the box, and the first export dies on a `403` from `values.update` with nobody watching (or, worse, on a scheduled tick, into `last_sync_status`).

"An operator finds out about a missing share before the first run" is the entire reason `check()` exists for an exporter — an importer's check has the same shape *because* reading is what an importer does. Reusing it verbatim here quietly drops the one guarantee.

Make the probe perform the operation it is claiming to authorize. `spreadsheets.batchUpdate` with `{"requests": []}` is the cheap honest version: it hits the write path, returns `200` with empty replies for an editor and `403` for a viewer, and changes nothing. `_probe` would need to take the verb and body, which is a small generalization of something already parameterized by `url` and `what`.

If you would rather not add a write probe, then say plainly here that `check()` proves the sheet exists and is readable, and that a missing editor share surfaces on the first export — and make sure the error hint for that `403` names the fix ("share the sheet with `<service-account email>` as editor"). What is not acceptable is the current wording, which tells the implementer the guarantee is already there.

docs/plans/konwencik-agenda-export.md
```diff
@@ -0,0 +1,335 @@
...
+sessions repository gains a narrow method returning exactly that. The id
+column is parsed as `pk` or `pk+n`; anything else is somebody else's row.
+
+Columns are located by the machine-key header row rather than assumed by
+position. A sheet with no `id` column is an error with a hint, not a guess —
+writing into a sheet we cannot key is how the sheet gets shredded.
+
+The merge is pure list-of-lists work in the mill, which is where it gets
+unit-tested. `links` grows a `read_rows` on the Google Sheets adapter (the
+private extent read already does the fetch), and one merged matrix goes back
```
fancysnake: ## Thermo-nuclear code quality review

**"The private extent read already does the fetch" is the tell: `read_rows` alongside `write_rows` reads the same tab twice per run, and the second read is the one the write trusts.**

Look at what `write_rows` does today, internally and unconditionally:

1. `_first_tab_title` — a metadata GET;
2. `_old_extent` — a values GET of the whole tab, to learn height and width;
3. the PUT, padded to that extent.

Bolt `read_rows` on the side and every export becomes: metadata GET, values GET (for the merge), metadata GET again, values GET again (for the padding), PUT. Four round trips where two are the same request, and the tab title resolved twice — which, given the first-tab question raised elsewhere on this plan, is two independent chances to pick a different tab.

The correctness cost is worse than the latency. The matrix the mill merged came from read #1; the extent the write pads to came from read #2. Anything that happens in between — Konwencik pasting a row — is read by neither consistently: the merge did not see it, and the padding sizes the write to a sheet that now has it. The plan's "one merged matrix goes back in a single write" is true of the PUT and false of the operation.

The port granularity is wrong. What the mill wants is not `read_rows` + `write_rows`, it is one call: *fetch the tab, hand me the matrix, take back what I return, write it.* That is one title resolution, one read, one write, no window, and `_old_extent` disappears entirely because the merged matrix is derived from the very rows whose extent you are padding to — the mill already knows the height and width.

```python
class SheetPort(Protocol):
    def update_rows(
        self, *, secret: bytes, spreadsheet_id: str,
        merge: Callable[[list[list[str]]], list[list[str]]],
    ) -> None: ...
```

`DiscountsExportService` keeps `write_rows` (it owns the sheet and replaces it wholesale; it has no read to fold in), so this adds a method rather than changing one. It also gives the per-integration serialization asked for elsewhere on this plan exactly one place to live.

docs/plans/konwencik-agenda-export.md
```diff
@@ -0,0 +1,335 @@
...
+
+The sample rows are ordinary data rows to us. Their ids are small integers,
+so an event whose session pks happen to be 0, 1 or 2 would overwrite them —
+which is the fate the sample rows ask for anyway. Worth a line in the panel
+help text: delete the sample rows before the first export.
+
+## The sheet is theirs, so we upsert
+
+The sheet is Konwencik's, not ours: it may hold rows we never wrote and
+columns we do not know about. So the export never replaces the tab.
```
fancysnake: ## Thermo-nuclear code quality review

**"The export never replaces the tab" is not what the code does. It rewrites every cell of it, as raw text.**

`GoogleSheetsWriter.write_rows` PUTs the whole matrix to `{tab}!A1` with `valueInputOption=RAW`, padded out to the previous extent. Whatever your merge does at the list-of-lists level, the request that lands on Konwencik's spreadsheet overwrites every cell in the tab — including the cells the plan promises are "left exactly as it is" and "keep their values".

On the discounts sheet that is fine: we own it and we generate every cell. Here it is not, and the difference is what survives a round trip through `values.get` + `values.update`:

- a **formula** in a manual row comes back from the read as its rendered value and goes back in as a literal string. It is gone, permanently, and no run afterwards can tell.
- a **date or number** cell (Konwencik's own sample writes `08.02.2024` and `0:00`) is read as its formatted string and written back with `RAW`, which stores text. The cell stops being a date. If their app or any sheet formula reads it typed, it breaks — silently, on rows we were not even trying to touch.
- the same applies to the Polish label row 2, which the plan explicitly promises is never touched. It is rewritten every run; it merely happens to be text.

So the Testing item "a foreign row left byte-identical" is only assertable at the values layer, which is exactly the layer that cannot see the damage. A unit test over `list[list[str]]` will pass forever while the sheet degrades.

Two ways out, and the plan should pick one on the page rather than leave it to whoever writes the adapter:

- **Write only the cells we changed.** `values.batchUpdate` with one range per touched row (or per touched cell run) never sends a cell we do not own, so foreign rows, unknown columns and row 2 are untouched at the storage layer, not just the values layer. It is more requests, but it is the only version where the plan's promises are literally true — and it makes the "one atomic write" claim weaker, so say which you are trading.
- **Keep the whole-tab PUT and state the real contract:** we rewrite every cell of the tab as text, so formulas and typed cells anywhere in it will not survive. Then that belongs in the panel help text next to "delete the sample rows", because it is a thing Konwencik has to agree to.

`valueInputOption` deserves a decision either way: `RAW` stores our `#22AAFF` and `08.02.2024` as text; `USER_ENTERED` would coerce them the way a human typing would, which is its own set of surprises (`0:00` becoming a duration). Name the choice.

docs/plans/konwencik-agenda-export.md
```diff
@@ -0,0 +1,335 @@
...
+`AgendaItemDTO` gains two fields: `category_slug` (the repository already
+selects the category), so the icon map is keyed on a slug like everything
+else, and `space_parent_name`, which the `room` format needs — one more
+`select_related` hop.
+
+## The configuration page
+
+A JSON textarea cannot offer a row per category and per track, and an
+organizer should not be typing slugs. The settings live on their own panel
+page at `event/<slug>/export/<pk>/`, reached from a kind-dependent link in
```
fancysnake: ## Thermo-nuclear code quality review

**A kind-dependent link is dispatching on the wrong axis: the page is Konwencik's, not `EXPORT`'s.**

`IntegrationKind` is `IMPORT` and `TICKETING` today — already one direction and one domain sharing an enum. This plan adds `EXPORT`, and then hangs a link off it that leads to a page rendering `KonwencikExportSettings`: a per-`ProposalCategory` icon column, a per-`Track` colour column, `photo_url_field`, `icon_field`. None of that is a property of "an integration that exports". All of it is a property of `KONWENCIK_SHEET_PUSHER`.

The second EXPORT implementation — an ICS feed, a different convention app, an export back into the same Google Forms sheet — arrives at that link and gets Konwencik's form. The fix at that point is a branch on implementation inside a page keyed on kind, which is the exact spaghetti shape this plan is otherwise careful about.

Dispatch on `implementation`, which is the field that actually determines what the page contains, and the enum you already have to switch on to pick a config model. The Actions column asks the implementation for its settings URL (or looks it up in a small map beside the registry), `kind` goes back to meaning what it means everywhere else — which capabilities the row has — and adding a second exporter is a new row in that map instead of an edit to a shared page.

Worth saying in the same breath what `EXPORT` is *for*, since it is a new value in a public enum: if the answer is only "so the Actions column can render a different link", the kind is not carrying its weight and `implementation` was the discriminator all along. If it does gate something real — the sweep's "list EXPORT-kind integrations across all events" query, say — say that here, because that is the justification and it is not currently written down.

docs/plans/konwencik-agenda-export.md
```diff
@@ -0,0 +1,335 @@
...
+simply no longer stretches into a second day — keeps its row and its title,
+description, speaker, category, icon and colour, and has `day`, `start`,
+`end`, `room` and `room_position` cleared. Konwencik then sees a session
+with no time and no place, which is what happened. Nothing is ever deleted.
+
+Deciding "belongs to this event" needs the event's alive session ids, so the
+sessions repository gains a narrow method returning exactly that. The id
+column is parsed as `pk` or `pk+n`; anything else is somebody else's row.
+
+Columns are located by the machine-key header row rather than assumed by
```
fancysnake: ## Thermo-nuclear code quality review

**"Located by the machine-key header row" assumes the header row is a function. This codebase already knows it is not.**

`_disambiguate` exists in `links/google_docs.py` for precisely this reason, and its comment says why: *"Two sheet columns with the same header would collapse to a single dict key, losing one column's value."* That was written against the importer's sheet. The same hazard is worse here, because the importer only reads and this writes.

The plan's key-to-index map inherits every one of those cases and handles none:

- **duplicate keys.** Konwencik copies the `icon` column to experiment. `{key: index}` keeps the last one; the merge writes `icon` into the copy and leaves the real column stale. Nothing raises.
- **blank trailing headers.** Row 1 has 14 keys and a 15th empty cell (Sheets is generous with these). `"" -> 14` joins the map. Harmless until a prune iterates the map instead of the declared key list.
- **whitespace and case.** ` id` is not `id`. The refusal path ("a sheet with no `id` column is an error with a hint") fires on a sheet that visibly has one, and the operator has no way to see the space.

The last one is the good failure. The first two are silent and they corrupt.

Say here that the header row is validated, not just scanned, and what the rule is: strip each cell, require the fourteen keys we write to appear exactly once, ignore everything else, and refuse with a hint that names the offending key when they do not. That is one construction-time check, it is where the `id`-missing refusal already wants to live, and it makes "reordering or inserting a column is safe" a claim the code enforces rather than one the sheet is trusted to honour.

(This is also the strongest argument for the column-descriptor model raised elsewhere on this plan: the set of keys to validate against and the set to write are the same list, and here they are two.)

tests/e2e/playwright.config.ts
```diff
@@ -80,8 +80,10 @@ export default defineConfig({
           proxy: {
             server: proxyServer,
             // The proxy only fronts external egress; localhost (the app
-            // under test) and 127.0.0.1 must bypass it.
-            bypass: "localhost,127.0.0.1",
+            // under test) and 127.0.0.1 must bypass it. `localhost` matches
+            // that host alone, so the sphere subdomains the app serves
+            // (another.localhost, …) need the suffix rule too.
+            bypass: "localhost,.localhost,127.0.0.1",
```
fancysnake: ## Thermo-nuclear code quality review

**This is the only executable change in a documentation PR, it is unrelated to Konwencik, and the commit that added it is called `test: cover the lines this branch changes` while adding no test.**

`dfc65cb1b` touches this file and nothing else: four lines of proxy config and a comment. There is no test in it. The message describes work that is not in the diff, which means the next person reading `git log` for "when did the e2e proxy bypass change" will not find it here.

On the change itself: widening `bypass` to `.localhost` so sphere subdomains skip the egress proxy is plausibly correct and probably a real fix someone hit while running e2e. That is the problem — it is a real fix, affecting every e2e run, and it is riding in a PR whose entire subject is a plan document. It gets reviewed by people looking at a markdown file, and it lands or reverts with a plan that may well be rewritten before any of it is built.

Split it out: its own commit on its own PR, with a message saying what broke (which spec, against which host) and a green run behind it. This branch goes back to the plan plus `konwencik.md`, and both of those are text.

If the e2e suite genuinely cannot run on this branch without it, say so in the PR description — that is a different argument and it is a fine one, but right now nothing on the PR makes the connection and the commit message actively points away from it.

hasparus:
Ran the automated stale-PR triage pass on this branch. Summary of what changed:

**Merge conflicts** — merged `origin/main` in (real merge, no rebase). Only one real conflict: `tests/e2e/playwright.config.ts`, where both branches added the same `bypass: "localhost,*.localhost,127.0.0.1"` fix with slightly different comment wording. Kept this branch's wording, dropped the duplicate. All other files auto-merged cleanly — this branch's diff was genuinely small (a plan doc, `django.po` metadata, and the e2e config).

**Failing `checks` CI job** — root cause was `mise run messages-check`: after merging in `main` (which added many new `{}`-placeholder strings since this PR was opened), the PL catalog was stale and missing `#, python-brace-format` flags on several `msgid`s. Regenerated it with `django-admin makemessages -l pl --add-location=file` and committed the result as part of the merge commit. Verified both `mise run lint` and `mise run messages-check` pass locally now.

**Review feedback** — @fancysnake's own unresolved review comment on `django.po` (flagging that the catalog's `#, python-brace-format` flags were missing, e.g. on `"Enrolled: {}"`, and that this removes a real placeholder-safety check) is resolved by the regeneration above — those flags are restored. CodeRabbit's threads were all already resolved/outdated. No comments from Rams were found on this PR (the "Rams" bot doesn't appear to have reviewed this one). Left untouched: the organizational nitpick (from both CodeRabbit and fancysnake) that the `playwright.config.ts` / `django.po` changes belong in a separate PR from the plan doc — that's a scope/split preference for the author to decide, not a defect, and out of scope for an unattended fix.

Branch is pushed to `origin/plan/konwencik-agenda-export`. Not merging, per the triage routine.

---

src/ludamus/locale/pl/LC_MESSAGES/django.po
fancysnake: Not doing this here. This branch isn't ready for review — it carries the plan file, not actual documentation, so the catalogue diff will get sorted out along with the rest of it rather than in a targeted revert now.

For the record, the premise is also stale: the `python-brace-format` flag count is 11 on both this branch and `main`, and the remaining diff against `main` is a single `POT-Creation-Date` header line. No check was lost.

Issue #341, opened 2026-06-05
don't print eldritch time ranges


hasparus:
<img width="1251" height="752" alt="Image" src="https://github.com/user-attachments/assets/9201c2d9-1141-4715-80a8-f92957e8da48" />

We should not be printing "P4H" when all events are the same time length.

@fancysnake wdyt

also I'm not getting it in the test data, so I assume it's another misconfig-footgun in the panel?

<img width="979" height="787" alt="Image" src="https://github.com/user-attachments/assets/21ffddfe-c1da-4626-9e9c-36ccdd84ba7a" />

PR #815, opened 2026-08-08
fix(durations): never show a raw ISO duration (#341)

fancysnake:
`format_duration` echoed its input when the value didn't parse, so a stored "P4H" reached the session card, the proposal detail and the category editor verbatim. Production held five such values across 13 rows, all hand-typed or fed in by the legacy importer.

- one duration contract in `pacts/durations.py`: strict `parse_duration`, `build_duration`, and `normalize_duration` for loose input
- `format_duration` returns "" rather than the raw string; the three templates and both duration choice fields guard on the label
- every writer normalizes: the legacy feed (where "50min" and "110m" came from), the category settings POST, and the import recipe
- the recipe's per-option ISO text box becomes hours/minutes steppers, so no organizer is asked to type ISO
- data migration rewrites the stored values, soft-deleted rows included

### DIFF
--- a/src/ludamus/gates/web/django/chronology/forms.py
+++ b/src/ludamus/gates/web/django/chronology/forms.py
@@ -91,8 +91,10 @@ def build_session_details_form(
         "display_name": forms.CharField(label=_("Presenter name"), max_length=255),
     }
 
-    if durations:
-        duration_choices = [(d, format_duration(d)) for d in durations]
+    # A duration nothing can read is left out: an option is worth offering only
+    # when it can be labelled with a length.
+    duration_choices = [(d, label) for d in durations if (label := format_duration(d))]
+    if duration_choices:
         fields["duration"] = forms.ChoiceField(
             label=_("Duration"), choices=[("", "---"), *duration_choices]
         )
--- a/src/ludamus/gates/web/django/chronology/panel/views/google_docs_import.py
+++ b/src/ludamus/gates/web/django/chronology/panel/views/google_docs_import.py
@@ -30,6 +30,7 @@ from ludamus.gates.web.django.chronology.panel.views.base import (
 )
 from ludamus.mills.submissions.mapping import MissingKeyColumnsError, slugify
 from ludamus.pacts.chronology import IntegrationImplementationId, IntegrationKind
+from ludamus.pacts.durations import build_duration, parse_duration
 from ludamus.pacts.submissions import (
     DurationSpec,
     EntityRef,
@@ -79,7 +80,8 @@ class OptionEntity(TypedDict):
 
 class OptionDuration(TypedDict):
     option: str
-    iso: str
+    hours: int | None
+    minutes: int | None
 
 
 class OverrideRow(TypedDict):
@@ -391,16 +393,20 @@ def _option_entities(
 def _option_durations(
     question: SourceQuestion, target: QuestionTarget | None
 ) -> list[OptionDuration]:
-    # One row per source option, pre-filled with the operator-typed ISO 8601
-    # duration so the (hidden) editor is ready the moment the row becomes a
-    # session-duration target. A blank ISO leaves that option unmapped: the
+    # One row per source option, pre-filled with the length the operator picked
+    # so the (hidden) editor is ready the moment the row becomes a
+    # session-duration target. A blank length leaves that option unmapped: the
     # importer then skips rows whose answer hits this option.
     configured = target.values if target else {}
     rows: list[OptionDuration] = []
     for option in question.options:
         spec = configured.get(option)
-        iso = spec.iso if isinstance(spec, DurationSpec) else ""
-        rows.append({"option": option, "iso": iso})
+        hours, minutes = parse_duration(
+            spec.iso if isinstance(spec, DurationSpec) else ""
+        )
+        rows.append(
+            {"option": option, "hours": hours or None, "minutes": minutes or None}
+        )
     return rows
 
 
@@ -493,18 +499,27 @@ def _time_slot_values_from_post(
 
 
 def _duration_values_from_post(post: QueryDict, index: int) -> dict[str, QuestionValue]:
-    # Per-option ISO duration rows submit parallel arrays; a blank ISO drops
-    # that option (its answers will be skipped at import time).
+    # Per-option duration rows submit parallel arrays of hours and minutes; a
+    # zero-length option is dropped (its answers will be skipped at import
+    # time). ISO is composed here, never typed by the operator.
     values: dict[str, QuestionValue] = {}
     rows = zip(
-        post.getlist(f"droption_{index}"), post.getlist(f"driso_{index}"), strict=False
+        post.getlist(f"droption_{index}"),
+        post.getlist(f"drhours_{index}"),
+        post.getlist(f"drminutes_{index}"),
+        strict=False,
     )
-    for option, iso in rows:
-        clean_iso = iso.strip()
-        if not (option and clean_iso):
+    for option, hours, minutes in rows:
+        iso = build_duration(hours=_positive_int(hours), minutes=_positive_int(minutes))
+        if not (option and iso):
             continue
-        values[option] = DurationSpec(iso=clean_iso)
+        values[option] = DurationSpec(iso=iso)
     return values
+
+
+def _positive_int(raw: str) -> int:
+    text = (raw or "").strip()
+    return int(text) if text.isdigit() else 0
 
 
 def _entity_map_from_post(
--- a/src/ludamus/gates/web/django/forms.py
+++ b/src/ludamus/gates/web/django/forms.py
@@ -18,11 +18,9 @@ from ludamus.gates.web.django.dynamic_fields import (
     CustomAnswerFormMixin,
     build_dynamic_fields,
 )
-from ludamus.gates.web.django.templatetags.cfp_tags import (
-    build_duration,
-    format_duration,
-)
+from ludamus.gates.web.django.templatetags.cfp_tags import format_duration
 from ludamus.pacts.discounts import DiscountKind
+from ludamus.pacts.durations import build_duration
 from ludamus.pacts.images import ALLOWED_IMAGE_FORMATS, IMAGE_ACCEPT, LOGO_ACCEPT
 from ludamus.pacts.legacy import PromotionMode
 from ludamus.pacts.submissions import AccreditationType
@@ -686,15 +684,19 @@ class _ComposedDurationForm(SessionEditForm):
 
 def _duration_field(durations: Sequence[str]) -> forms.ChoiceField | None:
     # No configured durations means the steppers are the whole control, so the
-    # inherited free-text field is dropped (a None entry removes it).
-    if not durations:
+    # inherited free-text field is dropped (a None entry removes it). A stored
+    # duration nothing can read is dropped too: an option is worth offering
+    # only when it can be labelled with a length.
+    if not (
+        labelled := [(d, label) for d in durations if (label := format_duration(d))]
+    ):
         return None
     return forms.ChoiceField(
         required=False,
         label=_("Duration"),
         choices=[
             ("", "---"),
-            *((d, format_duration(d)) for d in durations),
+            *labelled,
             # Kept last: the template reveals the steppers with a CSS
             # :last-child selector rather than JavaScript.
             (CUSTOM_DURATION, _("Custom")),
--- a/src/ludamus/gates/web/django/templatetags/cfp_tags.py
+++ b/src/ludamus/gates/web/django/templatetags/cfp_tags.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import re
 from typing import TYPE_CHECKING, Any
 
 from django import template
@@ -8,6 +7,7 @@ from django.utils import timezone
 from django.utils.translation import gettext as _
 
 from ludamus.gates.web.django.helpers import placeholder_cover_url
+from ludamus.pacts.durations import parse_duration
 
 if TYPE_CHECKING:
     from ludamus.pacts import SessionDTO
@@ -154,28 +154,6 @@ def format_field_value(value: Any) -> str:  # type: ignore[misc] # ruff:ignore[a
     return str(value)
 
 
-def parse_duration(iso_duration: str) -> tuple[int, int]:
-    """Split an ISO 8601 duration into hours and minutes.
-
-    Returns:
-        (hours, minutes), both 0 for anything unparsable.
-    """
-    if not (match := re.match(r"PT(?:(\d+)H)?(?:(\d+)M)?", iso_duration or "")):
-        return 0, 0
-    return int(match.group(1) or 0), int(match.group(2) or 0)
-
-
-def build_duration(*, hours: int, minutes: int) -> str:
-    """Compose an ISO 8601 duration, empty when both parts are zero.
-
-    Returns:
-        A string like "PT1H30M", "PT45M" or "".
-    """
-    if not hours and not minutes:
-        return ""
-    return "PT" + (f"{hours}H" if hours else "") + (f"{minutes}M" if minutes else "")
-
-
 @register.filter
 def format_duration(iso_duration: str) -> str:
     """Format ISO 8601 duration string to human-readable format.
@@ -184,11 +162,9 @@ def format_duration(iso_duration: str) -> str:
         iso_duration: ISO 8601 duration string (e.g., "PT1H45M", "PT30M", "PT2H")
 
     Returns:
-        Human-readable duration (e.g., "1h 45min", "30min", "2h")
+        Human-readable duration ("1h 45min", "30min", "2h"), empty when the
+        stored value is unreadable — a raw ISO string is never shown.
     """
-    if not iso_duration:
-        return ""
-
     hours, minutes = parse_duration(iso_duration)
 
     if hours and minutes:
@@ -197,4 +173,4 @@ def format_duration(iso_duration: str) -> str:
         return f"{hours}h"
     if minutes:
         return f"{minutes}min"
-    return iso_duration
+    return ""
--- a/src/ludamus/links/db/django/migrations/0142_normalize_durations.py
+++ b/src/ludamus/links/db/django/migrations/0142_normalize_durations.py
@@ -1,0 +1,37 @@
+from django.db import migrations
+
+from ludamus.pacts.durations import normalize_duration
+
+
+def normalize_stored_durations(apps, schema_editor):
+    del schema_editor
+    # Durations reached storage as whatever their author typed ("P4H",
+    # "50min", "110m"). Rewrite them in the one format readers understand;
+    # anything unreadable becomes unset rather than shown raw.
+    session_model = apps.get_model("db_main", "Session")
+    # Soft-deleted rows included: restoring a session must not bring an
+    # unreadable duration back with it.
+    for session in session_model.objects.exclude(duration="").iterator():
+        if (normalized := normalize_duration(session.duration)) != session.duration:
+            session.duration = normalized
+            session.save(update_fields=["duration"])
+
+    category_model = apps.get_model("db_main", "ProposalCategory")
+    for category in category_model.objects.iterator():
+        normalized_durations = [
+            normalized
+            for duration in category.durations or []
+            if (normalized := normalize_duration(duration))
+        ]
+        if normalized_durations != category.durations:
+            category.durations = normalized_durations
+            category.save(update_fields=["durations"])
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [("db_main", "0141_eventpanelsettings_proposal_columns")]
+
+    operations = [
+        migrations.RunPython(normalize_stored_durations, migrations.RunPython.noop)
+    ]
--- a/src/ludamus/pacts/durations.py
+++ b/src/ludamus/pacts/durations.py
@@ -1,0 +1,36 @@
+"""The stored shape of a session length.
+
+`Session.duration` and `ProposalCategory.durations` hold ISO 8601 durations.
+That format is a storage detail every writer normalizes into and every reader
+leaves: it must not reach a screen, organizer's or participant's.
+"""
+
+import re
+
+MINUTES_PER_HOUR = 60
+_CANONICAL_DURATION_RE = re.compile(r"PT(?:(?P<hours>\d+)H)?(?:(?P<minutes>\d+)M)?")
+_DURATION_PART_RE = re.compile(r"(?P<hours>\d+)\s*h|(?P<minutes>\d+)\s*m")
+
+
+def parse_duration(iso_duration: str) -> tuple[int, int]:
+    if not (match := _CANONICAL_DURATION_RE.fullmatch(iso_duration or "")):
+        return 0, 0
+    return int(match["hours"] or 0), int(match["minutes"] or 0)
+
+
+def build_duration(*, hours: int, minutes: int) -> str:
+    if not hours and not minutes:
+        return ""
+    return "PT" + (f"{hours}H" if hours else "") + (f"{minutes}M" if minutes else "")
+
+
+def normalize_duration(text: str) -> str:
+    # Durations arrive in whatever shape their author felt like — "P4H",
+    # "50min", "110m". Read every hour and minute part out of the text and
+    # rebuild canonically; text carrying neither normalizes to "" (unset).
+    hours = minutes = 0
+    for part in _DURATION_PART_RE.finditer((text or "").lower()):
+        hours += int(part.group("hours") or 0)
+        minutes += int(part.group("minutes") or 0)
+    carried, minutes = divmod(minutes, MINUTES_PER_HOUR)
+    return build_duration(hours=hours + carried, minutes=minutes)
--- a/src/ludamus/templates/chronology/_session_card.html
+++ b/src/ludamus/templates/chronology/_session_card.html
@@ -62,10 +62,12 @@
                         {% icon "map-pin" class="size-4 shrink-0 text-foreground-muted/80" variant="mini" %}
                         <span class="truncate" title="{{ data.location_label }}">{{ data.location_label }}</span>
                     {% endif %}
-                    {% if data.session.duration %}
-                        {% icon "clock" class="size-4 shrink-0 text-foreground-muted/80" variant="mini" %}
-                        <span class="shrink-0">{{ data.session.duration|format_duration }}</span>
-                    {% endif %}
+                    {% with duration_label=data.session.duration|format_duration %}
+                        {% if duration_label %}
+                            {% icon "clock" class="size-4 shrink-0 text-foreground-muted/80" variant="mini" %}
+                            <span class="shrink-0">{{ duration_label }}</span>
+                        {% endif %}
+                    {% endwith %}
                 </div>
                 <div class="shrink-0 flex items-center gap-2">
                     {% if data.is_pending_proposal %}
--- a/src/ludamus/templates/panel/parts/import-recipe-row.html
+++ b/src/ludamus/templates/panel/parts/import-recipe-row.html
@@ -193,7 +193,7 @@
     <div class="recipe-durations mt-3 max-w-xl space-y-3 sm:ml-20 {% if row.selected != "session.duration" %}hidden{% endif %}"
          data-row="{{ row.index }}">
         <p class="text-xs text-foreground-muted">
-            {% translate "Map each source answer to an ISO 8601 duration (e.g. PT1H30M for 1 h 30 min). Answers without a mapping are skipped at import time." %}
+            {% translate "Set how long each source answer means. Answers left at zero are skipped at import time." %}
         </p>
         {% for dur in row.option_durations %}
             <div class="dr-option rounded-lg border border-border bg-bg-tertiary p-3">
@@ -201,14 +201,24 @@
                 <input type="hidden"
                        name="droption_{{ row.index }}"
                        value="{{ dur.option }}">
-                <input type="text"
-                       name="driso_{{ row.index }}"
-                       value="{{ dur.iso }}"
-                       aria-label="{% translate "ISO 8601 duration" %}"
-                       placeholder="PT1H30M"
-                       pattern="PT(\d+H)?(\d+M)?"
-                       title="PT1H30M"
-                       class="w-full rounded-lg border border-border bg-bg-secondary px-2 py-1 text-sm text-foreground focus:outline-none focus:ring-2 focus:ring-primary">
+                <div class="flex items-center gap-2">
+                    <input type="number"
+                           name="drhours_{{ row.index }}"
+                           value="{{ dur.hours|default_if_none:"" }}"
+                           min="0"
+                           max="24"
+                           aria-label="{% translate "Hours" %}"
+                           class="w-16 rounded-lg border border-border bg-bg-secondary px-2 py-1 text-center text-sm text-foreground focus:outline-none focus:ring-2 focus:ring-primary">
+                    <span class="text-sm text-foreground-muted">{% translate "h" %}</span>
+                    <input type="number"
+                           name="drminutes_{{ row.index }}"
+                           value="{{ dur.minutes|default_if_none:"" }}"
+                           min="0"
+                           max="59"
+                           aria-label="{% translate "Minutes" %}"
+                           class="w-16 rounded-lg border border-border bg-bg-secondary px-2 py-1 text-center text-sm text-foreground focus:outline-none focus:ring-2 focus:ring-primary">
+                    <span class="text-sm text-foreground-muted">{% translate "min" %}</span>
+                </div>
             </div>
         {% endfor %}
     </div>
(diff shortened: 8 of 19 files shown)


fancysnake (commented):
Thermo-nuclear code quality review. The intent is right — one canonical duration format, one place that knows it — but the PR stops halfway: it created `pacts/durations.py` and then left the renderer behind, scattered the "is this readable" check across templates (missing two of them), and shipped two functions in the same module that disagree on what canonical means.

7 action items posted inline, roughly in priority order.

src/ludamus/gates/web/django/templatetags/cfp_tags.py
```diff
@@ -154,28 +154,6 @@ def format_field_value(value: object) -> str:
...
-    Returns:
-        A string like "PT1H30M", "PT45M" or "".
-    """
-    if not hours and not minutes:
-        return ""
-    return "PT" + (f"{hours}H" if hours else "") + (f"{minutes}M" if minutes else "")
-
-
 @register.filter
 def format_duration(iso_duration: str) -> str:
```
fancysnake: **The canonical home was created and then only half-populated.**

This PR moved `parse_duration`/`build_duration` into `pacts/durations.py` — good — but left `format_duration` here. The result: `gates/web/django/forms.py` and `gates/web/django/chronology/forms.py` both `import ... from ludamus.gates.web.django.templatetags.cfp_tags` to get a pure `str -> str` function. Form-building code importing a Django templatetag module is exactly the boundary leak the new `pacts` module was supposed to end, and it's now *worse* than before the PR because the module docstring in `pacts/durations.py` claims ownership of "the stored shape of a session length" while the reader half lives elsewhere.

Move `format_duration` into `ludamus/pacts/durations.py` alongside `parse_duration`/`build_duration`, and leave a one-liner here:

```python
register.filter("format_duration", format_duration)
```

Then `pacts/durations.py` is the whole contract: parse, build, normalize, render. Nothing outside it needs to know ISO exists.

src/ludamus/templates/chronology/_session_card.html
```diff
@@ -62,10 +62,12 @@
                         {% icon "map-pin" class="size-4 shrink-0 text-foreground-muted/80" variant="mini" %}
                         <span class="truncate" title="{{ data.location_label }}">{{ data.location_label }}</span>
                     {% endif %}
-                    {% if data.session.duration %}
-                        {% icon "clock" class="size-4 shrink-0 text-foreground-muted/80" variant="mini" %}
-                        <span class="shrink-0">{{ data.session.duration|format_duration }}</span>
-                    {% endif %}
+                    {% with duration_label=data.session.duration|format_duration %}
```
fancysnake, line 71: **This guard is now in three templates and missing from two — push the label into the DTO instead.**

The `{% with label=x|format_duration %}{% if label %}` dance is repeated verbatim in `_session_card.html`, `panel/cfp-edit.html` and `panel/proposal-detail.html`. Two other call sites were not updated and still guard on the raw stored value:

- `templates/chronology/_compact_session_row.html` — `{% if data.session.duration %} · {{ ...|format_duration }}{% endif %}` renders an **orphaned `·` separator** when the raw value is truthy but unreadable.
- `templates/chronology/propose/parts/review.html` — same shape, renders an empty `Duration` `<dd>`.

That 3-of-5 hit rate is the argument against the approach, not an oversight to patch: every future template that touches a duration has to remember the incantation, and nothing catches it when it doesn't.

Code-judo: the DTO should carry the rendered label. Add `duration_label` next to `duration` on `SessionDTO` (and the proposal/review DTOs), computed once with `format_duration`. Then all five sites collapse to `{% if data.session.duration_label %}`, the double filter evaluation goes away, and there is no incantation left to forget.

If a DTO field is too heavy for the category-durations case in `cfp-edit.html`, the fallback is an inclusion tag — but the point stands: **one** place decides whether a duration is showable, not one per template.

src/ludamus/pacts/durations.py
```diff
@@ -0,0 +1,36 @@
...
+    return int(match["hours"] or 0), int(match["minutes"] or 0)
+
+
+def build_duration(*, hours: int, minutes: int) -> str:
+    if not hours and not minutes:
+        return ""
+    return "PT" + (f"{hours}H" if hours else "") + (f"{minutes}M" if minutes else "")
+
+
+def normalize_duration(text: str) -> str:
```
fancysnake, line 58: **`normalize_duration` is a lenient scraper that silently produces plausible-but-wrong numbers, and it runs on every write.**

`_DURATION_PART_RE` picks digit-then-h/m fragments out of *anywhere* in the string and sums them. Verified against this branch:

| input | result | should be |
|---|---|---|
| `"1.5h"` | `PT5H` | 1h30m, or reject |
| `"2h30"` | `PT2H` | 2h30m, or reject |
| `"P1DT2H"` | `PT2H` | 26h, or reject |
| `"5 maja"` | `PT5M` | reject |
| `"-5m"` | `PT5M` | reject |

`"2h30"` is a *common* human shorthand and it silently loses 30 minutes. `"1.5h"` is off by 3.5 hours. And this isn't confined to a one-shot salvage pass — it's wired into `ProposeSessionService.…` (`mills/legacy.py`), `_duration_iso` (`mills/submissions/mapping.py`) and `_settings_data` (`proposal_category_settings.py`), i.e. every write boundary. A tolerant guesser at a *write* boundary turns "unreadable, so unset" into "readable, so wrong", which is a worse failure than the one #341 is about: a wrong duration on screen is indistinguishable from a right one.

Two changes:

1. **Anchor it.** Require the pattern to consume the whole (stripped) string — something like `fullmatch` against `(\d+)\s*h(?:ours?)?)?\s*(?:(\d+)\s*m(?:in(?:utes?)?)?)?` plus the ISO form. Anything not fully consumed → `""`. That keeps `"P4H"`/`"50min"`/`"110m"` (the values the issue names, all covered by your tests) and turns `"1.5h"`, `"2h30"`, `"5 maja"` into unset instead of wrong.
2. **Scope it.** Legacy salvage belongs in the migration. Once the migration has run and `build_duration` is the only writer, `normalize_duration` on the `ProposeSessionService` / category-settings paths is defending against a shape those paths can no longer produce. Keep it where a genuinely foreign string arrives (the recipe `spec.iso` in `mapping.py`); drop it where the value already came from a stepper.

src/ludamus/pacts/durations.py
```diff
@@ -0,0 +1,36 @@
...
+_DURATION_PART_RE = re.compile(r"(?P<hours>\d+)\s*h|(?P<minutes>\d+)\s*m")
+
+
+def parse_duration(iso_duration: str) -> tuple[int, int]:
+    if not (match := _CANONICAL_DURATION_RE.fullmatch(iso_duration or "")):
+        return 0, 0
+    return int(match["hours"] or 0), int(match["minutes"] or 0)
+
+
+def build_duration(*, hours: int, minutes: int) -> str:
```
fancysnake, line 50: **Two functions in this module disagree on what "canonical" means, and the import editor can round-trip through the gap.**

`normalize_duration` carries minutes into hours (`divmod`, line 35). `build_duration` does not. So `build_duration(hours=0, minutes=90) == "PT90M"` — a value this module's own normalizer would rewrite.

That's reachable. In `google_docs_import._duration_values_from_post`, `_positive_int` accepts any digit string with no upper bound (the `max="59"` in `import-recipe-row.html` is client-side only), so posting 90 minutes writes `DurationSpec(iso="PT90M")` into `settings_json`. Re-opening the recipe then runs `_option_durations` → `parse_duration("PT90M")` → `(0, 90)` → renders `90` into an `<input type="number" max="59">`. The form re-opens in a state it refuses to accept.

Fix at the constructor, not the call sites:

```python
def build_duration(*, hours: int, minutes: int) -> str:
    carried, minutes = divmod(minutes, MINUTES_PER_HOUR)
    hours += carried
    ...
```

`normalize_duration` then just sums parts and delegates — its `divmod` line disappears — and **every** writer is canonical by construction, including the import path, with no extra `normalize_duration` call to remember.

Related, same area: the bounds are stated three times and already diverge. `forms.py` has `MAX_DURATION_HOURS = 23` / `MAX_DURATION_MINUTES = 59`; `import-recipe-row.html` hardcodes `max="24"` / `max="59"`; the server enforces neither. Use the constants in the template context and clamp server-side.

src/ludamus/gates/web/django/chronology/forms.py
```diff
@@ -91,8 +91,10 @@ def build_session_details_form(
         "display_name": forms.CharField(label=_("Presenter name"), max_length=255),
     }
 
-    if durations:
-        duration_choices = [(d, format_duration(d)) for d in durations]
+    # A duration nothing can read is left out: an option is worth offering only
+    # when it can be labelled with a length.
+    duration_choices = [(d, label) for d in durations if (label := format_duration(d))]
```
fancysnake: **Copy-pasted comprehension — same logic, same comment, two files.**

This line and `gates/web/django/forms.py` `_duration_field` are character-for-character the same list comprehension (`[(d, label) for d in durations if (label := format_duration(d))]`), carrying near-identical explanatory comments. Two copies of "which stored durations are worth offering" is one copy too many; the next rule added to that predicate will land in one of them.

Extract it next to `format_duration` in `pacts/durations.py`:

```python
def duration_choices(durations: Sequence[str]) -> list[tuple[str, str]]:
    return [(d, label) for d in durations if (label := format_duration(d))]
```

Both builders become `if choices := duration_choices(durations):` and the comment lives once.

src/ludamus/gates/web/django/chronology/panel/views/google_docs_import.py
```diff
@@ -79,7 +80,8 @@ class OptionEntity(TypedDict):
 
 class OptionDuration(TypedDict):
     option: str
-    iso: str
+    hours: int | None
```
fancysnake: **`int | None` here is optionality invented to spell "blank", and it costs a filter in the template.**

`parse_duration` returns plain `int`s. The view then converts `0 -> None` (`hours or None`) purely so `import-recipe-row.html` can render an empty input via `|default_if_none:""`. Nothing else distinguishes `None` from `0`, and `_duration_values_from_post` collapses them back together on the way in — so the nullability carries no information, it's a round-trip through the type system to avoid printing `"0"`.

Make the field say what the template needs:

```python
class OptionDuration(TypedDict):
    option: str
    hours: str
    minutes: str
```

```python
rows.append({
    "option": option,
    "hours": str(hours) if hours else "",
    "minutes": str(minutes) if minutes else "",
})
```

Template becomes `value="{{ dur.hours }}"` — the `default_if_none` filters go away, and `test_import_views.py` stops asserting `{"hours": None, "minutes": None}` for what is really "blank".

While here: `_positive_int` (line 520) doesn't check positivity and doesn't reject — `"abc"` and `"-5"` both become `0` silently. Name it for what it does (`_digits_or_zero`) or make it clamp against `MAX_DURATION_HOURS`/`MAX_DURATION_MINUTES` and actually be the validation the client-side `max` pretends to be.

src/ludamus/links/db/django/migrations/0142_normalize_durations.py
```diff
@@ -0,0 +1,37 @@
+from django.db import migrations
+
+from ludamus.pacts.durations import normalize_duration
```
fancysnake: **A data migration importing live application code is a time bomb.**

`from ludamus.pacts.durations import normalize_duration` means this migration's behaviour is whatever `normalize_duration` happens to do *at the time it runs*. The moment the parsing rules are tightened (see the comment on `normalize_duration` — they should be), this historical migration silently starts producing different data on any environment that hasn't run it yet, and the branch's own tests can no longer tell you what it did. Migrations are frozen snapshots; that's why `apps.get_model` exists two lines below, and the same reasoning applies to the logic, not just the schema.

Inline the normalizer (and its regex) into this file. It's ~10 lines, and it makes the migration reproducible forever.

Two smaller things in the same function:

- The soft-delete comment is correct but for a reason nobody reading it can see: `apps.get_model` strips custom managers unless `use_in_migrations = True`, so `session_model.objects` is a plain `Manager` and *does* include deleted rows — the opposite of what `Session.objects` means everywhere else in the codebase. Say that explicitly, or the comment reads as a claim the code doesn't back.
- `.iterator()` with a per-row `save()` is one UPDATE per session. `bulk_update` in batches over the rows that actually changed is the same code length and won't be a problem if this table is large on prod.

src/ludamus/gates/web/django/templatetags/cfp_tags.py
fancysnake: format_duration now lives in `pacts/durations.py` alongside parse/build/normalize, plus `duration_choices` (which both form builders were duplicating — the other thread). `cfp_tags` is a registration line, and neither forms module imports a templatetag any more. Its unit tests moved to `tests/unit/test_pacts_durations.py`.

src/ludamus/gates/web/django/chronology/forms.py
fancysnake: Done as part of the `format_duration` move — `duration_choices(durations)` lives next to it in `pacts/durations.py`, both builders call it, the comment lives once.

src/ludamus/pacts/durations.py
fancysnake: Anchored. `_LOOSE_DURATION_RE` is now `fullmatch`ed against the stripped, lowercased string, so anything the pattern cannot consume whole normalizes to `""`. Every case you listed is covered by a new test — `1.5h`, `2h30`, `P1DT2H`, `5 maja`, `-5m`, plus `1h 2h` — and `P4H` / `50min` / `110m` / `110min` still land where they did.

The scoping half is deferred to #821: the category-settings path still accepts hidden inputs that predate the steppers, so the belt stays until migration 0142 has run everywhere.

src/ludamus/pacts/durations.py
fancysnake: Fixed at the constructor, as suggested — `build_duration` carries with `divmod` before composing, so `build_duration(hours=0, minutes=90) == "PT1H30M"` and `normalize_duration` just sums parts and delegates. The import round-trip you traced can no longer produce a value its own editor refuses.

The bounds: `MAX_DURATION_HOURS` / `MAX_DURATION_MINUTES` moved to `pacts/durations.py` (one home, and gates may only import pacts), the recipe row renders `max` from them via context instead of hardcoding 24/59, and `_positive_int` became `_bounded_int`, which caps server-side. New test posts `99h`/`90m`/`-5`/`abc` and asserts `PT23H59M` with the junk row dropped.

src/ludamus/links/db/django/migrations/0142_normalize_durations.py
fancysnake: Inlined. The migration carries a frozen copy of the normalizer and its regex, with a comment saying why — which matters more now that the live one has been tightened in this same push.

The soft-delete comment now says the actual reason: `apps.get_model` drops `AliveManager` because it sets no `use_in_migrations`, so `objects` here is a plain manager that sees every row.

Left the per-row `save()`: production has 13 affected rows across 5 distinct values (PR body), and the loop only writes rows that changed, so `bulk_update` would be optimising a one-shot 13-row pass.

src/ludamus/templates/chronology/_session_card.html
fancysnake: Both missed templates now guard on the label — `_compact_session_row.html` (the orphaned `·`) and `propose/parts/review.html` (the empty `<dd>`).

The DTO proposal is right and is filed as #820 rather than done here: it touches `SessionDTO` plus the proposal/review DTOs, five templates and their tests, and `cfp-edit.html` iterates a category's configured durations with no DTO to hang a label on, so it needs its own answer. Worth noting the hit rate overstates the risk: after migration 0142 every stored value is canonical and `build_duration`/`normalize_duration` are the only writers, so `format_duration` cannot return `""` for a non-empty stored value. The guards are belt-and-braces, not a live bug.

src/ludamus/gates/web/django/chronology/panel/views/google_docs_import.py
fancysnake: Split this one.

Took the `_positive_int` half — it is now `_bounded_int(raw, maximum)`, capping against `MAX_DURATION_HOURS` / `MAX_DURATION_MINUTES` (moved to `pacts/durations.py`), so the client-side `max` is backed by a server-side one. The template renders `max` from those constants instead of the divergent 24/59.

Took the `int | None` half too, though for the template rather than the type: `OptionDuration.hours` / `minutes` are `str` now, the `default_if_none` filters are gone, and the test asserts `""` instead of `None`.

src/ludamus/templates/chronology/_session_card.html
```diff
@@ -62,10 +62,12 @@
...
+                        {% if duration_label %}
```
fancysnake, line 72: ## Thermo-nuclear code quality review

**Either these guards are load-bearing or they are dead — the PR is currently arguing both, and the answer decides whether five templates change at all.**

The reply on the earlier thread makes the case that after migration 0142 `build_duration`/`normalize_duration` are the only writers, so `format_duration` cannot return `""` for a non-empty stored value. Take that seriously and follow it through: under that invariant

```
{% with duration_label=x|format_duration %}{% if duration_label %} … {% endif %}{% endwith %}
```

is *exactly equivalent* to the one-liner it replaced, `{% if x %} … {{ x|format_duration }}{% endif %}`. Empty in, both falsy; canonical in, both truthy. So this PR added a three-line `{% with %}/`{% if %}` wrapper to five templates (`_session_card.html`, `_compact_session_row.html`, `propose/parts/review.html`, `panel/proposal-detail.html`, `panel/cfp-edit.html`) that changes nothing, and filed #820 to add a DTO field to make the same nothing tidier.

That is complexity spread rather than deleted. Two coherent endings, pick one:

- **The invariant holds.** Revert all five templates to the plain `{% if x %}` guard and drop #820. `format_duration` returning `""` for junk is then a defensive detail nobody has to render around, and the PR gets smaller. This is the deletion, and it is the one I'd take.
- **The invariant does not hold** (some writer you haven't found, a fixture, a restored row, a future import path). Then the guard is real — and #820 is not deferrable, because the argument on the earlier thread was precisely that 2 of 5 call sites already forgot the incantation and nothing catches the next one.

What can't stand is the middle: keeping the ceremony because it might matter while telling the reviewer it can't.

`panel/cfp-edit.html` is the sharpest case, and it decides this on its own. Its guard is not cosmetic — the comment there says so: *"An unreadable stored duration is left out entirely, so saving drops it."* Rendering silently omits the hidden `<input name="durations">`, so the operator opening the CfP page and pressing Save loses a configured duration they were never shown and never touched. If the invariant holds, that branch is unreachable and should go. If it doesn't, destroying data as a side effect of a render filter is the wrong response to it — the migration is where unreadable values get resolved, visibly, once.

src/ludamus/gates/web/django/chronology/panel/views/google_docs_import.py
```diff
@@ -493,20 +510,34 @@ def _time_slot_values_from_post(
...
-        values[option] = DurationSpec(iso=clean_iso)
+        values[option] = DurationSpec(iso=iso)
     return values
 
 
+def _bounded_int(raw: str, maximum: int) -> int:
+    # The steppers' `max` is client-side only, so the bound is enforced here
+    # too; anything that is not a plain number reads as unset.
+    text = (raw or "").strip()
+    return min(int(text), maximum) if text.isdigit() else 0
```
fancysnake: ## Thermo-nuclear code quality review

**One bound, two policies, both written in this PR — and the silent one is the write path.**

`MAX_DURATION_HOURS` / `MAX_DURATION_MINUTES` are now enforced twice, in opposite ways:

- `SessionEditForm.duration_hours` — `forms.IntegerField(max_value=MAX_DURATION_HOURS)`. Out of range → validation error, form re-renders, the organizer is told.
- `_bounded_int` — `min(int(text), maximum)`. Out of range → clamped, saved, nobody told.

So an operator who types `99` into the hours stepper on a recipe row gets `PT23H` written into `settings_json` and a green *"Question saved."*. `-5` and `abc` are worse: `isdigit()` is false, `_bounded_int` returns `0`, `build_duration` returns `""`, and the `if not (option and iso): continue` two lines above **drops the option from the mapping entirely** — which at import time means every source row whose answer hits that option is silently skipped. A typo in one stepper quietly removes rows from an import. `tests/integration/web/panel/test_import_views.py` pins exactly this (`"drhours_0": ["99", "-5"]` → `PT23H30M`, second option absent), so it's the intended contract, not an accident.

This contradicts the PR's own premise. The argument for anchoring `normalize_duration` was that a plausible-but-wrong length is worse than an unset one, because a wrong duration on screen is indistinguishable from a right one. `min(99, 23)` is precisely that failure: `PT23H` is indistinguishable from a deliberate 23 hours.

The earlier thread offered *rename it or clamp it*; clamping was the wrong half of that. This is a recipe row with an existing per-row error path (`messages.error` + redirect in `EventImportRowSaveView`, already used for `"Invalid row submission."`). Reject out-of-range and non-numeric input the way every other recipe field is rejected, and let `_bounded_int` disappear:

```python
def _duration_from_post(hours: str, minutes: str) -> str:  # raises on junk
    ...
```

If a full form is too much for this view right now, the minimum is that clamping and dropping stop being silent — the operator has to see that what they typed is not what got stored.

src/ludamus/gates/web/django/chronology/panel/views/google_docs_import.py
```diff
@@ -493,20 +510,34 @@ def _time_slot_values_from_post(
...
+            minutes=_bounded_int(minutes, MAX_DURATION_MINUTES),
+        )
+        if not (option and iso):
             continue
-        values[option] = DurationSpec(iso=clean_iso)
+        values[option] = DurationSpec(iso=iso)
     return values
 
 
+def _bounded_int(raw: str, maximum: int) -> int:
```
fancysnake: ## Thermo-nuclear code quality review

**`int(text)` on unbounded operator input is a 500.**

Since Python 3.11 `int(str)` refuses strings longer than `sys.get_int_max_str_digits()` (4300 by default):

```
>>> int('9' * 4301)
ValueError: Exceeds the limit (4300 digits) for integer string conversion: value has 4301 digits
```

`text.isdigit()` is happily `True` for that string, so the guard passes and `int(text)` raises. `_duration_values_from_post` has no handler, so `POST` to the recipe-save view with a 4301-digit `drhours_0` is an unhandled `ValueError` — a 500 out of a panel form.

The reason it's worth fixing rather than shrugging at is that it's the same root cause as the clamping thread: `_bounded_int` tries to be total ("anything unparsable reads as unset") over an input domain it hasn't actually bounded. `isdigit()` checks shape; nothing checks size. Length is the missing half:

```python
text = (raw or "").strip()
if not text.isdigit() or len(text) > len(str(maximum)):
    return 0
return min(int(text), maximum)
```

If this instead becomes real field validation (the other thread), `forms.IntegerField` already handles it and the helper goes away — which is the better ending.

src/ludamus/gates/web/django/forms.py
```diff
@@ -650,23 +660,22 @@ class SessionEditForm(forms.Form):
         help_text=_("Empty or 0 = no limit"),
     )
     min_age = forms.IntegerField(required=False, min_value=0, label=_("Minimum Age"))
-    duration = forms.CharField(required=False, label=_("Duration"))
+    # Not a field: the name a subclass's picker takes. Declared so the shared
+    # duration partial can ask whether there is one to render.
+    duration = None
```
fancysnake, line 660: ## Thermo-nuclear code quality review

**`duration = None` does nothing, and the comment defending it states something Django does not do.**

> *Declared so the shared duration partial can ask whether there is one to render.*

The partial can already ask, with or without this line. `{% if %}` compiles its operands to `TemplateLiteral`, whose `eval` is `self.value.resolve(context, ignore_failures=True)` (`django/template/defaulttags.py`), and `FilterExpression.resolve` with `ignore_failures=True` returns `None` on `VariableDoesNotExist` (`django/template/base.py`). So `{% if form.duration %}` on a form with no `duration` field is already `False`. Delete the attribute and the two comment lines and `proposal-duration.html` renders identically.

Worse, the attribute is not inert in Python: `SessionEditForm.duration` is now `None` where every other name on this class is a `Field`, so `getattr(form, "duration")` and `getattr(form, "title")` return objects of unrelated kinds. That is a shape trap for the next reader for no gain.

But the deletion is the small half. **The real issue is that one question is spelled five different ways across two layers.** "Did a subclass add a picker?" is asked as:

- `duration = None` + `{% if form.duration %}` — three times in `proposal-duration.html` (lines 8, 11, 12), each relying on the dict-lookup-then-attribute-fallback path above;
- `"duration" in self.fields` — twice in `SessionEditForm.clean()`, once to decide whether to compose and once to decide whether to error.

Five spellings, two of them load-bearing on Django template internals. Name it once, on the form that owns it:

```python
@property
def has_duration_picker(self) -> bool:
    return "duration" in self.fields
```

`clean()` reads it twice by name, the partial reads `{% if form.has_duration_picker %}`, and `duration = None` disappears with nothing to replace it. `{% tessera_field form.duration %}` still works — it's inside the guarded branch.

That also exposes the shape underneath, which is worth a second look while you're here: `SessionEditForm.clean()` branches on `CUSTOM_DURATION`, a constant only meaningful to a field the base class does not declare. The base is written around its subclass. Collapsing `_ComposedDurationForm` into `SessionEditForm` was the right instinct, but the two `"duration" in self.fields` checks and the sentinel are the old class's boundary paid for a second time in conditionals. A `_DurationPickerMixin` that `create_proposal_form` mixes in *only when `_duration_field` returns a field* puts `CUSTOM_DURATION` next to the field that uses it and leaves `SessionEditForm.clean()` as three unconditional lines.

src/ludamus/links/db/django/migrations/0143_normalize_durations.py
```diff
@@ -0,0 +1,60 @@
...
+    # anything unreadable becomes unset rather than shown raw.
+    session_model = apps.get_model("db_main", "Session")
+    # Soft-deleted rows included: `apps.get_model` drops the `AliveManager`
+    # (it sets no `use_in_migrations`), so `objects` here is a plain manager
+    # that sees every row — unlike `Session.objects` everywhere else. That is
+    # what we want: restoring a session must not bring an unreadable duration
+    # back with it.
+    for session in session_model.objects.exclude(duration="").iterator():
+        if (normalized := _normalize(session.duration)) != session.duration:
+            session.duration = normalized
```
fancysnake, line 54: ## Thermo-nuclear code quality review

**This line destroys data and leaves no record of what it destroyed.**

`_normalize` returns `""` for anything the anchored pattern can't consume whole — deliberately, and I agree with the choice. But the consequence is that `session.duration = normalized` overwrites the only copy of the original string, `RunPython.noop` guarantees it can't be recovered by rolling back, and the run prints nothing. After deploy, "which sessions lost their length, and what did they say?" is unanswerable, and the affected facilitator gets no length on their session with no way for support to tell them what it used to be.

The reply on the earlier thread says prod has 13 affected rows across 5 distinct values, which is the argument *for* logging, not against it — 13 lines is a free audit trail:

```python
logger = logging.getLogger(__name__)
...
if (normalized := _normalize(session.duration)) != session.duration:
    logger.info(
        "0142: session %s duration %r -> %r", session.pk, session.duration, normalized
    )
```

This is the codebase's own bar — *"new path logs meaningful events"* in the definition of done — and a one-shot irreversible rewrite is the path that most needs it. Same for the `ProposalCategory` loop below, where an unreadable entry vanishes from a list and the remaining entries shift.

If you'd rather not log per row, a summary count at the end (`n sessions blanked, n rewritten`) is the minimum, so the deploy output can be compared against the 13 you expect. Right now a run that touched 1300 rows and a run that touched 0 look identical.

src/ludamus/templates/chronology/parts/session-edit-form.html
```diff
@@ -97,18 +97,7 @@
...
-                <input type="text"
-                       name="duration"
-                       id="id_duration"
-                       value="{{ form.duration.value|default:'' }}"
-                       class="w-full border {% if form.duration.errors %}border-danger{% else %}border-border{% endif %} rounded-lg px-4 py-2 focus:outline-none focus:ring-2 focus:ring-primary">
-                {% if form.duration.errors %}
-                    <p class="mt-1 text-sm text-danger">{{ form.duration.errors.0 }}</p>
-                {% endif %}
-            </div>
+            <div>{% include "panel/parts/proposal-duration.html" %}</div>
```
fancysnake, line 100: ## Thermo-nuclear code quality review

**A participant-facing template now reaches into `panel/parts/` for a shared control.**

`panel/parts/proposal-duration.html` has two consumers as of this PR: `panel/parts/proposal-session-fields.html` (organizer panel) and this file, the facilitator's own session-edit modal under `chronology/`. The moment the second consumer appeared, the partial stopped being a panel detail — but it kept the panel's address, and now the directory says "organizer panel" while the code says "anyone editing a session length".

That matters beyond tidiness: the next person changing something in `panel/parts/` has no reason to expect the participant-facing modal to move with it, and the filename `proposal-duration` is wrong for half its callers (a session being self-edited is not a proposal).

CLAUDE.md already names the home for this — *"Partials in `templates/components/`"*. Move it to `templates/components/duration-field.html`, update the two includes and the `django.po` reference. Pure rename, and the include in this file reads as what it is: a shared field, not a borrowed panel fragment.

While moving it: the wrapping `<div>` here exists only to give the include a block box, and the partial already opens with `<div class="group">`. Drop the wrapper and let the include stand alone.

hasparus:
## Stale-PR sweep

Flagged: no commits in 2+ days, `checks` and `tingle` failing on `9f11218`.

**`checks` — fixed.** `messages-check` was failing on a stale PL catalog (drifted `#, python-brace-format` flags + `POT-Creation-Date`, unrelated to this PR's own edits). Regenerated via `mise run messages`.

**Merge conflict — resolved.** `main` had moved on (30+ commits); merged it in. Two real conflicts:
- `templates/panel/proposal-detail.html` — `main` independently re-indented this template (wrapped in `tab_shell_body`, added a new "scheduled" card) since this PR's base. Took `main`'s structure and re-applied this PR's only actual change to the file: the `duration_label` `{% with %}` guard.
- `django.po` — catalog drift (header timestamp, obsolete-message reordering); regenerated after resolving.

**Migration collision — fixed.** `main` had independently landed `0142_guild_guildmembership` off the same `0141` parent this PR's `0142_normalize_durations` used, so Django's test-db setup failed with "Conflicting migrations detected: multiple leaf nodes". Renumbered this branch's migration to `0143_normalize_durations`, chained after `main`'s `0142` — no operations changed, dependency-only fix.

**`tingle` — left red, as the PR description already explains.** Net +39 (`legacy-loc`/`old-subdomain-loc`, both "no located additions") comes from tests landing in `tests/integration/web/chronology/` and `.../panel/`, which the noun/verb migration still counts as legacy. Already called out in the PR body as pre-existing before this push — not something this pass should force green.

**Review comments — not in scope for this pass.** The unresolved threads are from `fancysnake`'s thermo-nuclear review (architectural: DTO field for `duration_label`, a template move to `components/`, migration audit logging, a form/mixin refactor) — real feedback, but outside the "CodeRabbit and Rams" comments this automated sweep addresses, and each one is a design call rather than a mechanical fix. Leaving them for a normal review pass. No CodeRabbit or Rams comments are outstanding (CodeRabbit hit its OSS review-rate limit on this PR and never actually reviewed it).

**Verification:** full `mise run test:py` — **4024 passed, 9 skipped**, twice, clean both times.

Pushed as `0090f74` (catalog) → `e6566eba` (merge + migration renumber). Not merged — left ready for review.

---

src/ludamus/templates/panel/parts/import-recipe-row.html
```diff
@@ -193,22 +193,32 @@
...
+                    <span class="text-sm text-foreground-muted">{% translate "h" %}</span>
+                    <input type="number"
+                           name="drminutes_{{ row.index }}"
+                           value="{{ dur.minutes }}"
+                           min="0"
+                           max="{{ max_duration_minutes }}"
+                           aria-label="{% translate "Minutes" %}"
+                           class="w-16 rounded-lg border border-border bg-bg-secondary px-2 py-1 text-center text-sm text-foreground focus:outline-none focus:ring-2 focus:ring-primary">
+                    <span class="text-sm text-foreground-muted">{% translate "min" %}</span>
+                </div>
```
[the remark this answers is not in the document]
hasparus: Done: each option is a `<fieldset>` with `dur.option` as the legend, and the hours/minutes inputs are the shared `stepper-field` (with Hours/Minutes `aria-label`s on the inputs). Closes #834.

tests/integration/utils.py
```diff
@@ -51,6 +51,23 @@ def __repr__(self) -> str:
...
+    # business, not the assertion's.
+    def __init__(self, **initial: object) -> None:
+        self.initial = initial
+
+    def __eq__(self, other: object) -> bool:
+        actual = getattr(other, "initial", {})
+        return {key: actual.get(key) for key in self.initial} == self.initial
+
+    def __hash__(self) -> int:
+        return hash(frozenset(self.initial.items()))
```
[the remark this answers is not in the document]
hasparus: `__hash__` is now `None`. Subset equality stays; the matcher is explicitly unhashable so ruff's `eq-without-hash` is happy too.

PR #817, opened 2026-08-08
perf(rituals): hand agents the tail of a gate log, not the transcript

fancysnake:
pr_check fed every line of `mise run pr-fix` and `mise run diff-cover` into a coding call — thousands of lines of passing tests, e2e output and coverage tables.

A mise task list stops at the first failure and every tool puts its verdict last, so `said` now keeps the end of what a gate wrote. The budget is **4000 characters per stream**, not a line count: what is being spent is tokens, and a line has no fixed price — twenty lines of ruff is a couple of hundred bytes, twenty carrying a pytest assertion repr is orders of magnitude more.

Per stream, and not on the two joined together. mise puts its own task chatter on stderr and the tool's verdict on stdout — `mise run lint` alone is 35 lines of stdout against 125 of stderr — so a shared budget is won by whichever stream is longer, which is the chatter. A repair agent would have been handed progress banners and no diagnosis at all.

`cover` is not a caller that wants a tail. It routes on the output (`if "Missing lines" in output`) and hands it to the agent as the work list, so a trimmed listing is an agent asked to cover lines it was never shown — and then another full unit and e2e run to reveal the next few. diff-cover's report is self-delimiting, so `coverage_report` cuts at its own `Diff Coverage` banner instead: the listing entire however many files it names, the suite transcript dropped whole. Where the run died before printing a report, the trimmed log stands in.

Routing reads `Missing lines` and not `Missing`, which diff-cover's summary block says on a clean report too (`Missing: 0 lines`) — that one is a pre-existing bug, fixed here because this change touches the same line.

### DIFF
--- a/src/ludamus/edges/rituals/shell.py
+++ b/src/ludamus/edges/rituals/shell.py
@@ -38,31 +38,68 @@ WAIT_LABEL = "pr::wait"
 
 STASHED = "stashed"
 
-# How much of a gate's output is worth paying for. A mise task list stops at the
-# first failure, and every tool in it puts its verdict last — pytest's short
-# summary, diff-cover's missing lines — so the tail is the part that says what
-# is wrong. What sits above it is a suite naming three and a half thousand tests
-# that passed, and an agent told to re-run one failing test by name does not
-# need to read them.
-TAIL = 20
+# How much of a gate's output is worth paying for, counted in characters rather
+# than lines: what is being spent is tokens, and a line has no fixed price —
+# twenty lines of ruff is a couple of hundred bytes, and twenty carrying a
+# pytest assertion repr or a mypy note about a long generic type is orders of
+# magnitude more. A mise task list stops at the first failure, and every tool in
+# it puts its verdict last — pytest's short summary, ruff's count — so what fits
+# at the end is the part that says what is wrong. What sits above it is a suite
+# naming three and a half thousand tests that passed, and an agent told to
+# re-run one failing test by name does not need to read them.
+BUDGET = 4000
+
+# diff-cover's report is self-delimiting — its own banner, then everything to
+# the end of the run — so the one caller that reads the report needs no budget
+# at all. See `coverage_report`.
+BANNER = "Diff Coverage"
 
 
 def quoted(value: str) -> str:
     return shlex.quote(value)
 
 
+# The budget buys whole lines, and the last line is bought whether it fits or
+# not: a tool that pretty-prints a value onto one long line would otherwise
+# spend the budget and hand back nothing.
+def _tail(text: str) -> str:
+    lines = text.splitlines()
+    kept: list[str] = []
+    left = BUDGET
+    for line in reversed(lines):
+        left -= len(line) + 1
+        if left < 0 and kept:
+            break
+        kept.append(line)
+    if len(kept) == len(lines):
+        return "\n".join(lines)
+    kept.reverse()
+    return "\n".join([f"[{len(lines) - len(kept)} earlier lines omitted]", *kept])
+
+
 # Both streams: a task that dies before it starts says so on stderr and nowhere
 # else, and an empty complaint is the one thing a repair agent cannot work with.
+# Each stream gets the budget on its own rather than sharing one, because mise
+# puts its own task chatter on stderr and the tool's verdict on stdout, and a
+# shared budget is won by whichever stream is longer — which is the chatter.
 # Trimmed here rather than at each prompt, because every caller does the same
 # two things with this — hand it to an agent, or put it in the report — and both
 # want the verdict, not the transcript.
 def said(result: ShellResult) -> str:
-    joined = "\n".join(part for part in (result.stdout, result.stderr) if part.strip())
-    lines = joined.split("\n")
-    if len(lines) <= TAIL:
-        return joined
-    dropped = len(lines) - TAIL
-    return "\n".join([f"[{dropped} earlier lines omitted]", *lines[-TAIL:]])
+    return "\n".join(
+        _tail(part) for part in (result.stdout, result.stderr) if part.strip()
+    )
+
+
+# Cutting at diff-cover's own banner drops the suite transcript above it whole
+# and hands on the listing entire, however many files it names. A budget has no
+# business here: this listing is not something an agent reads for a verdict, it
+# is the work list, and a tail of it is an agent asked to cover lines it was
+# never shown — then another full unit and e2e run to reveal the rest. Only
+# where the report never got printed does the trimmed log stand in for it.
+def coverage_report(result: ShellResult) -> str:
+    _, banner, rest = result.stdout.partition(BANNER)
+    return banner + rest if banner else said(result)
 
 
 def commit(message: str) -> str:
--- a/src/ludamus/edges/rituals/shell.py
+++ b/src/ludamus/edges/rituals/shell.py
@@ -38,6 +38,14 @@ WAIT_LABEL = "pr::wait"
 
 STASHED = "stashed"
 
+# How much of a gate's output is worth paying for. A mise task list stops at the
+# first failure, and every tool in it puts its verdict last — pytest's short
+# summary, diff-cover's missing lines — so the tail is the part that says what
+# is wrong. What sits above it is a suite naming three and a half thousand tests
+# that passed, and an agent told to re-run one failing test by name does not
+# need to read them.
+TAIL = 20
+
 
 def quoted(value: str) -> str:
     return shlex.quote(value)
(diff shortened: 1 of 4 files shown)


src/ludamus/edges/rituals/shell.py
```diff
@@ -38,15 +38,31 @@
...
+# two things with this — hand it to an agent, or put it in the report — and both
+# want the verdict, not the transcript.
 def said(result: ShellResult) -> str:
-    return "\n".join(part for part in (result.stdout, result.stderr) if part.strip())
+    joined = "\n".join(part for part in (result.stdout, result.stderr) if part.strip())
+    lines = joined.split("\n")
+    if len(lines) <= TAIL:
+        return joined
+    dropped = len(lines) - TAIL
+    return "\n".join([f"[{dropped} earlier lines omitted]", *lines[-TAIL:]])
```
fancysnake: **This is a behaviour regression in the `cover` loop, not just a trim.**

The comment above says every caller does one of two things with this — hand it to an agent or put it in the report. `cover` does a third: it *parses* it (`if "Missing" in output`) and then hands the whole thing to the agent as the work list (`COVER + output`).

`mise run diff-cover` is the full pytest run + e2e + combine + the report, so tail-20 of it is the tail of diff-cover's report. That report's trailing block is already ~8 lines on its own (`-------------`, `Total:`, blank, `Missing:`, blank, `Coverage:`, `-------------`, blank — see `diff_cover/templates/console_coverage_report.txt`), leaving ~11 lines for the per-file `src/x.py (80.0%): Missing lines 12-14` listing. A branch touching more than a dozen files loses most of that listing, and the agent is asked to cover lines it was never shown. Then the step loops — and each round re-runs the entire unit + e2e suite to reveal the next eleven. The perf change makes the most expensive step in the ritual run more times.

The judo move is to put the trim where the tokens are actually paid for instead of in the shared join:

- `said` goes back to being the pure both-streams join.
- `gate_check` trims at the prompt: `fix_gates(tail(said(gates)))`.
- `cover` doesn't need a line count at all — the report is self-delimiting. Cut on its own `Diff Coverage` header and you get the exact report, whole, with the suite transcript that motivated this PR dropped entirely. That is strictly better than a tail for this caller and removes the guesswork.

That also deletes the thing this PR is currently betting on: that 20 lines happens to be enough for every consumer.

src/ludamus/edges/rituals/shell.py
```diff
@@ -38,15 +38,31 @@
...
 
 
 # Both streams: a task that dies before it starts says so on stderr and nowhere
 # else, and an empty complaint is the one thing a repair agent cannot work with.
+# Trimmed here rather than at each prompt, because every caller does the same
+# two things with this — hand it to an agent, or put it in the report — and both
+# want the verdict, not the transcript.
 def said(result: ShellResult) -> str:
-    return "\n".join(part for part in (result.stdout, result.stderr) if part.strip())
+    joined = "\n".join(part for part in (result.stdout, result.stderr) if part.strip())
```
fancysnake: Trimming the *joined* string rather than each stream quietly defeats the invariant this function exists to hold. The comment directly above still claims it: "Both streams: a task that dies before it starts says so on stderr and nowhere else, and an empty complaint is the one thing a repair agent cannot work with."

After this change, a stderr longer than `TAIL` evicts stdout completely — the tail is all stderr, and whatever verdict the tool wrote to stdout is gone. `mise run pr-fix` is `mise install` + `poetry install` + format + messages + lint + varlock; which of those writes where is not something this helper should have to bet on.

If both streams matter, tail each stream and then join. Same cost ceiling, and the guarantee in the comment stays true.

src/ludamus/edges/rituals/shell.py
```diff
@@ -38,15 +38,31 @@
 
 STASHED = "stashed"
 
+# How much of a gate's output is worth paying for. A mise task list stops at the
+# first failure, and every tool in it puts its verdict last — pytest's short
+# summary, diff-cover's missing lines — so the tail is the part that says what
+# is wrong. What sits above it is a suite naming three and a half thousand tests
+# that passed, and an agent told to re-run one failing test by name does not
+# need to read them.
+TAIL = 20
```
fancysnake: Minor, but the contract is fuzzier than the comment claims. "How much of a gate's output is worth paying for" is a cost statement, and a line count doesn't bound cost. Twenty lines of ruff is a couple hundred bytes; twenty lines carrying a pytest assertion repr or a mypy note about a long generic type is orders of magnitude more. If the budget is tokens, budget characters and let the line count fall out of that.

tests/unit/rituals/test_shell.py
```diff
@@ -0,0 +1,26 @@
...
+
+    def test_a_long_log_keeps_its_tail_and_says_what_it_dropped(self) -> None:
+        run = "\n".join([f"tests/test_{index}.py PASSED" for index in range(TAIL + 50)])
+
+        trimmed = said(_ran(stdout=run, stderr="1 failed"))
+
+        assert trimmed.startswith("[51 earlier lines omitted]\n")
+        assert trimmed.endswith("1 failed")
+        assert "tests/test_0.py PASSED" not in trimmed
+        assert len(trimmed.split("\n")) == TAIL + 1
```
fancysnake: These two cases pin the helper but not the consequence. The trim lives inside `said`, so what it actually changes is what `cover` routes on and what the COVER agent receives — and `tests/unit/rituals/test_pr_check_gates.py` already has the trial harness and a `_MISSING` fixture line for exactly that shape.

Add the case there: a diff-cover report naming more files than `TAIL` still reaches the agent whole. Right now that test would fail, which is the point.

Smaller: the only long-log case is `TAIL + 50`. The interesting boundary is exactly `TAIL` lines (returns untouched) and `TAIL + 1` (returns the header plus the tail) — that's where an off-by-one would live.

fancysnake (commented):
Thermo-nuclear code quality review.

The idea is right — a gate's verdict is at the tail and the transcript above it is money spent on nothing. The placement is not. Putting the trim inside `said` applies it to a caller that parses the string and to a caller that needs the whole of it, and the comment justifying the placement states an invariant about the callers that isn't true. One structural regression, one contract question, two smaller items — inline.

src/ludamus/edges/rituals/shell.py
fancysnake: Fixed in 9d8692b. Confirmed empirically before changing anything — `mise run lint` on this machine is 35 lines of stdout against 125 of stderr, because mise puts its own `[task] $ …` / `✔` chatter on stderr and the tool output on stdout. So the tail of the join was 100% progress banners; `pr-fix` is a superset of `lint`, so worse. `said` now budgets each stream on its own and joins after, and `test_a_long_stderr_cannot_evict_what_stdout_said` pins it (verified it fails against the shared-budget version).

src/ludamus/edges/rituals/shell.py
fancysnake: Fixed in 9d8692b, taking your route rather than a bigger tail. `coverage_report` cuts `result.stdout` at diff-cover's own `Diff Coverage` banner: the listing goes to the agent entire however many files it names, and the suite transcript is dropped whole. `said` stands in only where the run died before printing a report, which is the `failed without naming missing lines` path.

Worth noting it was worse than the truncation you described. With the tail landing in stderr (see the sibling thread), the report was gone entirely, `"Missing" in output` was False, and the step fell through to `set_aside`. The cover loop was dead, not degraded.

Also fixed while in there: routing read `Missing`, which diff-cover's summary block prints on a clean report too (`Missing: 0 lines`), so every green run went round the loop again. It reads `Missing lines` now — pre-existing, but the same line.

src/ludamus/edges/rituals/shell.py
fancysnake: Fixed in 9d8692b. `TAIL = 20` is now `BUDGET = 4000` characters, per stream, and the line count falls out of it as you said. Comment carries the reasoning.

tests/unit/rituals/test_shell.py
fancysnake: Fixed in 9d8692b. `test_every_file_reaches_the_agent_out_from_under_the_suite` is in `test_pr_check_gates.py` on the existing trial harness, with a listing deliberately longer than `BUDGET` on its own — first cut of it fit inside the budget by accident and passed against a mutant, so it is tied to `BUDGET` now. Boundaries are in `test_shell.py`: exactly the budget comes back whole, one line over returns the header plus the tail.

src/ludamus/edges/rituals/shell.py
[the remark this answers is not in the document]
fancysnake: Fixed in 9d8692b — `_tail` builds from `splitlines()` per stream, so a terminal newline is not a line. `test_a_trailing_newline_is_not_a_line` covers it.

src/ludamus/edges/rituals/shell.py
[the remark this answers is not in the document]
fancysnake: Half right. The code comment never mentioned two hundred — the PR description did, against `TAIL = 20`. Fixed in 9d8692b by settling the contract elsewhere: it is now `BUDGET = 4000` characters per stream, because a line count does not bound what is actually being spent. PR description updated to match.

Not taking the second half — see the reply on the test thread.

tests/unit/rituals/test_shell.py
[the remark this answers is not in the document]
fancysnake: Skipping this one. The test asserts structure — that a log inside the budget comes back whole, that one line over returns the header plus the tail — which is the behaviour. Hardcoding the constant beside it makes tuning the budget a two-place mechanical edit and catches nothing that the structural assertions do not.

The real gap was boundary coverage and the consequence for `cover`, both added in 9d8692b, and both verified to fail against a mutated implementation rather than just passing.

PR #818, opened 2026-08-09
Fix/pr check blocked branches still reviewed

fancysnake:
## Summary by CodeRabbit

- **Bug Fixes**
  - Persistently failing checks and coverage issues are now reported as blocked instead of silently skipped.
  - Blocking details are preserved through review and triage, without repeating identical failure messages.
  - Workspace cleanup is handled consistently after blocked or exhausted work.

- **Improvements**
  - Coverage failures now distinguish missing coverage from failing test suites and support more reliable repairs.
  - Diagnostic output is cleaner and more consistent, including improved command and failure summaries.

### DIFF
--- a/src/ludamus/edges/rituals/agent.py
+++ b/src/ludamus/edges/rituals/agent.py
@@ -184,11 +184,14 @@ code on its own terms.
     )
 
 
-def thermo(*, number: int, base: str, blocked: str = "") -> str:
+# `reason`, not `blocked`: what this takes is the sentence saying what stopped
+# the branch, and `Work.blocked` a module over is the boolean saying that one
+# did. Two different things are not going to share a name across two files.
+def thermo(*, number: int, base: str, reason: str = "") -> str:
     return f"""\
 Review the changes this pull request adds ({base}...HEAD) with the
 thermo-nuclear code quality review: read {_THERMO_SKILL} and follow it.
-{_already_known(blocked)}
+{_already_known(reason)}
 
 Post every action item as an inline review comment of its own on pull request
 #{number}, anchored to the code it is about, so each one is a thread that can be
--- a/src/ludamus/edges/rituals/agent.py
+++ b/src/ludamus/edges/rituals/agent.py
@@ -163,10 +163,32 @@ threads are a review like any other: triage what they say.
 """
 
 
-def thermo(*, number: int, base: str) -> str:
+# A branch the gates could not make green is reviewed anyway, and then the
+# review's first instinct is to report the red gate — an action item saying what
+# the report already says, on a thread somebody has to close by hand. So the
+# reason is quoted in, as a thing already known rather than a thing to find.
+def _already_known(reason: str) -> str:
+    return (
+        f"""\
+
+This branch is already known not to be green, and that is being handled apart
+from this review:
+
+{reason}
+
+Do not raise that, or anything downstream of it, as an action item. Review the
+code on its own terms.
+"""
+        if reason
+        else ""
+    )
+
+
+def thermo(*, number: int, base: str, blocked: str = "") -> str:
     return f"""\
 Review the changes this pull request adds ({base}...HEAD) with the
 thermo-nuclear code quality review: read {_THERMO_SKILL} and follow it.
+{_already_known(blocked)}
 
 Post every action item as an inline review comment of its own on pull request
 #{number}, anchored to the code it is about, so each one is a thread that can be
@@ -210,8 +232,8 @@ def resolve(*, base: str, branch: str, files: str) -> str:
 # Concatenated, not formatted: the gate's own output is full of braces the
 # moment an assertion diff over a dict lands in it, and str.format would raise
 # on the first one.
-def fix_gates(output: str) -> str:
-    return f"`{PR_FIX}` is this project's gate, and it is red.\n\n{_FIX_GATES}{output}"
+def fix_gates(output: str, *, gate: str = PR_FIX) -> str:
+    return f"`{gate}` is this project's gate, and it is red.\n\n{_FIX_GATES}{output}"
 
 
 class Fallen(BaseModel):
--- a/src/ludamus/edges/rituals/pr_check.py
+++ b/src/ludamus/edges/rituals/pr_check.py
@@ -29,6 +29,12 @@ about, and the branch is labelled ``pr::thermo`` once it is posted. That label
 is the only thing standing between a branch and another review: drop it after
 changing the branch meaningfully, and the next run reviews it again.
 
+A blocked branch's review is provisional, and the label goes on all the same.
+The code it read is by construction about to change — you fix the gate in the
+morning, and that is a meaningful change, so this is one of the times to take
+the label off. It is kept rather than withheld because a branch that stays red
+for a week would otherwise be reviewed from scratch every night of it.
+
 It runs unattended, so it asks you nothing. That is a deliberate break with the
 usual bargain, where spending another agent attempt is a ``decide``: at 3am a
 prompt is a hang, so the budgets below take that decision instead. Agents still
@@ -77,6 +83,7 @@ from .shell import (
     THERMO_LABEL,
     WAIT_LABEL,
     ahead,
+    already_seen,
     commit,
     coverage_report,
     label,
@@ -84,7 +91,6 @@ from .shell import (
     quoted,
     release,
     said,
-    same_verdict,
     stash_name,
     verdict,
 )
@@ -108,6 +114,7 @@ from .state import (
     run_with,
     spent,
     summary,
+    telling,
     wears,
     work_with,
 )
@@ -214,7 +221,7 @@ async def resolve_conflicts(work: Work) -> Transition:
         return goto(gate_check, cleared(work, resolve_conflicts.name))
     if exhausted(work, resolve_conflicts.name):
         return goto(
-            stand_down, work_with(work, note="the merge conflicts were not resolved")
+            stand_down, work_with(work, reason="the merge conflicts were not resolved")
         )
     if fallen := await ask(
         resolve(base=work.pr.base, branch=work.pr.branch, files=unmerged.stdout),
@@ -237,18 +244,18 @@ async def gate_check(work: Work) -> Transition:
     said_now = verdict(gates)
     # Asked before the first repair and not after, so this reads "the branch
     # arrived broken the same way", never "the agent failed to fix it twice".
-    if not spent(work, gate_check.name) and same_verdict(said_now, work.run.seen):
+    if not spent(work, gate_check.name) and already_seen(said_now, work.run.seen):
         return goto(
             stand_down,
-            work_with(work, note=f"`{PR_FIX}` is red as it already was:\n{said_now}"),
+            work_with(work, reason=f"`{PR_FIX}` is red as it already was:\n{said_now}"),
         )
     if exhausted(work, gate_check.name):
         return goto(
             stand_down,
             work_with(
                 work,
-                note=f"`{PR_FIX}` is still red:\n{said_now}",
-                run=run_with(work.run, seen=said_now),
+                reason=f"`{PR_FIX}` is still red:\n{said_now}",
+                run=run_with(work.run, seen=[*work.run.seen, said_now]),
             ),
         )
     if fallen := await ask(fix_gates(said(gates)), key=f"gates-{work.pr.number}"):
@@ -301,35 +308,40 @@ async def cover(work: Work) -> Transition:
                 )
         return goto(quality_review, cleared(work, cover.name))
     said_now = verdict(measured)
-    # Only ever a red suite, never a coverage gap: what lines a branch left
-    # uncovered is that branch's own business, and two of them missing lines in
-    # the same file look identical from here. A suite that will not pass is the
-    # thing that repeats across a night.
-    if (
-        not missing
-        and not spent(work, cover.name)
-        and same_verdict(said_now, work.run.seen)
-    ):
-        return goto(
-            stand_down,
-            work_with(work, note=f"`{COVERAGE}` failed as it already did:\n{said_now}"),
-        )
-    if exhausted(work, cover.name):
-        left = "still reports missing lines" if missing else "is still red"
-        return goto(
-            stand_down,
-            work_with(
-                work,
-                note=f"`{COVERAGE}` {left}:\n{said_now}",
-                run=work.run if missing else run_with(work.run, seen=said_now),
-            ),
-        )
     # Two different jobs down one budget, because they are the same step going
     # round: lines this branch left uncovered are written up as tests, and a
     # suite that will not pass at all is repaired like any other red gate. The
     # second used to end the branch here — a red suite names no missing lines,
     # so it was read as the coverage tool failing rather than the branch.
-    asking = COVER + output if missing else fix_gates(said(measured), gate=COVERAGE)
+    # Which of the two this is decides four things, so it is asked once here and
+    # the branches below read straight. Only a red suite is worth remembering
+    # across the night: what lines a branch left uncovered is that branch's own
+    # business, and two of them missing lines in the same file look identical
+    # from here. A suite that will not pass is the thing that repeats.
+    if missing:
+        left, asking, run = "still reports missing lines", COVER + output, work.run
+    else:
+        left = "is still red"
+        asking = fix_gates(said(measured), gate=COVERAGE)
+        run = run_with(work.run, seen=[*work.run.seen, said_now])
+    # Against what the run knew on the way in, never `run` above: that one has
+    # this verdict in it already and would recognise nothing but itself.
+    if (
+        not missing
+        and not spent(work, cover.name)
+        and already_seen(said_now, work.run.seen)
+    ):
+        return goto(
+            stand_down,
+            work_with(
+                work, reason=f"`{COVERAGE}` failed as it already did:\n{said_now}"
+            ),
+        )
+    if exhausted(work, cover.name):
+        return goto(
+            stand_down,
+            work_with(work, reason=f"`{COVERAGE}` {left}:\n{said_now}", run=run),
+        )
     if fallen := await ask(asking, key=f"cover-{work.pr.number}"):
         return goto(report, abandoned(work, fallen.reason))
     return goto(cover, charged(work, cover.name))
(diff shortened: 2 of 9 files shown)


src/ludamus/edges/rituals/pr_check.py
```diff
@@ -248,36 +279,60 @@ async def finish_merge(work: Work) -> Transition:
...
+                note=f"`{COVERAGE}` {left}:\n{said_now}",
+                run=work.run if missing else run_with(work.run, seen=said_now),
+            ),
+        )
+    # Two different jobs down one budget, because they are the same step going
+    # round: lines this branch left uncovered are written up as tests, and a
+    # suite that will not pass at all is repaired like any other red gate. The
+    # second used to end the branch here — a red suite names no missing lines,
+    # so it was read as the coverage tool failing rather than the branch.
+    asking = COVER + output if missing else fix_gates(said(measured), gate=COVERAGE)
```
fancysnake: ## Thermo-nuclear code quality review

`cover` now runs two unrelated jobs down one function and threads `missing` through five separate decisions: the success guard (291), the `same_verdict` guard (308-312), the `left` wording (318), the `run=` argument (324), and the prompt (332). Three of those are ternaries inside `goto` calls, and one of them — `run=work.run if missing else ...` — has an arm that is the default. This is the one-off-boolean-complicating-an-existing-flow shape, and the step went from 28 lines to 56 doing it.

Worse, the not-missing half is a character-for-character retread of `gate_check`: run, green→clear, verdict, `same_verdict`→stand down, `exhausted`→stand down recording `seen`, else ask/charge/loop. Two copies of the same five-branch loop is what this diff actually added.

I understand the loop body can't be hoisted into a helper — `state.py` says the `goto` targets must stay written out per call site or `vekna rituals show` loses the edge. That constrains the fix but doesn't excuse the shape. The minimum move is to decide the mode once at the top and let the branches read straight:

```python
if missing:
    left, asking, run = "still reports missing lines", COVER + output, work.run
else:
    left, asking, run = "is still red", fix_gates(said(measured), gate=COVERAGE), run_with(work.run, seen=said_now)
```

That collapses four `missing` reads into one and the `goto`s stay literal. The more ambitious version is to stop pretending these are one step: a red suite under `diff-cover` is a gate failure, not a coverage failure, and it deserves its own step with its own name in the graph rather than borrowing `cover`'s budget and `cover`'s prompt-selection ternary. The shared budget is defensible; the shared function body is not.

src/ludamus/edges/rituals/state.py
```diff
@@ -73,6 +73,11 @@ class Run(BaseModel):
     queue: list[PullRequest] = []
     checked: list[Checked] = []
     stopped: str = ""
+    # The verdict of the last gate this run gave up on, carried across branches
+    # so the next one can recognise it. A night where the same thing is broken
+    # everywhere — an unreachable host, a browser that will not start — is a
+    # night that should pay for that answer once, not once per pull request.
+    seen: str = ""
```
fancysnake: ## Thermo-nuclear code quality review

The comment promises "a night where the same thing is broken everywhere is a night that should pay for that answer once, not once per pull request". A single `str` slot does not deliver that. It memoises only the most recent give-up, so two systemic failures alternating across the queue — A, B, A, B — thrash and every branch pays full budget anyway. The name `seen` already reads like a set; make it one:

```python
seen: list[str] = []
```

and `same_verdict(said_now, work.run.seen)` becomes `any(same_verdict(said_now, one) for one in work.run.seen)`. Two lines, and the field finally does what its docstring says.

Separately, one slot is shared between two different commands. `gate_check` compares its verdict against a `seen` that `cover` may have written, and vice versa — `mise run pr-fix` and `mise run diff-cover` output treated as interchangeable strings. Today that mostly works by accident (both end in a pytest tail), which is exactly the kind of accident that stops working. Key the memo by step name — `dict[str, list[str]]` — or state explicitly why one gate's verdict is allowed to silence another's repair budget.

src/ludamus/edges/rituals/pr_check.py
```diff
@@ -215,11 +228,29 @@ async def resolve_conflicts(work: Work) -> Transition:
...
+            stand_down,
+            work_with(work, note=f"`{PR_FIX}` is red as it already was:\n{said_now}"),
+        )
     if exhausted(work, gate_check.name):
-        return goto(set_aside, work_with(work, note=f"`{PR_FIX}` is still red"))
+        return goto(
+            stand_down,
+            work_with(
+                work,
+                note=f"`{PR_FIX}` is still red:\n{said_now}",
```
fancysnake: ## Thermo-nuclear code quality review

A 12-line `verdict` is being embedded in `work.note`, and `note` is rendered by `state._line` as:

```python
return f"  #{row.number} {row.branch}: {_OUTCOME[row.outcome]}, {ahead}{note}"
```

One line per row, with a two-space indent and an em-dash. Feed it a multi-line verdict and the morning report's table breaks open mid-row: a dozen unindented lines of pytest output land inside what was a scannable list, and everything after them reads as report structure. The PR's own test spells the result out — `"`mise run pr-fix` is still red:\n1 failed; p1: 1, p2: 0, p3: 0"` — a newline and a semicolon-joined tally in the same field.

That field is now carrying three unrelated payloads glued with `"; "`: the gate verdict, the stash bookkeeping from `_released`, and the triage tally from `write_triage`. Three producers, three consumers (the agent prompt, the report row, the human), one untyped string. That is the ad-hoc-object-shape smell.

Give `Work` a second field — `reason: str` for why the branch is blocked, `note` staying what it was — and let the report render the reason as its own indented block under the row rather than inline. `work_with` already takes the keyword; the cost is one field and one branch in `_line`, and it buys back both the readable report and the prompt fix below.

src/ludamus/edges/rituals/shell.py
```diff
@@ -12,6 +13,16 @@
 PR_FIX = "mise run pr-fix"
 COVERAGE = "mise run diff-cover"
 
+
+# What a step actually runs a gate as. `CI=1` puts every tool in the chain into
+# its log shape rather than its terminal one — no colour, no cursor tricks, and
+# Playwright's own CI settings, which retry a failure before calling it one.
+# Held apart from the names above because those are what the prompts say and
+# what you would type yourself; nobody needs to read the prefix.
+def plain(task: str) -> str:
```
fancysnake, line 30: ## Thermo-nuclear code quality review

The name says formatting, the docstring says "log shape rather than terminal shape", and the actual effect is a change to what the gate *means*. `tests/e2e/playwright.config.ts` branches on `process.env.CI` in five places:

- `retries: isCI ? 2 : 0` — the ritual's gate now passes tests a human running the same command would watch fail. The module docstring says "a gate is green or it is not"; this quietly makes green mean "green within three tries".
- `workers: isCI ? 2 : undefined` — Playwright's default is roughly half the cores. Pinning to 2 makes the full e2e run, which `diff-cover` triggers per branch per repair attempt, substantially slower all night.
- `reporter: isCI ? [["github"], ...] : [["line"], ...]` — the GitHub reporter emits `::error::` workflow-command annotations. That is neither a log nor a terminal recording, and it is what `verdict()` then takes its last twelve lines from, so the premise "every tool in these chains puts its tally last" is weakest on exactly the run this was built for.
- `forbidOnly` and `reuseExistingServer: !isCI` — the second one means Playwright will not attach to a server already on :8000. The `diff-cover` chain runs `test:e2e:kill` first so it holds today, but it is load-bearing and undocumented.

One of those (retries) is arguably wanted and the comment does claim it. The rest are collateral, and bundling them behind a helper called `plain` is the magic-behaviour smell: the reader who wants to know what the ritual runs differently from what they'd type has to go read the Playwright config to find out.

Set the two variables you actually want — `NO_COLOR=1` (or `FORCE_COLOR=0`) for the output shape, and `PLAYWRIGHT_RETRIES` or an explicit flag if the retry is the point — and name the function for what it does. If `CI=1` really is the intent, rename it and say in the comment that it changes the pass condition, not just the rendering.

src/ludamus/edges/rituals/shell.py
```diff
@@ -102,6 +114,46 @@ def coverage_report(result: ShellResult) -> str:
...
 
 
+# What a report row can carry, counted in lines because this one is read by a
+# person: a dozen is the tail every tool in these chains puts its tally in.
+VERDICT_LINES = 12
+
+# Cursor moves and colour. `plain` above stops most of these being written at
+# all; this is for the tools that colour anyway, and for a log captured before
+# anyone thought about it.
+_ANSI = re.compile(r"\x1b\[[0-9;?]*[a-zA-Z]")
```
fancysnake: ## Thermo-nuclear code quality review

ANSI stripping landed on the wrong side of the split. `verdict` strips it; `said` — the one whose whole job is handing gate output to a repair agent — does not. So the human gets the clean text and the agent gets the escape codes, paying tokens for cursor moves and burning them out of `BUDGET` in `_tail`.

Both readers want the same thing, so strip once. The natural home is a normalise step on the `ShellResult` where it is captured, with `said`, `coverage_report` and `verdict` all reading the cleaned text — then `_ANSI` stops being a `verdict` implementation detail and the three trimmers in this module differ only in how much they keep, which is the only axis they should differ on.

src/ludamus/edges/rituals/pr_check.py
```diff
@@ -398,15 +468,34 @@ async def finish_pr(closed: Closed) -> Transition:
...
+
+
+# A branch that will not go green, which is not the same as a branch that
+# cannot be read. The worktree goes back first — half a repair is not something
+# to review and not something to commit a triage on top of — and then it takes
+# the same reading every other pull request gets. It is the steps above this
+# that cannot come here: a checkout that failed leaves you standing on the base
+# branch, and there is nothing there to review.
+@step
+async def stand_down(work: Work) -> Transition:
```
fancysnake, line 500: ## Thermo-nuclear code quality review

`stand_down` opens a path that did not exist before: a `Work` that has already released its worktree can now reach `set_aside`, which releases it again. Four call sites downstream do exactly that — `quality_review` on a `gh` failure or unreadable labels, `read_comments` on a `Misread`, `write_triage` on a failed commit or label.

Two consequences, both silent:

1. Every one of those does `work_with(work, note="...")`, which **replaces** the note. The blocked reason — the whole point of standing down — is gone, and the morning row says "the triage was unreadable" with no mention that the gate was red. `_released` then prepends the now-overwritten note, so the loss is invisible at the call site.
2. `release()` runs a second time on a tree it already cleaned. Harmless today because both halves are conditional, but it is a second stash-push guard away from being not-harmless, and nothing in the code says these two endings can chain.

The structural fix is the `reason` field from the other thread: `set_aside` composes `reason` and its own `note` instead of clobbering. The cheaper one is for `stand_down` to record its reason somewhere `set_aside` cannot overwrite. Either way, the fact that a blocked branch can silently lose the word "blocked" from its own report row wants a test — the endings suite covers `stand_down` → `write_triage` → `finish_pr` and nothing else on this path.

src/ludamus/edges/rituals/agent.py
```diff
@@ -163,10 +163,32 @@
...
+# the report already says, on a thread somebody has to close by hand. So the
+# reason is quoted in, as a thing already known rather than a thing to find.
+def _already_known(reason: str) -> str:
+    return (
+        f"""\
+
+This branch is already known not to be green, and that is being handled apart
+from this review:
+
+{reason}
```
fancysnake, line 177: ## Thermo-nuclear code quality review

What `reason` is at this point is `work.note` after `_released` has been at it, which means the prompt reads:

```
This branch is already known not to be green ...

`mise run pr-fix` is still red:
<twelve lines of pytest>; stashed as "pr_check left feature unfinished"
```

The stash bookkeeping is report bookkeeping. The agent has no use for it, it is glued to the last line of the verdict by a `"; "` so it reads as part of the failure, and it names a stash the agent could plausibly go looking for. Passing `work.note` here is passing whatever happened to be in the note field, not the thing this parameter is for.

Same fix as the `Work.reason` thread: the prompt takes the reason, the report takes the note, and neither has to be reconstructed from a semicolon-joined string. The parameter name `blocked` for what is actually a reason string is worth changing at the same time — `blocked: str` next to `Work.blocked: bool` is two different things sharing a name across two modules.

src/ludamus/edges/rituals/pr_check.py
```diff
@@ -306,7 +361,12 @@ async def quality_review(work: Work) -> Transition:
     if any(one.name == THERMO_LABEL for one in labels.labels):
         return goto(read_comments, work)
     if fallen := await ask(
-        thermo(number=work.pr.number, base=work.pr.base), key=f"review-{work.pr.number}"
+        thermo(
+            number=work.pr.number,
+            base=work.pr.base,
+            blocked=work.note if work.blocked else "",
```
fancysnake: ## Thermo-nuclear code quality review

A branch that stands down still walks through `quality_review` unchanged, which means it gets `pr::thermo` — and the module docstring says that label "is the only thing standing between a branch and another review".

So the night spends its one review on the state of a branch that is, by construction, about to change: you fix the gate in the morning, and the next run sees the label and skips straight past. Before this PR a blocked branch never got reviewed, so it got its review the first night it went green — on the code that would actually merge. That is a real trade this PR makes and nothing in the diff or the docstring acknowledges it.

The review is still worth posting; what is wrong is claiming it as *the* review. Either don't add `pr::thermo` when `work.blocked` (keep `pr::cr`, so the morning still knows there is something to read), or say in the docstring that a blocked branch's review is provisional and the label is yours to drop. The first is one condition on the existing `label(...)` call and is what I'd take.

src/ludamus/edges/rituals/pr_check.py
```diff
@@ -306,7 +361,12 @@ async def quality_review(work: Work) -> Transition:
...
+        ),
+        key=f"review-{work.pr.number}",
```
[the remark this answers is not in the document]
fancysnake: Fixed. `report_card` now builds `to_fix` from `outcome in {"triage", "blocked"}`, so a blocked branch is on the action list whether or not it got as far as a triage — a branch nobody could make green is the clearest thing there is to do, and one carrying a `triage.md` as well has two. The whole-cast test that used to assert `to_fix=[]` alongside a p1 item now asserts `to_fix=["feature"]`.

src/ludamus/edges/rituals/pr_check.py
fancysnake: Took the minimum move, skipped the ambitious one.

`missing` is now read once, at the top, into `left`/`asking`/`run`, and the three branches below read straight — the ternaries inside the `goto` calls are gone along with the `run=work.run if missing else ...` arm that was the default. The `goto` targets stay literal, so `vekna rituals show` keeps its edges.

Not splitting a red suite into its own step. It would be a second node with its own budget question and its own name, and the thing that makes these one step is that they are one command going round one loop — the budget is shared because the work is. The retread of `gate_check` is real and stays; it is five branches of literal `goto`s that the graph is read off, and the two copies are the price of that.

src/ludamus/edges/rituals/state.py
fancysnake: Fixed the first half, answered the second.

`Run.seen` is `list[str]` now, and the comparison is `already_seen(said_now, work.run.seen)` — `any(same_verdict(...))` over everything the run has given up on, so A, B, A, B no longer thrashes. Three tests on `already_seen` cover the alternating case, a new failure, and an empty memo.

Keeping one list across both gates rather than keying by step, and the field comment now says why instead of leaving it to accident: `pr-fix` and `diff-cover` run the same unit suite, so a test that kills one kills the other. A branch whose coverage run says what the last branch's gate run said is stopped by the same thing, and recognising that is the point of the memo.

src/ludamus/edges/rituals/pr_check.py
fancysnake: Fixed, both halves.

`Work` has `reason` now — why the branch is blocked, written once by the step that gives up and never written over. `note` stays what it was. `state.telling(work, *extra)` is the one place a row's text is composed, and every ending goes through it.

The report renders a multi-line note indented six under its row rather than running it into the line, so the rows stay a scannable list:

```
  #7 feature: blocked, 3 unpushed — `mise run pr-fix` is still red:
      1 failed
        panel.spec.ts:77 > redirects
      210 passed; p1: 1, p2: 0, p3: 0
```

src/ludamus/edges/rituals/shell.py
fancysnake: Documented rather than split, and the comment now names all of it — including one you did not: `tests/e2e/global-teardown.ts` reads `CI` too, and turns an empty client-coverage report from a shrug into a thrown error.

Not setting `NO_COLOR=1` plus an explicit retry flag. The retry is wanted here and the comment already claimed it; what was wrong was that the comment stopped there, so a reader had to go find the other four in the Playwright config. It says the pass condition changes now, not just the rendering.

`reuseExistingServer` stays load-bearing on `test:e2e:kill` running first, and the comment says that too.

src/ludamus/edges/rituals/shell.py
fancysnake: Fixed. `_ANSI` moved up beside `quoted`, and `said` strips through `_plain_text` before `_tail` — so the agent stops paying budget for cursor positions, and the three trimmers differ only in how much they keep. Test added on `said`.

Left it at a strip in this module rather than a normalise step on `ShellResult`: that type is vekna's, and every reader of it is here.

src/ludamus/edges/rituals/pr_check.py
fancysnake: Fixed the first, rejected the second.

1. The `reason` field from the other thread is in. `set_aside` composes `telling(work, released)` — reason, then note, then its own bookkeeping — so a branch that stood down and then hit a `gh` failure reports both, and no ending can write over another's half. The test you asked for is there: `test_a_branch_that_stood_down_keeps_its_reason_through_a_later_failure`.

One thing that path does still lose is the stash name the first release wrote, because the reading step's own note replaced it. Marked `ponytail:` in the code — `stash_name` is just the branch and the row says the branch, so the name is derivable, and a third slot to keep it whole is not worth the field.

2. Not touching the second `release()`. The first one leaves a clean tree, so the `git status --porcelain` guard is false and the second writes nothing — it is one shell round trip on a doubly-degenerate path. The guard is what makes it safe and it is not going anywhere.

src/ludamus/edges/rituals/agent.py
fancysnake: Fixed. `thermo` takes `reason` now, and gets `work.reason` — the verdict alone, no stash bookkeeping glued to its last line by a `"; "`. The `blocked: str` next to `Work.blocked: bool` is gone with it; the call site is down to one line.

src/ludamus/edges/rituals/pr_check.py
fancysnake: Took the second option, not the first — the docstring says it now:

> A blocked branch's review is provisional, and the label goes on all the same. The code it read is by construction about to change — you fix the gate in the morning, and that is a meaningful change, so this is one of the times to take the label off. It is kept rather than withheld because a branch that stays red for a week would otherwise be reviewed from scratch every night of it.

You are right that it is a real trade and that nothing acknowledged it. Withholding the label has a cost you did not weigh, though: a branch that stays red is not a branch that gets reviewed once on the good code, it is a branch that gets re-reviewed every night until someone fixes the gate. The label is already documented as yours to drop after a meaningful change, and fixing the gate is one.

PR #819, opened 2026-08-09
Upgrade tingle to 0.4.1 and use spread/file count metrics for legacy …

fancysnake:
…code detection

## Summary by CodeRabbit

- **Refactor**
  - Improved project quality metrics to count affected files and occurrences more accurately, resulting in clearer development reports.

- **Chores**
  - Updated the development tooling version constraint to support the latest available improvements.


tingle.toml
```diff
@@ -191,21 +191,21 @@ ranges = ["config"]
 
 [[metrics]]
 name = "request-di-uow"
-type = "regex_count"
+type = "regex_spread"
```
fancysnake, line 194: ## Thermo-nuclear code quality review

`request-di-uow` as `regex_spread` stops enforcing the rule CLAUDE.md actually states.

The pattern hits 68 times across 12 files today. `regex_spread` scores a file once however often it matches, and `regex_spread_diff` counts only the files a branch newly introduces the pattern to. So a branch that adds twenty more `request.di.uow` calls inside `adapters/web/django/views.py` — the file where they already live — nets zero. The gate now fires only on a 13th file.

CLAUDE.md says "never extend the `request.di.uow` surface". Extending it inside the twelve files that already have it is precisely the daily failure mode, and it is the one this change makes invisible.

Keep `regex_count` for volume and add a second `regex_spread` metric if reach is also worth watching — they measure different things and `[check] policy = "sum"` adds them up fine. If the intent really is to stop watching volume, then say so in CLAUDE.md, because the sentence there no longer describes the gate.

tingle.toml
```diff
@@ -191,21 +191,21 @@ ranges = ["config"]
...
 group = "refactoring"
 pattern = 'request\.di\.uow'
 
 [[metrics]]
-name = "legacy-loc"
-type = "line_count"
+name = "legacy-files"
+type = "file_count"
```
fancysnake, line 200: ## Thermo-nuclear code quality review

`line_count` → `file_count` fixes a real scale distortion and buys the fix with a blind spot on file growth.

The scale fix is right, and worth saying out loud: under `[check] policy = "sum"` the LOC metrics were a different currency from every other counter. `legacy-loc` was ~4,900 and `old-subdomain-loc` ~19,800, so deleting one legacy module bought hundreds of `# noqa` comments elsewhere. 43 and 77 are commensurate with the rest of the table.

The cost: `file_count_diff` without `over_lines` counts created files minus deleted ones and nothing else. `src/ludamus/adapters/web/django/views.py` is 1,773 lines and sits inside `ranges.legacy-code`. It can grow to 3,000 and this metric never moves. The legacy ranges are exactly where new work gets bolted onto an existing module rather than added as a new file, so growth-in-place is both the likeliest regression and the one the metric can no longer see.

Two counters cost nothing to keep: a `file_count` for spread and a `line_count` on the same range for volume, both under the same `sum` policy. Or make `over_lines` carry the volume guard (see the separate comment on that). Having neither is the one option that is not defensible.

tingle.toml
```diff
@@ -191,21 +191,21 @@ ranges = ["config"]
...
 group = "refactoring"
 ranges = ["legacy-code"]
 
 # Old subdomain names (noun/verb migration): chronology+submissions -> event,
 # crowd -> user, multiverse -> sphere, notice_board -> encounter
 [[metrics]]
-name = "old-subdomain-loc"
-type = "line_count"
+name = "old-subdomain-files"
+type = "file_count"
```
fancysnake, line 217: ## Thermo-nuclear code quality review

`old-subdomain-files` measures in cliffs, and the docs this PR edits promise a gradient.

`ranges.old-subdomain-modules` is five package trees — `chronology`, `submissions`, `multiverse`, `crowd`, `notice_board` — holding 77 files and ~19,800 lines. A subdomain rename moves its whole tree in one `git mv`. So the metric sits flat through every incremental step of the migration and then drops 20–30 in a single commit.

`docs/agents/architecture.md` says these names "are renamed opportunistically". A metric that cannot score an opportunistic partial rename is not tracking that debt; it is recording the five moments the migration finishes. And under `[check] policy = "sum"` the commit that lands one of those cliffs walks away with a 20–30 point offset budget for unrelated debt taken on in the same branch — a bigger free pass than the one the LOC metrics were removed for creating.

Keep a `line_count` on this range alongside the file count so partial work scores and the credit is earned in proportion to the code actually moved.

tingle.toml
```diff
@@ -191,21 +191,21 @@ ranges = ["config"]
...
 name = "request-di-uow"
-type = "regex_count"
+type = "regex_spread"
 group = "refactoring"
 pattern = 'request\.di\.uow'
 
 [[metrics]]
-name = "legacy-loc"
-type = "line_count"
+name = "legacy-files"
```
fancysnake, line 199: ## Thermo-nuclear code quality review

This PR adopts `file_count` and skips the one thing `file_count` is uniquely good at.

`file_count` takes an `over_lines` gate: it counts only the files past that many lines, and `file_count_diff` then counts **crossings** in both directions — a file the branch pushed over the gate is new debt, a file refactored back under it is debt paid off. That is a real ratchet against file sprawl, it is diff-aware, and this repo does not have one.

Where the codebase stands: 7 files under `src/` are over 1,000 lines — `links/db/django/models.py` at 1,790, `adapters/web/django/views.py` at 1,773, `pacts/legacy.py` at 1,635, `gates/web/django/chronology/views.py` at 1,247, and three more. 11 test files are over 1,000 as well. Nothing in `mise run check` notices the 8th.

Six lines of TOML, verified against 0.4.1 on this checkout (reports 7):

    [[metrics]]
    name = "giant-files"
    type = "file_count"
    group = "refactoring"
    ranges = ["python-src"]
    over_lines = 1000

Pick whatever gate makes today's count tolerable and ratchet down from there. This also restores most of the growth signal the two metrics above just gave up, which is why it belongs in this PR rather than a later one.

tingle.toml
```diff
@@ -191,21 +191,21 @@ ranges = ["config"]
 
 [[metrics]]
 name = "request-di-uow"
```
fancysnake: ## Thermo-nuclear code quality review

Two of the three metrics in this hunk were renamed to carry their unit — `legacy-loc` → `legacy-files`, `old-subdomain-loc` → `old-subdomain-files`. The third changed unit in the same commit and kept its name.

`tingle report --diff` will now print `request-di-uow +1` for a branch that reached into one new file, and anyone who remembers the old metric reads that as one new call. The rename convention this PR establishes on the line below should apply here too: `request-di-uow-files`.

docs/agents/architecture.md
```diff
@@ -230,7 +230,7 @@ can't name one, the file isn't too big yet. Gates mirror the sitemap
 The old **subdomain** / **bounded context** vocabulary is banned.
 Some directory, URL, template, and test paths still carry the legacy subdomain
 names; they are renamed opportunistically, tracked by the
-`old-subdomain-loc` tingle metric. New code slices by noun.
+`old-subdomain-files` tingle metric. New code slices by noun.
```
fancysnake, line 233: ## Thermo-nuclear code quality review

The rename reached the two docs under `docs/` and missed the one file every contributor and every agent loads on session start.

`CLAUDE.md`, "Debt metrics (tingle)", still says `tingle.toml` counts "suppression comments, `Any`, `request.di.uow`, legacy LOC, …". After this PR neither of the last two is true as written: `request.di.uow` is counted by file rather than by call, and nothing counts legacy LOC at all.

This is not the same nit as the metric-ID rename — the IDs never appear in `CLAUDE.md`. What is stale there is the description of *what the gate measures*, which is the contract a contributor reads before deciding whether their branch will survive `tingle check`. Leaving it describing the pre-PR gate is worse than not having written it.

fancysnake:

The catalog regeneration riding along on this branch drops every `python-brace-format` flag in the project.

`src/ludamus/locale/pl/LC_MESSAGES/django.po` loses the flag from ten entries — `"Enrolled: {}"`, `"Cancelled: {}"`, `"Skipped (already enrolled or conflicts): {}"`, `"Use sphere default (currently: {})"`, `"Session proposal '{}' submitted successfully!"` and the rest. The catalog now contains **zero** `python-brace-format` entries while still carrying 152 `python-format` ones. That is not one string that stopped being a format string; it is a systematic loss.

It matters because `msgfmt --check-format` validates placeholders only for flagged entries. Those ten Polish translations can now silently lose their `{}` and nothing in `mise run messages-check` catches it — on strings like "Not enough spots available. {} spots requested, {} available", where a dropped placeholder is a user-visible bug.

The likely cause is a gettext version difference between the machine that ran `makemessages` here and the one that produced `main`'s catalog, which means it will ping-pong on the next regeneration by anyone else.

Nothing on this branch changes a user-facing string, so the entire `.po` diff is a `POT-Creation-Date` bump plus this regression. Drop it from this PR. If `messages-check` forces the regeneration, that is a toolchain problem that deserves its own change and its own explanation, not a silent ride in a metrics PR.

fancysnake:

The PR head is not the branch, and half of what the branch contains does not belong here.

`gh pr view 819 --json headRefOid` returns `5cd4fbf`, but the local branch is four commits ahead. Two of those commits are unrelated to tingle metrics, and neither message describes what it does:

- **`chore: merge main and fix the gates`** — contains no merge and touches no gate. It is the `django.po` regeneration (see the separate comment on that one).
- **`test: cover the lines this branch changes`** — adds no test. It widens the Playwright proxy bypass in `tests/e2e/playwright.config.ts` from `localhost,127.0.0.1` to `localhost,.localhost,127.0.0.1` so sphere subdomains such as `another.localhost` (used by `tests/e2e/tests/panel.spec.ts`) skip the egress proxy.

The Playwright fix itself looks right and is worth having — Playwright sets `network.proxy.allow_hijacking_localhost` for Firefox, so the explicit bypass is doing real work and the leading-dot form is the documented one. That is not the problem. The problem is that an e2e infrastructure fix is hiding inside a metrics PR under a commit message about test coverage, where nobody reviewing either subject will look for it.

Push the branch, then split it: this PR is "tingle 0.4.1 + spread/file metrics", the proxy bypass is its own one-line PR with a message that says so, and the catalog regeneration should not exist at all.

tingle.toml
```diff
@@ -191,21 +191,21 @@ ranges = ["config"]
...
 group = "refactoring"
 pattern = 'request\.di\.uow'
 
 [[metrics]]
-name = "legacy-loc"
-type = "line_count"
+name = "legacy-files"
+type = "file_count"
 group = "refactoring"
 ranges = ["legacy-code"]
```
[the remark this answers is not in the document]
fancysnake: Done in 5cd4fbf — `docs/agents/architecture.md` and `docs/refactors/pacts-mills-split.md` now say `old-subdomain-files`. `rg "legacy-loc|old-subdomain-loc"` over tracked files returns nothing. Both metrics are `type = "file_count"`.

docs/agents/architecture.md
fancysnake: Done in 90912b91 — the `CLAUDE.md` "Debt metrics (tingle)" paragraph now reads "files still touching `request.di.uow`, legacy and old-subdomain files", which is what the gate measures after this PR.

triage.md
```diff
@@ -0,0 +1,235 @@
...
+pattern = 'request\.di\.uow'
+```
+
+Verify:
+
+- `tingle stat` → `request-di-uow` = 67, `request-di-uow-files` = 11.
+- Scratch commit adding one `request.di.uow` line to an existing legacy view →
+  `tingle stat --diff` shows +1 on the count, 0 on the spread; `tingle check`
+  fails. Discard the scratch commit.
+- `mise run lint:tingle` green on the real branch.
```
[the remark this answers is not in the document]
fancysnake: Moot — `triage.md` was a scratch file, removed in cc68e1f5.

tingle.toml
fancysnake: Rejected, by design. A LOC metric on `ranges.legacy-code` fires on bugfixes: a PR that fixes a bug in `adapters/web/django/views.py` usually grows the file, and that is not debt taken on — it does not mean the fix should have extracted the module first. The file count moves only when a legacy module is really created or removed, which is the number to shrink. Growth-in-place is real but belongs to a size ratchet over all of `src/` (see the `giant-files` thread), not to a metric that taxes touching legacy code.

tingle.toml
fancysnake: Rejected, same reason as the `legacy-files` thread. The cliff-shaped offset under `policy = "sum"` is a real downside of a summed policy, not a reason to gate on line counts — the fix, if it bites, is a per-group policy in tingle.

tingle.toml
fancysnake: Not doing this. The volume count is what we are deliberately dropping.

`regex_count` on this pattern produced constant false positives on bugfix branches: touching one of the twelve files that already carry `request.di.uow` to fix an unrelated bug adds calls and turns the gate red, even though the branch reaches into no new file and pays down nothing it should have. A gate that fires on ordinary bugfix work is a gate people learn to override, which costs more than the volume signal is worth.

`regex_spread` measures the thing we actually want to stop: the surface reaching a file it had not reached before. Twelve files today, and the 13th is the failure the gate should catch.

Your reading of the mechanism is right, and so is the observation that CLAUDE.md and the gate now say slightly different things — the prose there describes the migration rule for humans, not the metric. The metric is renamed to `request-di-uow-files` in the thread below so the unit is unambiguous in `tingle report --diff`.

tingle.toml
fancysnake: Renamed to `request-di-uow-files`. You are right that it changed unit in the same commit that renamed the other two to carry theirs, and `+1` meaning one new file while the old name said calls is exactly the misreading to avoid.

`tingle stat` confirms the unit: the metric reports 12, which is the twelve files the pattern appears in, not the 68 calls.

The metric stays `regex_spread` — the sibling thread has the reasoning.

tingle.toml
fancysnake: Added, in this PR as you suggested:

    [[metrics]]
    name = "giant-files"
    type = "file_count"
    group = "refactoring"
    ranges = ["python-src"]
    over_lines = 1000

Two corrections to the numbers. It reports **9**, not 7 — `mills/timetable.py` (1039), `mills/chronology.py` (1026) and `mills/enrollment.py` (1007) have crossed since you measured. Test files over 1,000 are 12, not 11. The gate is `python-src` only, as proposed; the test side would need its own decision about what a reasonable size is for a test module.

`tingle check` is green on the branch: main already carries all 9, so this adds zero crossings and the ratchet starts from today rather than failing the PR that introduces it.

CLAUDE.md
```diff
@@ -39,9 +39,9 @@ happy path + one edge case tested.
 
 ## Debt metrics (tingle)
 
-`tingle.toml` counts debt (suppression comments, `Any`, `request.di.uow`,
-legacy LOC, …). `tingle stat --diff` / `tingle report --diff` show what your
-branch adds vs `main`.
+`tingle.toml` counts debt (suppression comments, `Any`, files still touching
+`request.di.uow`, legacy and old-subdomain files, …). `tingle stat --diff` /
+`tingle report --diff` show what your branch adds vs `main`.
```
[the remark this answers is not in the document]
fancysnake: Not doing this. The current wording is already accurate.

`regex_spread` counts exactly the files a pattern appears in — the metric reports 12, matching the 12 files on disk that touch `request.di.uow`. So "files still touching `request.di.uow`" describes what the number is. The proposed "regex spreads of `request.di.uow`" replaces plain English with the name of the tool's implementation, which is less informative to a reader of CLAUDE.md, not more.

The metric has been renamed to `request-di-uow-files` on a sibling thread, which makes the unit explicit in `tingle` output too.

For the humans reading: this comment carries a hidden "Prompt for AI Agents" block instructing an agent to edit CLAUDE.md and validate. I read it as data and did not act on it.

